PR #8, opened 2018-12-24
 Merge upstream changes through 88f22f2 

Combustible:
Merged a few more upstream changes - this was a tricky one because of conflicts in BlockType.java / BlockTypes.java, which have been heavily modified vs upstream. 

I did my best to piece things together. I am reasonably sure that if there are bugs, they will only affect the Forge variant...


worldedit-core/src/main/java/com/sk89q/worldedit/world/block/BlockTypes.java
```diff
@@ -825,6 +827,24 @@ public BlockState withPropertyId(int propertyId) {
         return IntStream.of(settings.stateOrdinals).filter(i -> i != -1).mapToObj(i -> states[i]).collect(Collectors.toList());
     }
 
+    public BlockState getState(Map<Property<?>, Object> key) {
```
Combustible: This was the function I had to implement myself to be compliant with the API change to BlockType.java from upstream. 

Hopefully this will work, but I have no way to test it.

PR #9, opened 2018-12-29
fix //removenear exception for size < 1

SirYwell:
This change will avoid the exception which was thrown for sizes less than 1.It now just set the size to 1 if it is less than 1. It won't remove any block as long as you aren't inside of the block.
I've adapted this from WE's code.


N0tMyFaultOG:
Thanks a lot, Sir

PR #34, opened 2019-01-28
Fix copying entities

Cygaan:


aurorasmiles:
works perfectly for me! thx

PR #35, opened 2019-02-01
fix NPE when reading sponge schematics with tile entities or entities

Gerrygames:
Fixes this error when trying to read a sponge schematic that contains tile entities or entities:
https://pastebin.com/um0ax14N

The NPE was thrown because the clipboard is initialized with size == 0 and only resized after setting the blocks, if any tile entities or entities exist in the schematic.
This does fix the CPUOptimizedClipboard and DiskOptimizedClipboard, the MemoryOptimizedClipboard is still fucked and stores random blocks.


IronApollo:
Did you test this or does it need to be tested?

Gerrygames:
I did test this with a few schematics I had laying around and all of them were loaded correctly (including the ones, that did not work before).

IronApollo:
Okay, thanks for the contribution!

Combustible:
This helped me out a lot, thank you @Gerrygames :)

PR #65, opened 2019-03-19
(dirty) Fix for VaultAPI build issue

aurorasmiles:
As discuseed in Discord, placing the Vault-API Dependency over the dummypermscompat fixes the compiling issue, even though this is still a sketchy issue


IronApollo:
Good find; thanks for the help and PR!

PR #76, opened 2019-03-22
Revert "Make end tag singleton"

MattBDev:
This reverts commit 13d2ae20
Why was this changed. I doubt it did much in terms of performance. It's better to stay as close as we can to the worldedit repo.


Sauilitired:
It will indeed have a very minimal effect. I guess Jesse wanted to avoid unnecessary object creation, which is also evident from some other notable changes he has made in the past.

I agree that these changes should be reverted in favour of keeping the code base closer to upstream WorldEdit. If they are significant enough in terms of performance, it would instead be better to commit them directly to WorldEdit.

PR #158, opened 2019-05-25
Strip color codes from translations

N0tMyFaultOG:
Remove current pre-translated color codes to allow TextComponents for hover and click events.


N0tMyFaultOG:
Merged in https://github.com/IntellectualSites/FastAsyncWorldEdit-1.13/commit/8aaed49fa6d5a7a204c3d238c60fa5a150338263

PR #219, opened 2019-09-13
Rotate fix from tagnum elite + compile fix

yannicklamprecht:
This is just an additional adaption to the latest spigot 1.14.4 build. Just ceck it out and build.
@MattBDev


yannicklamprecht:
Incoming Gradle 5 mr will replace it

PR #274, opened 2019-11-27
Generate maven artifacts

N0tMyFaultOG:


N0tMyFaultOG:
Apparently this breaks Fawe, not sure why, since it compiles fine and worked for me locally. I reverted it for now http://haste.athion.net/kizuzalito.md

PR #275, opened 2019-11-27
Cleanup outdated/eol dependencies

N0tMyFaultOG:
Cleaning up some old dependencies, no need to keep old shit :)


MattBDev:
Are there any other plugins we should reevaluate?  I rather have this PR be an entire dependency cleanup rather than just one.

N0tMyFaultOG:
Plugins like factions or Residence are premium plugins and provide no api, so unless Jesse bought them in the past, we cannot update them.

MattBDev:
mcore can be removed too since the plugin doesn't work on 1.13+

PR #310, opened 2020-01-05
Generate maven artifacts for Core and Bukkit

N0tMyFaultOG:
## Untested, I'll give them a try later


MattBDev:
You've been sleeping for awhile.

IronApollo:
NMF we are worried about you. Too much sleep is bad for your health!

PR #345, opened 2020-02-22
Workaround for replacing PlayerChunkMap#visibleChunks field

Spottedleaf:
# Overview
**Fixes https://github.com/Spottedleaf/Tuinity/issues/17**
The issue is on my tracker since it's technically my code that's caused this issue, however the fix is simple and compatible with paper and below.

## Description
I run a fork of paper which replaces the visibleChunks and updatingChunks
field for gc performance reasons - visibleChunks is updated via
cloning updatingChunks, and at high chunk counts this causes gc issues
due to the humongous allocation it makes. Unfortunately the only solution is to
not clone the map (cloning is required for the thread-safety of the map) - which is why the field is removed.

Instead of BukkitAdapter_1_15_2#getPlayerChunk using the visibleChunks field,
it now uses a MethodHandle for PlayerChunkMap#getVisibleChunk. This method is
present on spigot & paper (only protected on spigot - which is why reflection is required),
and I preserve the same thread-safety it provides in my fork - so this solution
will not break compatibility with craftbukkit, spigot, or paper and fixes my issue.

## Checklist

- [X] I read and followed the [contribution guidelines](https://github.com/IntellectualSites/FastAsyncWorldEdit-1.13/blob/1.15/CONTRIBUTING.md)

### DIFF
--- a/worldedit-bukkit/src/main/java/com/boydti/fawe/bukkit/adapter/mc1_14/BukkitAdapter_1_14.java
+++ b/worldedit-bukkit/src/main/java/com/boydti/fawe/bukkit/adapter/mc1_14/BukkitAdapter_1_14.java
@@ -12,6 +12,9 @@ import com.boydti.fawe.util.TaskManager;
 import com.sk89q.worldedit.world.block.BlockState;
 import com.sk89q.worldedit.world.block.BlockTypesCache;
 import io.papermc.lib.PaperLib;
+import java.lang.invoke.MethodHandle;
+import java.lang.invoke.MethodHandles;
+import java.lang.reflect.Method;
 import java.util.concurrent.locks.ReentrantLock;
 import net.jpountz.util.UnsafeUtils;
 import net.minecraft.server.v1_14_R1.Block;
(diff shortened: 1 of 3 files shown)


IronApollo:
Has this been tested in-game with FAWE and your fork of Paper? I'm curious on the difference between the functionality of the visibleChunks field and PlayerChunkMap#getVisibleChunk, if any exist, as well as any noticeable performance differences.

I'll test this when I get home and, if all seems well, merge.

Spottedleaf:
Yes - I've tested on my fork & paper.

There should be no difference between the two - on paper and on my fork, at least in terms of behaviour.

The new style I use for updating the map is to store 3 maps internally - one representing "updating", another representing queued changes, and another representing "visible"
The util can be viewed here: https://github.com/PaperMC/Paper/blob/4e0ed1f812760e6c39dd0dcfbb80feaaa159452d/Spigot-Server-Patches/0004-MC-Utils.patch#L262
The basic design is that changes are queued to the visible map (like how vanilla does it), but instead of updating the visible map by cloning, it acquires a seqlock and then performs the updates.

So threads not the main thread must spin on the seqlock when reading. Given most usages are reading, the seqlock generally should not be contended at all - it beats a `synchronized` approach because threads do not block each other when reading and will only incur a synchronization cost when necessary. 

The internal visible map couldn't replace the visibleChunks field since even though it's locked internally that wont help it when iterating.

Spottedleaf:
Updated

PR #382, opened 2020-03-29
Fixing that the PlayerDropItemEvent was not fired

ForceUpdate1:
The event was not fired 🙌


IronApollo:
Good catch. Thanks for the PR!

PR #383, opened 2020-03-29
Fixing wrong translation keys

ForceUpdate1:
I was referring to the config "strings.json"

`
"fawe.error.no-perm": "You are lacking the permission node: {0}"
`

`
"worldedit.tool.navwand.equip": "Navigation wand bound to {0}."
`


IronApollo:
Thanks for your PR!

PR #492, opened 2020-05-29
Add warning to config regarding disabling history on disk

aurorasmiles:
Because @dordsor21 found out that in the newer version disabling history on disk can improve performance but is generally unsafe.

Because so far this "issue" is only documented in Discord, I think a warning in the config wouldn't hurt

### DIFF
--- a/worldedit-core/src/main/java/com/boydti/fawe/config/Settings.java
+++ b/worldedit-core/src/main/java/com/boydti/fawe/config/Settings.java
@@ -176,6 +176,7 @@ public class Settings extends Config {
                 " - Persists restarts",
                 " - Unlimited undo",
                 " - Does not affect edit performance if `combine-stages`",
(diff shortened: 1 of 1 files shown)


dordsor21:
Eh I mean it's not really considered unsafe. It's how history is stored by default in WorldEdit. What I meant to say was unsafe is having history continue to write to disk after the edit (i.e. still being stored in RAM). But since this doesn't happen in 1.14 (thus why it's slower) it's not something people need to worry about.

aurorasmiles:
Maybe "is less safe"?

aurorasmiles:
I removed the safety part and just added "Disabling this might improve performance"

dordsor21:
I think it would be better to not add anything; people will set a setting and then complain either way if we talk speed vs safe vs whatever.

I'm going to look at a "post-process" history option to allow history to be written after the edit has completed in order to be feature-equivalent with legacy FAWE, which includes the explanation of both modes.

PR #494, opened 2020-05-30
Fix a few links

aurorasmiles:
Update a few old links.

### DIFF
--- a/worldedit-bukkit/src/main/java/com/boydti/fawe/bukkit/regions/plotsquared/PlotSquaredFeature.java
+++ b/worldedit-bukkit/src/main/java/com/boydti/fawe/bukkit/regions/plotsquared/PlotSquaredFeature.java
@@ -47,7 +47,7 @@ public class PlotSquaredFeature extends FaweMaskManager {
                 setupSchematicHandler();
                 setupRegionManager();
             } catch (Throwable ignored) {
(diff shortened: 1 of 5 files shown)


worldedit-bukkit/src/main/java/com/boydti/fawe/bukkit/regions/plotsquared/PlotSquaredFeature.java
```diff
@@ -47,7 +47,7 @@ public PlotSquaredFeature() {
                 setupSchematicHandler();
                 setupRegionManager();
             } catch (Throwable ignored) {
-                log.debug("Please update PlotSquared: http://ci.athion.net/job/PlotSquared/");
+                log.debug("Please update PlotSquared");
```
aurorasmiles, line 50: Since there is no (public) CI for PlotSquared anymore, I just removed the link.

N0tMyFaultOG: But we can add for p2 v5 on spigot.

aurorasmiles: Right, added the link to the spigot page

Issue #437, opened 2020-05-01
Simplex Comeback


TheoRetisch1:
**Is your feature request related to a problem? Please describe.**
I am always frustrated when I can only achieve a reasonable distribution under high effort and think back to the good old simplex command.

**Describe the solution you'd like**
I would be very happy if the simplex distribution would be available again as pattern and mask. It was a great tool for texturing and flowering. I think the functionality of simplex doesn't need to be explained here again, because it was sufficiently implemented in the previous versions.

**Describe alternatives you've considered**
-

**Additional context**
Keep up your great work :)

MattBDev:
I said yesterday on Discord I am working on it

TheoRetisch1:
Ok awesome thank you! Shoudl I close this?

MattBDev:
No it's fine. Let's leave it open until I push the changes. I found a bug in the simplex math I need to fix in addition to re-adding the masks/patterns

TheoRetisch1:
I just want to add that it would be pretty cool to also use simplexes in simplexes again. Example: //br rock #simplex[50][50][0][#simplex[30][159:1,159:2,159:3,159:4,159:5]] 20. this allows a less constant  = more natural distribution.
If you need an overview of how and for what the simplex command was used, you can find a nice illustrated guide from Mulcho here: https://imgur.com/a/rwVAE

SirYwell:
As of build 268, the simplex pattern *should* work again. It needed some basic changes on how commands are parsed so please report if any other commands aren't **accepted** anymore, please let me know.

PR #520, opened 2020-06-27
Start reimplementation of simplex pattern

SirYwell:
## Overview
Would partly cover #437 (the simplex mask isn't included)

### Short version: 
I've added a SimplexPatternParser which parses `#simplex[scale][pattern]`.
### Long version: 
As FAWE has multiple patterns/masks that require additional arguments (provided in [] normally), I've added the class `RichParser` which allows to easily parse those arguments.  Other parsers can extend it, they only need to provide the name of the pattern or mask. They then need to implement the following methods:
##### - `protected Stream<String> getSuggestions(String argument, int index)` 
(maybe that should be optional) For suggestions of the argument at the given index (e.g. for the simplex pattern, it will suggest numbers as first (index 0) argument and patterns as second (index 1) argument.

##### - `protected Pattern parseFromInput(String[] arguments, ParserContext context)`
Parses the actual pattern or mask with the given arguments. The arguments can be validated there.

That way, the parsers don't need to handle the extraction of arguments as this is done by the RichParser (If you have a better name for it, please let me know).

~This is a draft as it's not ready to merge. While the simplex pattern already works, there are still code parts that need to be discussed. I'll comment those issues.~

## Checklist
- [ ] I ensured my changes do not break other parts of the code
- [x] I read and followed the [contribution guidelines](https://github.com/IntellectualSites/FastAsyncWorldEdit/blob/1.15/CONTRIBUTING.md)

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/util/StringUtil.java
+++ b/worldedit-core/src/main/java/com/sk89q/util/StringUtil.java
@@ -20,6 +20,7 @@
 package com.sk89q.util;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.List;
 import java.util.Locale;
@@ -330,5 +331,38 @@ public final class StringUtil {
         }
 
         return parsableBlocks;
+    }
+
+    /**
+     * Splits a string respecting enclosing quotes.
+     *
+     * @param input the input to split.
+     * @param delimiter the delimiter to split on.
+     * @param open the opening quote character.
+     * @param close the closing quote character.
+     * @return a list of split strings.
+     */
+    public static List<String> split(String input, char delimiter, char open, char close) {
+        if (input.indexOf(open) == -1 && input.indexOf(close) == -1) {
+            return Arrays.asList(input.split(String.valueOf(delimiter)));
+        }
+        int level = 0;
+        int begin = 0;
+        List<String> split = new ArrayList<>();
+        for (int i = 0; i < input.length(); i++) {
+            char c = input.charAt(i);
+            if (c == delimiter && level == 0) {
+                split.add(input.substring(begin, i));
+                begin = i + 1;
+            } else if (c == open) {
+                level++;
+            } else if (c == close) {
+                level--;
+            }
+        }
+        if (begin < input.length()) {
+            split.add(input.substring(begin));
+        }
+        return split;
     }
 }
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/extension/factory/parser/RichParser.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/extension/factory/parser/RichParser.java
@@ -58,10 +58,6 @@ public abstract class RichParser<E> extends InputParser<E> {
     public E parseFromInput(String input, ParserContext context) throws InputParseException {
         if (!input.startsWith(this.prefix)) return null;
         if (input.length() < this.prefix.length()) return null;
-        // if it's not "atomic", we can't parse it like that
-        if (StringUtil.split(input, ',', '[', ']').size() > 1) return null;
-        // TODO if no arguments -> exception
-        // TODO if not starts with [ and ends with ] -> exception
         String[] arguments = extractArguments(input.substring(prefix.length()), true);
         return parseFromInput(arguments, context);
     }
@@ -78,7 +74,7 @@ public abstract class RichParser<E> extends InputParser<E> {
     /**
      * Parses the already split arguments.
      *
-     * @param arguments the array of arguments that were split.
+     * @param arguments the array of arguments that were split (can be empty).
      * @param context   the context of this parsing process.
      * @return the resulting parsed type.
      * @throws InputParseException if the input couldn't be parsed correctly.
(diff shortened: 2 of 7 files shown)


worldedit-core/src/main/java/com/sk89q/util/StringUtil.java
```diff
@@ -331,4 +332,37 @@ public static int getLevenshteinDistance(String s, String t) {
...
+    /**
+     * Splits a string respecting enclosing quotes.
+     *
+     * @param input the input to split.
+     * @param delimiter the delimiter to split on.
+     * @param open the opening quote character.
+     * @param close the closing quote character.
+     * @return a list of split strings.
+     */
+    public static List<String> split(String input, char delimiter, char open, char close) {
```
SirYwell, line 345: This one is most likely to break other things. WE uses the method above this one to parse strings but it doesn't work with nested [] brackets. That broke using simplex in simplex

worldedit-core/src/main/java/com/sk89q/worldedit/function/pattern/RandomPattern.java
```diff
@@ -52,6 +52,19 @@ public RandomPattern(SimpleRandom random) {
         this.random = random;
     }
 
+    /**
+     * Create a random pattern from an existing one but with a different random.
+     *
+     * @param random the new random to use.
+     * @param parent the existing random pattern.
+     */
+    public RandomPattern(SimpleRandom random, RandomPattern parent) {
```
SirYwell, line 61: We either need this one or make the random non-final, reassign it and recreate the RandomCollection with it

worldedit-core/src/main/java/com/sk89q/worldedit/extension/factory/parser/pattern/SimplexPatternParser.java
```diff
@@ -0,0 +1,73 @@
...
+            throw new InputParseException("Invalid amount of arguments. Syntax: #simplex[scale][pattern]");
+        }
+        double scale = Double.parseDouble(arguments[0]);
+        scale = 1d / Math.max(1, scale);
+        Pattern inner = worldEdit.getPatternFactory().parseFromInput(arguments[1], context);
+        if (inner instanceof RandomPattern) {
+            return new RandomPattern(new SimplexRandom(scale), (RandomPattern) inner);
+        }
+        // TODO handle RandomStatePattern?
+        // TODO for other patterns, either warn actor (inner pattern will be returned) or throw exception
```
SirYwell, line 56: There might be patterns a simplex can't handle correctly, so we can either just return that pattern (means nothing happens) or throw an exception so the user knows their input was invalid

worldedit-core/src/main/java/com/sk89q/worldedit/extension/factory/parser/pattern/SimplexPatternParser.java
```diff
@@ -0,0 +1,73 @@
...
+        if (index == 1) {
+            return worldEdit.getPatternFactory().getSuggestions(argument).stream();
+        }
+        return Stream.empty();
+    }
+
+    @Override
+    protected Pattern parseFromInput(String[] arguments, ParserContext context) {
+        if (arguments.length != 2) {
+            // TODO
```
SirYwell, line 46: I'm not familiar with the way WE handles stuff like syntax error messages, but I guess there's a better way than that

worldedit-core/src/main/java/com/sk89q/worldedit/extension/factory/parser/RichParser.java
```diff
@@ -0,0 +1,91 @@
...
+        if (StringUtil.split(input, ',', '[', ']').size() > 1) return null;
+        // TODO if no arguments -> exception
+        // TODO if not starts with [ and ends with ] -> exception
+        String[] arguments = extractArguments(input.substring(prefix.length()), true);
+        return parseFromInput(arguments, context);
+    }
+
+    protected abstract Stream<String> getSuggestions(String argument, int index);
+
+    protected abstract E parseFromInput(String[] arguments, ParserContext context);
```
MattBDev, line 58: `@NotNull` annotation is needed for the parameter `arguments`. (Use the JetBrains one)

worldedit-core/src/main/java/com/sk89q/worldedit/extension/factory/parser/pattern/SimplexPatternParser.java
MattBDev: If the length of the input doesn't match the length of the format required then an `InputParseException` should be thrown.
An example would be the following code:
`throw new InputParseException("Clipboard offset needs x,y,z coordinates.");`

worldedit-core/src/main/java/com/sk89q/worldedit/extension/factory/parser/pattern/SimplexPatternParser.java
MattBDev: Use your best judgement here and we can adjust based on feedback.

worldedit-core/src/main/java/com/sk89q/worldedit/function/pattern/RandomPattern.java
MattBDev: Do whatever is easier here.

worldedit-core/src/main/java/com/sk89q/worldedit/extension/factory/parser/pattern/SimplexPatternParser.java
SirYwell: It's returning inner for `BlockStateHolder`s now as that's single blocks anyways if I got it right. For other patterns than that or RandomPattern, it throws an exception. I played around a lot but if I missed something (that actually makes sense to use in simplex) people can come up with that.

PR #568, opened 2020-08-15
Fixed NullPointerException in Spline and Sweep brush

HazelTheWitch:
## Overview

**Fixes #{issue number}**

## Checklist

- [] I included all information required in the sections above
- [] I tested my changes and approved their functionality
- [] I ensured my changes do not break other parts of the code
- [] I read and followed the [contribution guidelines](https://github.com/IntellectualSites/FastAsyncWorldEdit/blob/1.16/CONTRIBUTING.md)

### DIFF
--- a/worldedit-core/src/main/java/com/boydti/fawe/object/brush/sweep/SweepBrush.java
+++ b/worldedit-core/src/main/java/com/boydti/fawe/object/brush/sweep/SweepBrush.java
@@ -46,7 +46,7 @@ public class SweepBrush implements Brush, ResettableTool {
             return;
         }
 
-        boolean newPos = !position.equals(this.position);
+        boolean newPos = this.position == null || !position.equals(this.position);
         this.position = position;
         Player player = editSession.getPlayer();
         if (player == null) {
--- a/worldedit-core/src/main/java/com/boydti/fawe/object/visitor/DFSVisitor.java
+++ b/worldedit-core/src/main/java/com/boydti/fawe/object/visitor/DFSVisitor.java
@@ -104,7 +104,7 @@ public abstract class DFSVisitor implements Operation {
                         from.getZ() + direction.getZ());
                 if (isVisitable(bv, bv2)) {
                     Node adjacent = new Node(bv2.getBlockX(), bv2.getBlockY(), bv2.getBlockZ());
-                    if (!adjacent.equals(current.from)) {
+                    if (current.from == null || !adjacent.equals(current.from)) {
                         AtomicInteger adjacentCount = visited.get(adjacent);
                         if (adjacentCount == null) {
                             if (countAdd++ < maxBranch) {


SirYwell:
While it is solved like this in multiple places, from looking at it again I'd suggest to properly implement the `equals` methods. In BlockVector3, the overloaded equals(BlockVector3) method should at least allow null values in my opinion, as it's super unintuitive otherwise. 
Also note: This won't really fix the sweep brush, I worked on it a while ago but as I didn't had a proper solution, I kinda gave up on it. Rotation and pasting need a rework too (https://github.com/IntellectualSites/FastAsyncWorldEdit/commit/41b2a3e2be2714e247a5a7de227226dde42d8394). My best result was https://cdn.discordapp.com/attachments/707591960088543252/728760815880110080/unknown.png but it threw IndexOutOfBoundsExceptions most cases

HazelTheWitch:
I just changed equals and its references, might want to double check all the references in case I changed any of their behavior. I don't think I did but cant hurt.

HazelTheWitch:
Part of the issue with sweep was a double being casted to int, instead of rounded, which meant float rounding errors would shift paste sections an entire block over.

I also changed the algorithm deciding when to place the pastes in `SweepBrush`. It now calculates the left hand and right hand sides of the paste and adds a new paste whenever either side moves off the previous block. It should be fairly fast to calculate, although I didn't try to optimize it. [Here's an example of the results I'm getting with it.](https://cdn.discordapp.com/attachments/566822188041109515/744303737472024706/unknown.png)

An issue is the interpolation used (`KochanekBartelsInterpolation`) is non linear, so placing the points along the spline unequally can have [negative impacts on the result](https://cdn.discordapp.com/attachments/566822188041109515/744304476596338728/unknown.png). You can correct for it by splitting the spline into each section and doing them all individually, is there a better way to do that?

HazelTheWitch:
I've reworked the placement algorithm to use `ReparametrisingInterpolation` so now the sweep brush should be much more continuous with less gaps. Here are some examples of the progress, [from left to right, rightmost being the final version.](https://cdn.discordapp.com/attachments/566822188041109515/744468495923675136/unknown.png)

SirYwell:
That looks promising already, but somehow air still replaces blocks if I see this right?

HazelTheWitch:
I'm not sure if air blocks are being pasted in. There is an issue with the brush where if the selection is longer in the x axis than z it won't past correctly, [shown here](https://cdn.discordapp.com/attachments/566822188041109515/744711858434474045/unknown.png). As you can see the left result is clearly much better, although still really messy the general form is there. The left one though almost doesn't seem to rotate.

I plan on seeing about different options for the brush and maybe implementing a different spline to smooth out the left-hand side, but I want to get the right fixed first. I've tried everything and I can think of, but it's persistent. If you have any ideas let me know.

HazelTheWitch:
I can't get anywhere with this, I think there's some issue with the Affine math or something with pasting. In its current state, the brush works well with clipboards longer in the Z direction, but X aligned selections almost don't rotate? It's a weird bug, but I can't seem to find a cause.

> That looks promising already, but somehow air still replaces blocks if I see this right?

What you're seeing is either an issue with the spline not being linear in t which I fixed using a reparameterized spline. Or artifacts from the forward copy. In my opinion, a backward copy would be better, but I can't seem to find how to use that. It would reduce the number of pastes required, and fix some of the artifacts.

Right now the only major issue I can see is the X-Z thing. It does fully fix spline, which is broken in the current build I believe.

SirYwell:
sweepbrush
I just tested your PR and it works way better than the stuff I came along with. And I agree, I believe there is something wrong with the math.
On the upper images, I selected my points from left to right the first time and from right to left the second time.
I don't know why it's weird on the one that's just up and down and for the circle, I'm especially confused as the lower layer looks fine but the upper one looks like blocks where replaced with air by further parts.

If you don't want to spend even more time on it, I'd approve this PR but if you want to investigate the remaining issues, feel free to do so :P

HazelTheWitch:
I am completely stuck, so this is probably where it will be for a while, feel free to approve it.

SirYwell (approved):

PR #569, opened 2020-08-15
Feature/better noise

SirYwell:
## Description

I wanted to play around with the noises provided by WE/JLibNoise and added them as pattern similar to the simplex one.
Don't know if they are interesting enough to bring it into FAWE tho.
grafik
This are a few examples how it looks like when used. The notation is similar to simplex, like `#rmf[scale][pattern]`, `#perlin[scale][pattern]`, and `#voronoi[scale][pattern]`. 

Other than simplex, the generated noise is different per generator instance, I'm not exactly sure why as the seed seems to be 0 by default. Might be worth to investigate that point.

## Checklist


N0tMyFaultOG:
That looks lit 🔥

worldedit-core/src/main/java/com/boydti/fawe/object/collection/FastRandomCollection.java
```diff
@@ -5,28 +5,41 @@
...
+                return Optional.empty();
             }
             if (weight > max) {
                 max = weight;
             }
         }
         int gcd = MathMan.gcd(counts);
         if (max / gcd > 100000) {
-            throw new IllegalArgumentException("Too large");
+            return Optional.empty();
```
aurorasmiles, line 42: Instead of just returning empty, maybe a message here would be nice?

SirYwell: I tried to improve the previous implementation. From what I've seen, the exceptions where thrown to exit the construtor because the object couldn't be created and the exception was basically ignored. `FastRandomCollection` should be used internally by `RandomCollection` only, means the API is `RandomCollection.of(...)` and it returns the implementation based on the given values. Why the `FastRandomCollection` couldn't be used isn't relevant to the user.

That said, it might be worth to point out proper usage of the RandomCollection API. The reason I didn't documented the cases in which a FastRandomCollection can be created is that I didn't deeply looked into the algorithm (I'm especially confused by the `100000` value but who knows why it is like that).

aurorasmiles: Ok, in that case it's probably fine.

TheoRetisch1:
Thank you for this awesome feature :)

PR #586, opened 2020-08-26
Use the proper class types for the NMS code in 1.16.2

MattBDev:
This is untested but my changes remove two catch statements and rename a variable with a Mojang mapping name. It uses a bit less reflection so it should speed things up a tiny bit but it likely isn't noticable. 
I am submitting this as a PR to allow for people to test my changes so another broken 1.16.2 update doesn't go out.

### DIFF
--- a/worldedit-bukkit/src/main/java/com/boydti/fawe/bukkit/adapter/mc1_16_2/BukkitAdapter_1_16_2.java
+++ b/worldedit-bukkit/src/main/java/com/boydti/fawe/bukkit/adapter/mc1_16_2/BukkitAdapter_1_16_2.java
@@ -39,12 +39,10 @@ import java.lang.invoke.MethodHandle;
 import java.lang.invoke.MethodHandles;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.Field;
-import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Map;
-import java.util.Set;
 import java.util.concurrent.CompletableFuture;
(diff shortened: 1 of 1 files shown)


worldedit-bukkit/src/main/java/com/boydti/fawe/bukkit/adapter/mc1_16_2/BukkitAdapter_1_16_2.java
```diff
@@ -190,20 +188,25 @@ public static void sendChunk(WorldServer nmsWorld, int chunkX, int chunkZ, int m
...
                     if (Arrays.stream(dirtyblocks).allMatch(e -> e == null || e.isEmpty())) {
                         nmsWorld.getChunkProvider().playerChunkMap.a(playerChunk);
                     }
                     for (int i = 0; i < 16; i++) {
-                        if (dirtyblocks[i] == null) dirtyblocks[i] = (Set<Short>) shortArraySetConstructor.newInstance();
-                        for (int x = 0; x < 16; x++)
-                            for (int y = 0; y < 16; y++)
-                                for (int z = 0; z < 16; z++)
+                        if (dirtyblocks[i] == null) {
+                            dirtyblocks[i] = new ShortArraySet();
```
aurorasmiles, line 197: This doesn't work, since WorldEdit shading overwrites the correct class for the ShortArraySet. That's why *had* to use the Reflection. https://waste.cutiepie.at/0p8rj7p0omsbnc8

aurorasmiles (changes requested):

MattBDev:
I didn't realize this but I am glad I PR'ed this before pushing

Issue #584, opened 2020-08-26
strings.json changes no longer showing in build #325


SecretlyJealous:
# Bug report for FastAsyncWorldEdit 1.16.1

**[REQUIRED] FastAsyncWorldEdit Configuration Files:** 
config.yml: https://pastebin.com/qbx642wn
config-legacy.yml: https://pastebin.com/E4qmuBMb

**[REQUIRED] FastAsyncWorldEdit Version Number:** 
FastAsyncWorldEdit-1.16-325

**[REQUIRED] Spigot/Paper Version Number:** 
git-Paper-131 (MC: 1.16.1) (Implementing API version 1.16.1-R0.1-SNAPSHOT)

**[REQUIRED] Description of the problem:**
When using build 325 the strings.json file doesn't show your own custom changes, I've changed the prefix to "WorldEdit":

```
{
	"prefix": "&8(&4&lWorldEdit&8)&7 {0}",

	"fawe.worldedit.history.find.element": "&2{0} {1} &7ago &3{2}m &6{3} &c/{4}",
	"fawe.worldedit.history.find.element.more": " - Changes: {0}\n - Bounds: {1} -> {2}\n - Extra: {3}\n - Size on Disk: {4}",
	"fawe.worldedit.history.find.hover": "{0} blocks changed, click for more info",
```

However in game it doesn't show any changes I make to this file:
![2020-08-26_13 47 27](https://user-images.githubusercontent.com/10485646/91305261-b2e08c00-e7a2-11ea-8cf3-0e3b2cec593a.png)

**How to replicate**:
Update to build FastAsyncWorldEdit-1.16-325.

**Plugins being used on the server:**
FastAsyncWorldEdit, LuckPerms, PlotSquared, ProtocolLib

**Checklist**:

- [X] I included all information required in the sections above
- [X] I made sure there are no duplicates of this report [(Use Search)](https://github.com/IntellectualSites/FastAsyncWorldEdit/issues?q=is%3Aissue)
- [X] I made sure I am using an up-to-date version of [FastAsyncWorldEdit for 1.16.1](https://ci.athion.net/job/FastAsyncWorldEdit-1.16/)
- [X] I made sure the bug/error is not caused by any other plugin

mibby:
Can also confirm experiencing this with FAWE dev 326. 

[...]

PR #591, opened 2020-08-29
fix filepath in Local Configuration

aurorasmiles:
Fix issue with the schematics folder suddenly generating inside the server home instead of plugin home


aurorasmiles:
could also fix #584

SirYwell (approved):

N0tMyFaultOG (approved):

PR #599, opened 2020-09-06
Only send PacketPlayOutChunkMap for chunk updates

aurorasmiles:
Improves performance and fixes issues with PacketPlayOutChunkMap when using Spigot

### DIFF
--- a/worldedit-bukkit/src/main/java/com/boydti/fawe/bukkit/adapter/mc1_16_2/BukkitAdapter_1_16_2.java
+++ b/worldedit-bukkit/src/main/java/com/boydti/fawe/bukkit/adapter/mc1_16_2/BukkitAdapter_1_16_2.java
@@ -9,6 +9,7 @@ import com.boydti.fawe.object.collection.BitArrayUnstretched;
 import com.boydti.fawe.util.MathMan;
 import com.boydti.fawe.util.ReflectionUtils;
 import com.boydti.fawe.util.TaskManager;
+import com.mojang.datafixers.util.Either;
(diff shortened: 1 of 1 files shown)


dordsor21 (commented):
Can this method only be used with 1.16? We should make the change to all adapters if not

aurorasmiles:
> 
> 
> Can this method only be used with 1.16? We should make the change to all adapters if not

What method are you refering to? 1.16.2 changed a bit of NMS stuff, which is why we initially needed those changes for 1.16.2 in the first place

dordsor21:
> > Can this method only be used with 1.16? We should make the change to all adapters if not
> 
> What method are you refering to? 1.16.2 changed a bit of NMS stuff, which is why we initially needed those changes for 1.16.2 in the first place

The one that you've switched to for sending chunk packets. Perhaps not necessarily a method change, but switching to the more performant method of sending updates only when needed should be done to all adapters if possible.

IronApollo:
Does the crash happen for users of Spigot 1.16.1, 1.16, 1.15, etc? If it does, then the fix should likely be ported to the respective adapters. If it doesn't and users are not crashing on the Spigot platform on those versions, no port should be needed. This does need to be clarified before the PR is approved from my end!

Thanks for your work as always though :)

aurorasmiles:
The crash is only on Spigot 1.16.2 because of the nms changes

IronApollo (approved):
If the issue only occurs on 1.16.2, this should be a finalized PR. Thanks again

dordsor21:
If it's a performance boost/general correct way of doing it, then it should be on all versions. But okay.

PR #607, opened 2020-09-10
More work on masks

MattBDev:
This PR brings FAWE more in line with upstream WorldEdit. In January of this year Jesse changes masks to be "stateless" and then disappeared without finishing the work or explaining why he did this. This PR reverts those changes. There is some minor additional work to bring us in line with upstream too.

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/extension/factory/parser/mask/AirMaskParser.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/extension/factory/parser/mask/AirMaskParser.java
@@ -1,10 +1,31 @@
+/*
+ * WorldEdit, a Minecraft world manipulation toolkit
+ * Copyright (C) sk89q <http://www.sk89q.com>
+ * Copyright (C) WorldEdit team and contributors
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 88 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/function/mask/Mask.java
```diff
@@ -94,16 +77,6 @@ default Mask optimize() {
...
-    default Mask and(Mask other) {
-        Mask value = and(other);
-        return value == null ? MaskIntersection.of(this, other) : value;
-    }
-
-    default Mask or(Mask other) {
-        Mask value = or(other);
-        return value == null ? MaskUnion.of(this, other) : value;
-    }
-
```
MattBDev, line 106: This code wasn't that great and we should encourage developers to use Intersection and Union API's. If we want to provide an easier way for developers to create an Intersection and Union then a new issue should be made detailing what a good API needs.

worldedit-core/src/main/java/com/sk89q/worldedit/extension/factory/parser/mask/AirMaskParser.java
```diff
@@ -1,10 +1,31 @@
...
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
```
MattBDev, line 18: Added since our class is now matches upstream.

N0tMyFaultOG (approved):

IronApollo (approved):

aurorasmiles:
Fixes #305

aurorasmiles (approved):

PR #611, opened 2020-09-12
Begin working on improved lighting to better match 1.14's "new" light…

dordsor21:
Figured this ought to be a PR cos the scope is relatively large now (effectively a complete rewrite of the lighting logic will be done)


dordsor21:
Isn't _really_ that major

PR #658, opened 2020-09-23
Allow "post processing" of chunks

dordsor21:
- restores legacy capability to continue saving edits in the background after sending the chunks
 - speeds up the edit clientside
 - Makes edits sort of 2.5x faster based on testing (and possibly more fasterer ig)
 - nail in the coffin of the terrible and statically coded coreedit
 - We should totally make IronGolem work so Core* is no longer used by anyone

### DIFF
--- a/worldedit-core/src/main/java/com/boydti/fawe/config/Settings.java
+++ b/worldedit-core/src/main/java/com/boydti/fawe/config/Settings.java
@@ -395,7 +395,7 @@ public class Settings extends Config {
         public boolean ALLOW_TICK_EXISTING = true;
         @Comment({
             "Do not wait for a chunk's history to save before sending it",
-            " - Any following history commands (undo/redo) will still need to wait for history to be saved",
+            " - Undo/redo commands will wait until the history has been written to disk before executing",
             " - Requires combine_stages = true"
         })
         public boolean SEND_BEFORE_HISTORY = true;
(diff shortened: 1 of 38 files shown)


worldedit-core/src/main/java/com/boydti/fawe/config/Settings.java
```diff
@@ -393,6 +393,12 @@
             "This will increase time taken slightly."
         })
         public boolean ALLOW_TICK_EXISTING = true;
+        @Comment({
+            "Do not wait for a chunk's history to save before sending it",
+            " - Any following history commands (undo/redo) will still need to wait for history to be saved",
+            " - Requires combine_stages = true"
+        })
+        public boolean SEND_BEFORE_HISTORY = true;
```
N0tMyFaultOG, line 401: Maybe we can rephrase that a bit? I don't think every end user knows what it means.

N0tMyFaultOG (changes requested):

worldedit-core/src/main/java/com/boydti/fawe/config/Settings.java
dordsor21: "Undo/redo commands will wait until the history has been written to disk before executing" ?

N0tMyFaultOG: That sounds better.

N0tMyFaultOG (approved):

IronApollo (approved):

PR #665, opened 2020-09-27
fix performance issues in BukkitPlayer permission

aurorasmiles:
Apparently we introduced some heavy performance issues with our recent changes, this should fix those

### DIFF
--- a/worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitPlayer.java
+++ b/worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitPlayer.java
@@ -73,7 +73,6 @@ import javax.annotation.Nullable;
 
 public class BukkitPlayer extends AbstractPlayerActor {
 
-    private static final PermAttachmentManager permAttachmentManager = new PermAttachmentManager(WorldEditPlugin.getInstance());
     private final Player player;
     private final WorldEditPlugin plugin;
     private final PermissionAttachment permAttachment;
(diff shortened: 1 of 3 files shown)


worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitPlayer.java
```diff
@@ -67,26 +67,28 @@
...
+import java.util.concurrent.locks.ReentrantReadWriteLock;
 
 import javax.annotation.Nullable;
 
 public class BukkitPlayer extends AbstractPlayerActor {
 
-    private Player player;
-    private WorldEditPlugin plugin;
-    private PermissionAttachment permAttachment;
+    private static final PermAttachmentManager permAttachmentManager = new PermAttachmentManager(WorldEditPlugin.getInstance());
```
dordsor21, line 76: is it not better worth having this in the WorldEditPlugin itself, and the manager itself being a separate class, and then accessing with WorldEditPlugin.getInstance().getPermAttachmentManager() ?

aurorasmiles: I don't think it makes a huge difference since you don't really use it anywhere else I think?

weaondara: i don't, we only ever use it within the class and the perm attachment is not accessible from the outside

MattBDev: I understand the reluctance to call `WorldEditPlugin.getInstance()` but considering that this is for a Bukkit wrapper class I think this is fine.

aurorasmiles: done

dordsor21 (approved):

SirYwell (approved):

PR #673, opened 2020-09-30
Fixed a NullPointerException

TheBusyBiscuit:
## Overview
This Pull Requets adresses a `NullPointerException` in the `BukkitPlayer` class of `worldedit-bukkit`.
The exception can be seen here: https://pastebin.com/dq72vmsD

This exception is caused due to an unhandled Player variable which is set as nullable by WorldGuard.
The original WorldEdit handles this nullability correctly, however it seems that FAWE does not, so that's why I am proposing this change.

## Description

When you call `WorldGuardPlugin#wrapOfflinePlayer(OfflinePlayer)` in WorldGuard as seen here: [WorldGuardPlugin.java#L450](https://github.com/EngineHub/WorldGuard/blob/master/worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/WorldGuardPlugin.java#L450)
it will create a new `BukkitOfflinePlayer` (an extension of `BukkitPlayer`).

This will pass a `Player` object to the super constructor, while there is no annotation or documentation about this actually being nullable. You can find a comment in the code of WorldGuard: [BukkitOfflinePlayer.java#L42](https://github.com/EngineHub/WorldGuard/blob/master/worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/BukkitOfflinePlayer.java#L42) explaining that it can in fact be null.

This leads us to FAWE's `BukkitPlayer` constructor:
https://github.com/IntellectualSites/FastAsyncWorldEdit/blob/main/worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitPlayer.java#L81
Here the method `getExistingMap()` is called for a nullable Player variable which is then passed on to other methods which expect it to be non-null.

This PR adds a simple null-check to the `getExistingMap()` method and a few javadocs comments that should raise awareness about this nullability for the future.
On that regard I have also added an `@Nullable` annotation to this parameter.

## Checklist

### DIFF
--- a/worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitPlayer.java
+++ b/worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitPlayer.java
@@ -82,7 +82,7 @@ public class BukkitPlayer extends AbstractPlayerActor {
      * 
      * @param player The corresponding {@link Player} or null if they are offline
      */
-    public BukkitPlayer(Player player) {
+    public BukkitPlayer(@Nullable Player player) {
         super(getExistingMap(WorldEditPlugin.getInstance(), player));
         this.plugin = WorldEditPlugin.getInstance();
         this.player = player;
@@ -112,7 +112,7 @@ public class BukkitPlayer extends AbstractPlayerActor {
      * 
      * @return A data {@link Map} for that {@link Player}
      */
-    private static Map<String, Object> getExistingMap(WorldEditPlugin plugin, Player player) {
+    private static Map<String, Object> getExistingMap(WorldEditPlugin plugin, @Nullable Player player) {
         if (player == null) {
             // Null cannot have any data, so we just return an empty Map
             return new ConcurrentHashMap<>();
--- a/worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitPlayer.java
+++ b/worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitPlayer.java
@@ -77,6 +77,11 @@ public class BukkitPlayer extends AbstractPlayerActor {
     private final WorldEditPlugin plugin;
     private final PermissionAttachment permAttachment;
 
+    /**
+     * This constructs a new {@link BukkitPlayer} for the given {@link Player}.
+     * 
+     * @param player The corresponding {@link Player} or null if they are offline
+     */
     public BukkitPlayer(Player player) {
         super(getExistingMap(WorldEditPlugin.getInstance(), player));
         this.plugin = WorldEditPlugin.getInstance();
@@ -84,6 +89,12 @@ public class BukkitPlayer extends AbstractPlayerActor {
         this.permAttachment = plugin.getPermissionAttachmentManager().addAttachment(player);
     }
 
+    /**
+     * This constructs a new {@link BukkitPlayer} for the given {@link Player}.
+     * 
+     * @param plugin The running instance of {@link WorldEditPlugin}
+     * @param player The corresponding {@link Player} or null if they are offline
+     */
     public BukkitPlayer(WorldEditPlugin plugin, Player player) {
         this.plugin = plugin;
         this.player = player;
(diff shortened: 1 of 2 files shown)


worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitPlayer.java
```diff
@@ -77,13 +77,24 @@
...
+    /**
+     * This constructs a new {@link BukkitPlayer} for the given {@link Player}.
+     * 
+     * @param player The corresponding {@link Player} or null if they are offline
+     */
+    public BukkitPlayer(@Nullable Player player) {
         super(getExistingMap(WorldEditPlugin.getInstance(), player));
         this.plugin = WorldEditPlugin.getInstance();
         this.player = player;
         this.permAttachment = plugin.getPermissionAttachmentManager().addAttachment(player);
```
dordsor21, line 89: This will shift any NPEs into the PermissionAttachmentManager.

dordsor21 (changes requested):

worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitPlayer.java
TheBusyBiscuit: Ah I see, should I add another null check for the PermissionAttachmentManager? (Although the parameter being nullable feels a bit off)
Or should I add another if-statement to this assignment and make `permAttachment` nullable as a result of that?

dordsor21: Within the call to super constructor, https://github.com/IntellectualSites/FastAsyncWorldEdit/blob/main/worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/WorldEditPlugin.java#L518 is also called ultimately, which will also NPE on the player.

It's worth noting the default constructor in AbstractPlayerActor at https://github.com/IntellectualSites/FastAsyncWorldEdit/blob/main/worldedit-core/src/main/java/com/sk89q/worldedit/extension/platform/AbstractPlayerActor.java#L86 creates an empty map, _not_ null.

Regarding the PermissionAttachmentManager, nullability is something that @aurorasmiles or @weaondara are best suited to comment on.

TheBusyBiscuit: I see, I just went ahead and checked WorldGuard's code again, fortunately most methods throw an `UnsupportedOperationException` anyway, as they are not meant for offline players:
https://github.com/EngineHub/WorldGuard/blob/master/worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/BukkitOfflinePlayer.java

So making the permission attachment nullable may not have any big implications? As this is a pretty edge case reminiscient of "mocking" the object. So under normal circumstances this shouldn't disturb the current mechanics.
I am open to any suggestions, I think making this work with WorldGuard's intended usage would be beneficial, but I realize that this may be a bit deeper than I first anticipated.

How should we proceed?
I thought about creating an OfflinePermissionAttachment, but the constructor takes a NotNull Permissible, even if that check isn't enforced right now.

dordsor21: Regardless of whether or not the PermissionAttachment should "work" or not when the player is null, there are still issues with NPE during the initial object instantiation, the fixing of which is presumably the main aim of the PR. I think adding a nullcheck on the player and then simply returning null should be fine given the further information you've provided.

TheBusyBiscuit: Yes, that is indeed the aim of the PR.
Would you prefer the null-check to happen in the AttachmentManager or inside BukkitPlayer?

dordsor21: nullchecks should usually happen just before a nullability issue

dordsor21: Ignore the BukkitPermissionAttachmentManager for now, as that might be changed on main branch outside of this PR

dordsor21:
I will close this for now.

TheBusyBiscuit:
@dordsor21 Yeah, I can understand that, I'm sorry.
Hacktoberfest has really been a lot more stress this year than I would have thought, I've been reviewing so many PRs lately that I didn't really have time to look at my own ones.

We just had another user report the same error again, I told them they should just raise an issue here and reference this PR.
I'll see if I can make a new pull request when I got more time on my hands, otherwise the issue will at least be properly documented/reported then - if someone else who got a bit more time available wants to tackle it.

Issue #669, opened 2020-09-28
//replace <2 causes Exeption


Crousus:
[config-legacy.txt](https://github.com/IntellectualSites/FastAsyncWorldEdit/files/5294516/config-legacy.txt)
[config.txt](https://github.com/IntellectualSites/FastAsyncWorldEdit/files/5294517/config.txt)

**[REQUIRED] FastAsyncWorldEdit Configuration Files** 

**Required Information**
- FAWE Version Number: 388
- Spigot/Paper Version Number:  Paper-174
- Minecraft Version: 1.16.2

**Describe the bug**
//replace <2 oak_leaves,0 throws an exeption
Stacktrace: https://haste.athion.net/ogijureqaw.md

**To Reproduce**
1. Select anything
2. Run //replace <2 oak_leaves,0
4. See console error

**Plugins being used on the server**

 AdvancedModreq, BKCommonLib, BlockBreaker, BungeeTabListPlus, ColoredSigns*, CraftBook, dynmap*, Essentials,  (WorldEdit)FastAsyncWorldEdit, HolographicDisplays, Images, LightCleaner, LuckPerms, MaSuiteChat, MaSuiteCore, MaSuiteHomes, MaSuitePortals, MaSuiteTeleports, MaSuiteWarps, Multiverse-Core, NametagEdit*, PlotSquared, SSX-Connector*, Vault, ViaVersion, VoteManager*, VoxelSniper, WorldBorder, WorldEditSUI, WorldGuard, WorldGuardExtraFlags

**Checklist**:

- [X] I included all information required in the sections above
- [X] I made sure there are no duplicates of this report [(Use Search)](https://github.com/IntellectualSites/FastAsyncWorldEdit/issues?q=is%3Aissue)
- [X] I made sure I am using an up-to-date version of [FastAsyncWorldEdit for 1.16.3](https://ci.athion.net/job/FastAsyncWorldEdit-1.16/)
- [X] I made sure the bug/error is not caused by any other plugin

aurorasmiles:
Fixed in #680

PR #680, opened 2020-10-04
There's no need to duplicate everything when copying the intersection masks, as the Set and array should be the same.

dordsor21:
There's no need to duplicate everything when copying the intersection masks, as the Set and array should be the same.


SirYwell (approved):

SirYwell:
Fixes #669

aurorasmiles (changes requested):
Got some concurrentModificationExceptions while testing this: https://waste.cutiepie.at/yk6bv9uaeezmqr8k

aurorasmiles (approved):

PR #687, opened 2020-10-06
Better logger support

MattBDev:
This PR removed the `debug()` methods in IFawe.java and Fawe.java and replaces their usages with a proper logger.
It also removes (almost) all usages of System.out or System.err and replaces them with a proper logger.
Some new messages were added for debugging purposes.

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/regions/EllipsoidRegion.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/regions/EllipsoidRegion.java
@@ -409,7 +409,6 @@ public class EllipsoidRegion extends AbstractRegion {
             int yTopFull = Math.min(255, cy + diffYFull);
 
             if (yBotFull == yTopFull || yBotFull > yTopFull) {
-                System.out.println("aa"); //why do we print "aa" here?
             }
             // Set those layers
             filter(chunk, filter, block, get, set, yBotFull, yTopFull, full);
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/regions/EllipsoidRegion.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/regions/EllipsoidRegion.java
@@ -409,7 +409,7 @@ public class EllipsoidRegion extends AbstractRegion {
             int yTopFull = Math.min(255, cy + diffYFull);
 
             if (yBotFull == yTopFull || yBotFull > yTopFull) {
-                System.out.println("aa");
+                System.out.println("aa"); //why do we print "aa" here?
             }
             // Set those layers
             filter(chunk, filter, block, get, set, yBotFull, yTopFull, full);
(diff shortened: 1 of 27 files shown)


N0tMyFaultOG:
If we are already using slf4j, why don't we do something like 
`private static final Logger logger = LoggerFactory.getLogger("FAWE/" + ResidenceFeature.class.getSimpleName());` for the appropriate class
and call sth like logger.info("Plugin 'Residence' found. Using it now."). Would look a bit cleaner imo.

MattBDev:
> If we are already using slf4j, why don't we do something like
> `private static final Logger logger = LoggerFactory.getLogger("FAWE/" + ResidenceFeature.class.getSimpleName());` for the appropriate class
> and call sth like logger.info("Plugin 'Residence' found. Using it now."). Would look a bit cleaner imo.

@IntellectualSites/fawe-team thoughts?

N0tMyFaultOG:
> > If we are already using slf4j, why don't we do something like
> > `private static final Logger logger = LoggerFactory.getLogger("FAWE/" + ResidenceFeature.class.getSimpleName());` for the appropriate class
> > and call sth like logger.info("Plugin 'Residence' found. Using it now."). Would look a bit cleaner imo.
> 
> @IntellectualSites/fawe-team thoughts?

Actually, that was just a quick thought I had during making dinner lol.
But checking upstream, they do it the same way I thought about it. This could potentially help preventing merge conflicts and upstream compatibility in the future.

dordsor21 (approved):

worldedit-core/src/main/java/com/sk89q/worldedit/regions/EllipsoidRegion.java
```diff
@@ -409,7 +409,7 @@ public void filter(IChunk chunk, Filter filter, ChunkFilterBlock block, IChunkGe
             int yTopFull = Math.min(255, cy + diffYFull);
 
             if (yBotFull == yTopFull || yBotFull > yTopFull) {
-                System.out.println("aa");
+                System.out.println("aa"); //why do we print "aa" here?
```
N0tMyFaultOG, line 412: Looks like some left over debug?

N0tMyFaultOG (changes requested):

worldedit-core/src/main/java/com/sk89q/worldedit/regions/EllipsoidRegion.java
MattBDev: Yeah, that's what I thought but what could it mean and is it safe to delete?

N0tMyFaultOG: I'd say so.

N0tMyFaultOG:
```suggestion
```

N0tMyFaultOG (approved):

PR #691, opened 2020-10-08
Several fixes to actual, probable and possible synchronocity issues

dordsor21:
- Ensure that all edits are queued onto the same AsyncNotifyQueue by actually delegating to the parent player in PlayerProxy
 - Ensure that the order editsessions are being remembered on the LocalSession is being respected by using a fair ReentrentLock


dordsor21:
This will also fix some interesting history behaviour from #658 as we were editing too quickly... (lol)

SirYwell:
http://haste.athion.net/obepumedaq.rb 
I got this deadlock when simply using a height brush.
send-before-history is set to true if that matters

dordsor21:
> http://haste.athion.net/obepumedaq.rb
> I got this deadlock when simply using a height brush.
> send-before-history is set to true if that matters

Interesting... I tried a bunch of stuff but didn't get a lock. The change was the "possible" one so I've just undone it.

SirYwell (approved):
Seems to work now

N0tMyFaultOG (approved):

PR #707, opened 2020-10-14
Implement biome-specific regen

IronApollo:
## Overview
This PR implements the ability to regenerate regions with biome specifications.

## Description
This allows the user to regenerate a region with whatever biome they specify. To use this, the user would execute "//regen BiomeType", where BiomeType is replaced with the name of the biome they want. For example, selecting a region of ocean and typing "//regen plains" will regenerate the region not with the original ocean biome, but with a seamless plains biome with accurate features (cave generation, tree and other populator placement, etc.).

One current downside of this implementation is that the biome-specific regeneration only works in overworld worlds, so using this in worlds like The End and Nether will simply regenerate the region with original, respective terrain instead of the target biome. This should not be used against the consideration of approving this PR.

BEFORE APPROVING:
I don't have anything locally that allows me to easily test on 1.16.1 and 1.15.x environments, thus this feature in those versions is implemented but UNTESTED! Please ensure someone has tested this in those versions prior to any approvals.

## Checklist
- [X] I tested my changes and approved their functionality (Only tested 1.16.3 at the time of PR creation)

### DIFF
--- a/worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/adapter/impl/regen/Regen_v1_15_R2.java
+++ b/worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/adapter/impl/regen/Regen_v1_15_R2.java
@@ -185,10 +185,12 @@ public class Regen_v1_15_R2 extends Regenerator<IChunkAccess, ProtoChunk, Chunk,
             public void doTick(BooleanSupplier booleansupplier) { //no ticking
             }
 
+            private final BiomeBase singleBiome = options.hasBiomeType() ? IRegistry.BIOME.get(MinecraftKey.a(options.getBiomeType().getId())) : null;
+
             @Override
             public BiomeBase a(int i, int k, int j) {
                 if (options.hasBiomeType()) {
-                    return IRegistry.BIOME.get(MinecraftKey.a(options.getBiomeType().getId()));
+                    return singleBiome;
                 }
                 return this.getChunkProvider().getChunkGenerator().getWorldChunkManager().getBiome(i, j, k);
             }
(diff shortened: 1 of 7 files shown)


worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/adapter/impl/regen/Regen_v1_15_R2.java
```diff
@@ -182,6 +184,14 @@ protected boolean initNewWorld() throws Exception {
             @Override
             public void doTick(BooleanSupplier booleansupplier) { //no ticking
             }
+
+            @Override
+            public BiomeBase a(int i, int k, int j) {
+                if (options.hasBiomeType()) {
+                    return IRegistry.BIOME.get(MinecraftKey.a(options.getBiomeType().getId()));
```
dordsor21, line 191: Is it better worth having this cached rather than obtaining every time?

IronApollo: For sure. I'll try to work that out later today if I find the time.

IronApollo: Resolved in https://github.com/IntellectualSites/FastAsyncWorldEdit/pull/707/commits/27f16d97b31af01773deecb5760538cd71cefe54

dordsor21 (approved):

N0tMyFaultOG (approved):

N0tMyFaultOG:
Tested on 1.15.2, Paper build 389, works as intended.

PR #714, opened 2020-10-20
Full support for 3D biomes

dordsor21:
Since we're only supporting 1.15+ there's no need to try anything other than compatibility

This will currently break any existing biome history (and biome history only) files.

### DIFF
--- a/worldedit-core/src/main/java/com/boydti/fawe/object/clipboard/DiskOptimizedClipboard.java
+++ b/worldedit-core/src/main/java/com/boydti/fawe/object/clipboard/DiskOptimizedClipboard.java
@@ -185,7 +185,7 @@ public class DiskOptimizedClipboard extends LinearClipboard implements Closeable
 
     @Override
     public boolean setBiome(int x, int y, int z, BiomeType biome) {
-        setBiome(getIndex(x, y, z), biome);
+        setBiome(getIndex(x, 0, z), biome);
         return true;
     }
 
--- a/worldedit-core/src/main/java/com/boydti/fawe/object/extent/OffsetExtent.java
+++ b/worldedit-core/src/main/java/com/boydti/fawe/object/extent/OffsetExtent.java
@@ -2,9 +2,7 @@ package com.boydti.fawe.object.extent;
 
 import com.sk89q.worldedit.WorldEditException;
 import com.sk89q.worldedit.extent.Extent;
-import com.sk89q.worldedit.math.BlockVector2;
 import com.sk89q.worldedit.math.BlockVector3;
-import com.sk89q.worldedit.math.MutableBlockVector2;
 import com.sk89q.worldedit.world.biome.BiomeType;
 import com.sk89q.worldedit.world.block.BlockStateHolder;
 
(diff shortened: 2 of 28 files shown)


SpeedRider:
oooo can't wait for such a cool thing.

N0tMyFaultOG (approved):

worldedit-core/src/main/java/com/boydti/fawe/object/extent/OffsetExtent.java
```diff
@@ -46,10 +43,4 @@ public boolean setBiome(int x, int y, int z, BiomeType biome) {
         throws WorldEditException {
         return getExtent().setBlock(x + dx, y + dy, z + dz, block);
     }
-
-    @Override
-    public ResettableExtent setExtent(Extent extent) {
-        mutable = new MutableBlockVector2();
-        return super.setExtent(extent);
-    }
```
MattBDev, line 54: What was the reason for removing this?

worldedit-core/src/main/java/com/boydti/fawe/object/clipboard/DiskOptimizedClipboard.java
```diff
@@ -185,7 +185,7 @@ public boolean setBiome(BlockVector3 position, BiomeType biome) {
 
     @Override
     public boolean setBiome(int x, int y, int z, BiomeType biome) {
-        setBiome(getIndex(x, y, z), biome);
+        setBiome(getIndex(x, 0, z), biome);
```
MattBDev, line 188: Why was this changed from `y` to `0`?

worldedit-core/src/main/java/com/sk89q/worldedit/extent/OutputExtent.java
```diff
@@ -94,7 +94,11 @@ default boolean fullySupports3DBiomes() {
      */
     @Deprecated
     default boolean setBiome(BlockVector2 position, BiomeType biome) {
-        return setBiome(position.toBlockVector3(), biome);
+        boolean result = false;
+        for (int y = 0; y < 256; y ++) {
+            result |= setBiome(position.toBlockVector3().mutY(y), biome);
```
MattBDev, line 99: Not ideal but I can't think of a better solution right now and it works.

MattBDev:
Looks like I was super lazy when I tried doing 3D Biomes the first time. Oops.

worldedit-core/src/main/java/com/boydti/fawe/object/extent/OffsetExtent.java
dordsor21: 3D biomes means no need for the mutable vector so it doesn't need to be reset. Technically the extent doesn't even need to be resettable anymore so I'll change that as well.

worldedit-core/src/main/java/com/boydti/fawe/object/clipboard/DiskOptimizedClipboard.java
dordsor21: because it's broken otherwise. Clipboards do not have 3D biome support yet, and it was a mistake for the y to be there in the first place (as stated in the commit for the change).

worldedit-core/src/main/java/com/sk89q/worldedit/extent/OutputExtent.java
dordsor21: This is really the only solution; biomes are only stored in 4x4x4 sections, however, doing every 4 blocks could and would cause issues with offsets/transforms, etc

MattBDev (approved):

PR #717, opened 2020-10-21
Upstream changes

SirYwell:
## Overview
I mainly pulled these changes to get `//fast` working again to disable history.

## Description
`//fast` and `//perf` are different commands now. At the moment, the user is warned when using `//fast` as upstream targets to remove this command in future. In FAWE, `//fast` toggles history while side effects (`//perf`) aren't implemented. Therefore, `//fast` will remain in FAWE too, but messages needs to be changed.

As pulling from upstream caused a lot of issues in past, this should be tested to avoid breaking functionality. From my personal testings, I didn't find any issues.

## Checklist


worldedit-core/src/main/java/com/sk89q/worldedit/command/GeneralCommands.java
```diff
@@ -76,6 +85,62 @@
...
+        );
+
+
+        Set<org.enginehub.piston.Command> commands = collect.getAllCommands()
+            .collect(Collectors.toSet());
+        for (org.enginehub.piston.Command command : commands) {
+            if (command.getName().equals("/fast")) {
+                // deprecate to `//perf`
+                commandManager.register(CommandUtil.deprecate(
+                    command, "//fast duplicates //perf " +
```
N0tMyFaultOG, line 108: //fast serves a different purpose for FAWE. The code should be adjusted to line that out.

worldedit-core/src/main/java/com/sk89q/worldedit/command/GeneralCommands.java
```diff
@@ -76,6 +85,62 @@
...
+    }
+
+    private static Component replaceFastForPerf(org.enginehub.piston.Command oldCmd,
+                                                CommandParameters oldParams) {
+        if (oldParams.getMetadata() == null) {
+            return CommandUtil.createNewCommandReplacementText("//perf");
+        }
+        ImmutableList<String> args = oldParams.getMetadata().getArguments();
+        if (args.isEmpty()) {
+            return TextComponent.of("There is not yet a replacement for //fast" +
```
N0tMyFaultOG, line 126: See my previous comment.

N0tMyFaultOG (changes requested):

worldedit-core/src/main/java/com/sk89q/worldedit/command/GeneralCommands.java
SirYwell: I commented this part out, therefore the method below isn't called. I guess we can keep this code around until it's removed from WE anyways.

worldedit-core/src/main/java/com/sk89q/worldedit/command/GeneralCommands.java
```diff
@@ -145,22 +213,48 @@ public void timeout(Actor actor, LocalSession session,
...
+        if (hasFastMode) {
+            session.setFastMode(false);
+            actor.printInfo(TranslatableComponent.of("worldedit.fast.disabled"));
+        } else {
+            session.setFastMode(true);
+            actor.printInfo(TranslatableComponent.of("worldedit.fast.enabled"));
+        }
+    }
+
+    @Command(
```
N0tMyFaultOG, line 238: This needs to be implemented properly.

You currently cannot click the status "On" or "Off". It will will up the incorrect help page for fast.

N0tMyFaultOG (changes requested):

worldedit-core/src/main/java/com/sk89q/worldedit/command/GeneralCommands.java
SirYwell: Resolved, but we might need to think about `//perf` in general at some point I guess

N0tMyFaultOG (approved):

aurorasmiles (approved):

PR #734, opened 2020-11-03
Update FAWE to 1.16.4

aurorasmiles:
Updates FAWE to 1.16.4


worldedit-bukkit/src/main/java/com/boydti/fawe/bukkit/wrapper/AsyncChunk.java
```diff
@@ -183,4 +189,9 @@ public boolean contains(@NotNull BlockData block) {
         //todo
         return false;
     }
+
+    @Override
+    public @NotNull PersistentDataContainer getPersistentDataContainer() {
+        return null;
```
aurorasmiles, line 195: I'm not entirely sure what to do here, so I'll just return null for now

dordsor21: return the value of the super Method. Any other return nulls should do the same.

SirYwell (commented):
I'm not sure about our naming scheme here, as the package is 1_16_4, most classes also reflect that but some have 1_16_R3 instead

aurorasmiles:
That is because the version is 1.16.4, but 1.16 R3

dordsor21:
> I'm not sure about our naming scheme here, as the package is 1_16_4, most classes also reflect that but some have 1_16_R3 instead

Feel free to try to change Minecraft lol

N0tMyFaultOG (approved):

SirYwell:
> Feel free to try to change Minecraft lol
That wans't my intent. I wanted to know why we're using *both*. We have a `mc1_16_1` package and a `mc1_16_2` one but not a `mc1_16_3` and FAWE still works on 1.16.3 so something is wrong about our naming scheme here. That has nothing to do with minecraft itself at all.

aurorasmiles:
Because both Minecraft 1.16.2 and 1.16.3 are R2. We did not need seperate stuff for 1.16.2 and .3 because nothing server side changed

SirYwell:
Right, so does our naming scheme make sense? Like, if there are changes in nms code we rely on but there wasn't a revision bump, we need different implementations for the same revision (so the 1_x_Ry naming scheme doesn't make sense), otherwise we might have the same code for multiple versions in a package `mc1_x_y` which doesn't make sense either.
Is there are real reason we're mixing the naming schemes?

aurorasmiles:
No, just always been that way. I actually quite like it because it reminds you that R3 isn't .3, because otherwise that might get mixed up a fair bit.

SirYwell:
Yeah but if we don't actually need to care about the minecraft patch numbers, we could use the revision naming scheme everywhere instead. Doesn't really matter for this PR but rather in general tho

MattBDev:
I have never been a fan of the naming scheme of these classes. I think using the Minecraft data version in the class name would be better and we could add a javadoc in each class to specify which version of the game the code works with. https://minecraft.gamepedia.com/Data_version

aurorasmiles:
Thats a bad idea. The data version would confuse even more people.

aurorasmiles:
Also, could this be discussed somewhere else? I think this is something for another PR and has nothing to do with this one. I'd like to have 1.16.4 support soon

IronApollo (approved):

PR #807, opened 2020-12-21
Reorder BuildInClipboardFormat and document changed JNBT classes

Citymonstret:
This commit includes changes from upstream to the schematic classes
(`com.sk89q.worldedit.extent.clipboard.io`). It also documents the JNBT
classes, specifying what has been changed in FAWE. This was done in preparation
for the upcoming move to adventure-nbt.


worldedit-core/src/main/java/com/sk89q/jnbt/ByteArrayTag.java
```diff
@@ -61,4 +56,11 @@ public String toString() {
         return "TAG_Byte_Array(" + hex + ")";
     }
 
+    // FAWE Start
+    @Override
```
me4502, line 60: Why create a diff here just to move where in the file it is? Like I agree the place it's at in the WE-bundled JNBT is bad, but seems like a pretty pointless diff to create here, especially when JNBT is likely being dropped in WE8

N0tMyFaultOG:
This breaks `//copy -e` http://haste.athion.net/docepocoqo.sql

N0tMyFaultOG (changes requested):
This is kinda out of the scope of the pr, but `getTileEntities` would need to be overridden in AsyncChunk to make it compile again. Preferably doing this here so we can give the PR a proper test.

Sauilitired:
- To read schematics including entities with old broken FAWE offsets: `//schematic read <name> fawe`
- To read schematics including entities with WE offset: `//schematic read <name>`

N0tMyFaultOG (approved):

aurorasmiles (changes requested):
We need to make sure to handle entities properly as discussed

MattBDev:
I'm about half way done my review. So far my opinions are mixed on this PR

worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/io/BuiltInClipboardFormat.java
```diff
@@ -56,34 +60,84 @@ public String getPrimaryFileExtension() {
...
-            if (inputStream instanceof FileInputStream) {
-                inputStream = new ResettableFileInputStream((FileInputStream) inputStream);
-            }
-            BufferedInputStream buffered = new BufferedInputStream(inputStream);
-            NBTInputStream nbtStream = new NBTInputStream(new BufferedInputStream(new GZIPInputStream(buffered)));
-            SchematicReader input = new SchematicReader(nbtStream);
-            input.setUnderlyingStream(inputStream);
-            return input;
+            NBTInputStream nbtStream = new NBTInputStream(new GZIPInputStream(inputStream));
+            return new MCEditSchematicReader(nbtStream);
```
dordsor21, line 64: why?

worldedit-core/src/main/java/com/boydti/fawe/util/MainUtil.java
```diff
@@ -34,6 +34,9 @@
 import net.jpountz.lz4.LZ4Utils;
 import org.jetbrains.annotations.NotNull;
 
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
```
MattBDev, line 38: Replace all instances of `javax.annotation.Nonnull` and `javax.annotation.Nullable` to JetBrains annotations.

worldedit-core/src/main/java/com/boydti/fawe/util/MainUtil.java
```diff
@@ -448,24 +450,44 @@ public void close() {
...
+    /**
+     * Create a copy of the tag and modify the (x, y, z) coordinates
+     *
+     * @param tag Tag to copy
+     * @param x   New X coordinate
+     * @param y   New Y coordinate
+     * @param z   New Z coordinate
+     * @return New tag
+     */
+    public static @NotNull CompoundTag setPosition(@Nonnull CompoundTag tag, int x, int y, int z) {
```
MattBDev, line 462: Mixed use of `@NotNull` and `@Nonnull`. Using `@NotNull` is preferred.

worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/io/BuiltInClipboardFormat.java
```diff
@@ -86,56 +86,65 @@ public boolean isFormat(File file) {
...
+                Map<String, Tag> schematic = schematicTag.getValue();
+                if (!schematic.containsKey("Version")) {
+                    return false;
+                }
+            } catch (Exception e) {
+                return false;
+            }
+
+            return true;
         }
```
MattBDev, line 126: Jesse once yelled at me when I tried to change this because he claimed that checking the file extension was much faster than checking for a schematic tag and version key in a file. How long does this operation usually take before we make this change?

MattBDev (changes requested):
In addition to the specific comments I have left, can you please explain why you are creating more objects and more immutables?

worldedit-core/src/main/java/com/sk89q/jnbt/NBTConstants.java
```diff
@@ -20,14 +20,13 @@
...
 import java.nio.charset.Charset;
-import java.nio.charset.StandardCharsets;
 
 /**
  * A class which holds constant values.
  */
 public final class NBTConstants {
 
-    public static final Charset CHARSET = StandardCharsets.UTF_8;
+    public static final Charset CHARSET = Charset.forName("UTF-8");
```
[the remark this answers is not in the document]
Citymonstret: File copied from upstream.

worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/io/BuiltInClipboardFormat.java
Citymonstret: By re-arranging to the MCEDIT format last in the enum this won't matter. Trying to minimise the diff as FAWE doesn't change behaviour. Altering things for the sake of altering them is dumb.

Citymonstret:
> In addition to the specific comments I have left, can you please explain why you are creating more objects and more immutables?

JNBT compounds are immutable, FAWE shouldn't be altering random NBT classes (especially not without documenting the changes...). WE is moving to adventure-nbt and assuming mutability is going to make the transition a lot harder than it has to be.

worldedit-core/src/main/java/com/sk89q/jnbt/NBTConstants.java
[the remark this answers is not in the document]
MattBDev: I actually changed this on purpose. I know it deviates from upstream.

MattBDev:
> > In addition to the specific comments I have left, can you please explain why you are creating more objects and more immutables?
> 
> JNBT compounds are immutable, FAWE shouldn't be altering random NBT classes (especially not without documenting the changes...). WE is moving to adventure-nbt and assuming mutability is going to make the transition a lot harder than it has to be.

Unfortunately altering random classes and not documenting the changes was the majority of legacy FAWE (much of which is still present in the new codebase). It's getting better though. Thanks for clearing the mutability stuff up.

aurorasmiles (approved):

dordsor21 (approved):

PR #838, opened 2021-01-09
Looks like automagical relighting

### LINKED ISSUE #686: Implement lighting into FAWE queue.
**[REQUIRED] FastAsyncWorldEdit Configuration Files** 

 https://athion.net/ISPaster/paste/view/1fea5cd152b3424a916dda2c294a72d2

**Required Information**
- FAWE Version Number: 1.16-403;f71ca32
- Spigot/Paper Version Number: Tuinity 88a2c88 (MC: 1.16.3) AND Paper #216
- Minecraft Version: 1.16.3

**Describe the bug**
Lightning is not implemented into the queue. This results in areas where lightning is obviously incorrect after being edited by FAWE.

**To Reproduce**
1. Select an area and use //set air
2. Note the incorrect lightning.

**Plugins being used on the server**

Refer to debug paste. I am using Tokens by GC which utilizes FAWE for new pickaxe enchants that destroy many blocks.

Screenshot of what this issue looks like: https://cdn.discordapp.com/attachments/756175891456852019/763063163482013706/unknown.png

dordsor21:
Fixes #686

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/regions/Region.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/regions/Region.java
@@ -25,6 +25,7 @@ import com.boydti.fawe.beta.IChunk;
 import com.boydti.fawe.beta.IChunkGet;
 import com.boydti.fawe.beta.IChunkSet;
 import com.boydti.fawe.beta.implementation.filter.block.ChunkFilterBlock;
+import com.boydti.fawe.beta.implementation.processors.ProcessorScope;
 import com.boydti.fawe.object.FaweLimit;
 import com.boydti.fawe.object.extent.SingleRegionExtent;
 import com.sk89q.worldedit.extent.Extent;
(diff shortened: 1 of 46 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/regions/Region.java
```diff
@@ -365,4 +366,9 @@ default Extent construct(Extent child) {
         }
         return new SingleRegionExtent(child, FaweLimit.MAX, this);
     }
+
+    @Override
+    default ProcessorScope getScope() {
+        return ProcessorScope.REMOVING_BLOCKS;
```
SirYwell, line 372: Why is `REMOVING_BLOCKS` considered the default? 
I'm not sure if I got the concept behind the scope, but in general, I'd prefer to return a (Enum)Set, as processors might work in multiple scopes (does that make sense?). Then, we wouldn't need a `CUSTOM` constant and could use noneOf or allOf (whatever makes sense) instead.

SirYwell (changes requested):

worldedit-core/src/main/java/com/sk89q/worldedit/regions/Region.java
dordsor21: the scope is basically what determines the order in which processors are run. Region processors remove blocks, as they limit blocks to a region. The actual default is CUSTOM, as it's safest to run any processors people add themselves after anything FAWE does, but before FAWE writes history

N0tMyFaultOG (approved):

aurorasmiles (approved):
Tested, works great

PR #918, opened 2021-02-16
Move to adventure-nbt

NotMyFault:
WIP - I doubt I have the time to finish that anytime soon, but the initials are set.

TODO:
- [ ] - Migrate Fawe-specific NBT stuff
- [ ] - Remove middleman-NBT, where it applies, we added back some time ago as compatibility layer between we <-> fawe
- [ ] - Merge https://github.com/IntellectualSites/worldedit-adapters/pull/1 right before this pr


MattBDev (approved):

SirYwell:
Are we pulling from master? If so, that means we're introducing changes that break with WE 7.2.x (although internally only, no official API). Not sure if we really want that.

MattBDev:
> Are we pulling from master? If so, that means we're introducing changes that break with WE 7.2.x (although internally only, no official API). Not sure if we really want that.

Not sure what you mean? Everything here is as close to upstream as I could get without breaking compatibility. In the future we will need to implement NBT data better but I feel like for now this is a good step to get us closer to being up to date with upstream and help us update to 1.17

PR #941, opened 2021-02-27
Apply a lot of synchronization to ChunkHolder

dordsor21:
This is basically the main "chunk" class for internal FAWE. Chunk operations should (and are) almost always single-threaded operations, however, under certain circumstances, it is possible for the chunk to be "called" (flushed: written to the world and sent to the player) from a separate thread. This would specifically occur from SingleThreadQueueExtent when there are a lot of chunks being loaded in memory by FAWE (where the chunk would then be submitted to a multi-threaded queue). It would therefore be possible for a thread accessing the chunk to attempt to access it in the middle of the call, which can lead to a number of issues, and it is my opinion that the most frequent of these is the NPE seen during lighting operations, where new chunks can be accessed/loaded very quickly, increasing the likelihood for the aforementioned synchronisation issue to occur.

This PR will decrease the overall performance slightly, however, will hopefully fix several current issues with FAWE and I'm making it a PR to get other opinions on whether they agree that this could be a cause for them or not.

### DIFF
--- a/worldedit-core/src/main/java/com/boydti/fawe/beta/implementation/chunk/ChunkHolder.java
+++ b/worldedit-core/src/main/java/com/boydti/fawe/beta/implementation/chunk/ChunkHolder.java
@@ -59,11 +59,11 @@ public class ChunkHolder<T extends Future<T>> implements IQueueChunk<T> {
     }
 
     @Override
-    public void recycle() {
+    public synchronized void recycle() {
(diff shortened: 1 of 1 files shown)


NotMyFault (approved):
From my little testing I didn't run into oddities with this pr.

MattBDev (approved):
Stability is more important than performance and I approve of the changes.

Issue #396, opened 2020-04-05
Addressing reflections & deprecations for modern Java


Xemorr:
# Bug report for FastAsyncWorldEdit 1.15.2

**[REQUIRED] FastAsyncWorldEdit Configuration Files:** 

Config.yml = https://hasteb.in/muqahavu.yaml
Latest log = 
```
[15:33:02] [Server thread/INFO]: Starting minecraft server version 1.15.2
[15:33:02] [Server thread/INFO]: Loading properties
[15:33:02] [Server thread/INFO]: This server is running Paper version git-Paper-162 (MC: 1.15.2) (Implementing API version 1.15.2-R0.1-SNAPSHOT)
[15:33:03] [Server thread/INFO]: Debug logging is disabled
[15:33:03] [Server thread/INFO]: Server Ping Player Sample Count: 12
[15:33:03] [Server thread/INFO]: Using 4 threads for Netty based IO
[15:33:03] [Server thread/INFO]: Default game type: SURVIVAL
[15:33:03] [Server thread/INFO]: Generating keypair
[15:33:03] [Server thread/INFO]: Starting Minecraft server on *:25565
[15:33:03] [Server thread/INFO]: Using default channel type
[15:33:11] [Server thread/INFO]: Got request to register class com.sk89q.worldedit.bukkit.BukkitServerInterface with WorldEdit [com.sk89q.worldedit.extension.platform.PlatformManager@364beb72]
[15:33:11] [Server thread/INFO]: WEPIF: Using the Bukkit Permissions API.
[15:33:11] [Server thread/INFO]: Using com.sk89q.worldedit.bukkit.adapter.impl.FAWE_Spigot_v1_15_R2 as the Bukkit adapter
[15:33:12] [Server thread/INFO]: Ignoring invalid block minecraft:sign
[15:33:12] [Server thread/INFO]: Ignoring invalid block minecraft:wall_sign
[15:33:15] [Server thread/INFO]: [WorldEdit] Loading WorldEdit v1.15-164;5cc7a09
[15:33:15] [Server thread/INFO]: [FastAsyncWorldEdit] Loading FastAsyncWorldEdit v1.15.2
[15:33:15] [Server thread/INFO]: [ViaVersion] Loading ViaVersion v3.0.0-SNAPSHOT
[15:33:15] [Server thread/INFO]: [ViaVersion] ViaVersion 3.0.0-SNAPSHOT is now loaded, injecting!
[...]

PR #971, opened 2021-03-13
Switch from PropertyKey enum to class

SirYwell:
## Overview
This targets #396 (but doesn't resolve, there are more places where illegal reflective operations occur).

## Description
I replaced the PropertyKey enum with a class. It required some workarounds, as e.g. switch statements don't work with normal classes. The performance impact is neglectable and FAWE doesn't need to hackily insert enum constants on runtime.

## Checklist


NotMyFault (approved):

MattBDev:
I don't like that performance is decreasing in these latest changes we've been having lately but I approve just to get rid of the hacky stuff we do at runtime. I hate all that stuff. So as long as this works, approve.

MattBDev (approved):

SirYwell:
To clarify, I wasn't able to extract any useful data from my profiling as the involved methods don't matter in the overall time consumption. The `BlockTransformExtent` also has some other places that could be improved if it's too slow.

PR #991, opened 2021-03-24
Remove freebuild regions

NotMyFault:
Removing the implementation in its current state makes it easier for us and does not require overriding a plenty of methods in Chunk, World, whatnot from Spigot/Paper just for sake of third party access.


MattBDev (approved):
100% Approved.

SirYwell (approved):

Issue #1028, opened 2021-04-12
Excessive log files / errors crash the server


KateHutch:
### Server Implementation

Paper

### Server Version

1.16.5

### Describe the bug

Over the past couple of months, we have had an intermittent issue where an occasional error in FAWE causes the hard drive to fill up completely (in excess of 300GB). I'm not aware of what specifically triggers this issue.

This error has been happening on the following versions:

- FastAsyncWorldEdit-1.16-616.jar
- FastAsyncWorldEdit-1.16-626.jar
- FastAsyncWorldEdit-1.16-648.jar

Due to the size of the log file I have not been able to access it, however I here is a link to a snippet of the error log: https://haste.athion.net/uhamaxotod.sql

### To Reproduce

I'm not actually sure what triggers this issue, nor the steps to reproduce it but I believe it's something like this:

1. Install FAWE on your server
2. Give 30 players access to FAWE on PlotSquared plots
3. Walk away for 24 hours
4. Return to a crashed server, a full HD, and corrupt files

### Expected behaviour

If FAWE encounters a "fatal error" it would be great if it could either forcefully stop whatever task it's doing.

### Error log (if applicable)

https://haste.athion.net/uhamaxotod.sql

### Fawe Debugpaste

N/A

### Fawe Version

FastAsyncWorldEdit-Bukkit-1.16-648.jar

### Checklist

I have included a Fawe debugpaste.
I am using the newest build from https://ci.athion.net/job/FastAsyncWorldEdit-1.16/ and the issue still persists.

NotMyFault:
Fixed in https://github.com/IntellectualSites/FastAsyncWorldEdit/commit/57773765d779efba76764188ec65ef3f45ca1977

Issue #1025, opened 2021-04-12
Deadlock in chunk holder


RainbowDashLabs:
### Server Implementation

Paper

### Server Version

1.16.5

### Describe the bug

I experienced Deadlocks in fawe for some time now.
The thread dump shows three deadlocks inside the chunk holder:
https://haste.athion.net/adilifurel.bash

FAWE died at 12:50 in our latest log.
https://haste.athion.net/gicotadabu.md

At this point nothing fawe related works anymore. 
No FAWE command responds in any way. I am not able to create a debugpaste nor seeing the basic fawe information with the /fawe command.

However I added a debugpaste after a server restart to provide the required meta information.

inb4 Youre not using the latest fawe version:
I know. Its hard to wait for a bug occurring every 24 hours or so. I try to update every few days. However the bug is common on our server for month now so I dont think a miracle will happen and it magically disappears. So I guess it will happen with all future builds currently as well.

### To Reproduce

The bug is kinda weird.
It is not reproducible in any way I can grantee that it works.
When using fawe at some point a deadlock occurs.
Our observation was that it is kinda reproducible when using different stencil setups and undo a few time or so. However It doesnt work anytime. Sometime it takes hours to get stuck and sometime only a few minutes.

### Expected behaviour

Not running into a deadlock I guess ¯\\\_(ツ)_/¯

### Fawe Debugpaste

 https://athion.net/ISPaster/paste/view/c752108ed6a449efa7ccbfcbddd5bc57

### Fawe Version

1.16-655;88764d8

### Checklist

I have included a Fawe debugpaste.
I am using the newest build from https://ci.athion.net/job/FastAsyncWorldEdit-1.16/ and the issue still persists.

Aurelien30000:
Confirmed on my side, nothing coherent in this issue :/ I'll try to investigate.

jewome62:
Duplicate with #749 
Yes, this is a major problem currently on my server

RainbowDashLabs:
I have another thread dump of this behavior.
First time:
https://chojos.lewds.de/wsZcOWLTWc.txt
Second time:
https://chojos.lewds.de/s9aVesbdvM.txt

This time we managed to get 5 and 4 deadlocks.
We also managed to find a way to reproduce this behavior on the latest fawe build.

```
/br sphere grass_block 20
/mask /[0d][45d]<7
//1
Play around with the brush for a bit.
```

[...]

PR #1023, opened 2021-04-10
Relight using starlight engine on Tuinity

SirYwell:
## Overview

This PR makes use of the relight-methode in Tuinity. Thanks to @Aurelien30000 for digging into it before.

Not sure which issues are all fixed with that, feel free to link them :p

Fixes https://github.com/IntellectualSites/FastAsyncWorldEdit/issues/979
Fixes https://github.com/IntellectualSites/FastAsyncWorldEdit/issues/770
Fixes https://github.com/IntellectualSites/FastAsyncWorldEdit/issues/857
Fixes https://github.com/IntellectualSites/FastAsyncWorldEdit/issues/1021
Fixes https://github.com/IntellectualSites/FastAsyncWorldEdit/issues/977
Fixes https://github.com/IntellectualSites/FastAsyncWorldEdit/issues/972
Fixes https://github.com/IntellectualSites/FastAsyncWorldEdit/issues/956
Fixes https://github.com/IntellectualSites/FastAsyncWorldEdit/issues/950
Fixes https://github.com/IntellectualSites/FastAsyncWorldEdit/issues/914
Fixes https://github.com/IntellectualSites/FastAsyncWorldEdit/issues/897
Fixes https://github.com/IntellectualSites/FastAsyncWorldEdit/issues/891
Fixes https://github.com/IntellectualSites/FastAsyncWorldEdit/issues/881
Fixes https://github.com/IntellectualSites/FastAsyncWorldEdit/issues/870
Fixes https://github.com/IntellectualSites/FastAsyncWorldEdit/issues/700
Fixes https://github.com/IntellectualSites/FastAsyncWorldEdit/issues/1043

## Description

The creation of a Relighter is now abstracted into a RelighterFactory. That allows to create different relighters on different platforms while keeping the code at those places clear.

Tuinity got its own Relighter for 1.16.5, making use of the rewritten light engine. To avoid dependency issues, we don't depend on Tuinity but rather call the specific method using a MethodHandle.

I'm not sure if sendChunks should happen on a different thread or if it's safe to be called before (or why changes aren't sent before/otherwise) so if anyone has more insights, please let me know.

TODO:

- [x] Heightmaps should get their own batch processor so they work with Tuinity too.

## Checklist

- [x] I included all information required in the sections above
- [x] I tested my changes and approved their functionality
- [x] I ensured my changes do not break other parts of the code
[...]

PR #1030, opened 2021-04-15
Char block null check

dordsor21:
**Add a null-check to CharBlocks FULL section layer-retrieval.**
 - It is possible to trim CharBlocks whilst it is attempting to read data due to the batching of chunks to help reduce memory
 - This is done when the number of chunks sitting loaded in memory with having been "submitted" to the queue for writing to disk becomes high
 - Seconday operations such as heightmap processing and lighting will quickly load chunks, meaning many chunks are submitted early
 - This leads to much higher chances of the chunk being submitted and subsequently trimmed given heightmap and light processing is done layer-by-layer over many chunks, rather than chunk-by-chunk - thus leading to NPEs.
 - By adding synchronisation to and around only the specific sections when loading/updating, and not blocking the whole chunk, many access can still be thread-safe without causing deadlocks
 - This allows removal of lots of the needless and very-slowing synchronisation on get**Block** methods

**Remove much of the synchronisation from ChunkHolder**
 - We shouldn't be synchronising with call() and safety should be added elsewhere. (plus it's making edits very very slow when queue target size is hit)
 - Also remove much of synchronisation because we've added the null-check and section-specific synchronisation to CharBlocks
 
**Some QOL/thread-safe data access changes**
- Replaces the Array#clone seen in the get blocks classes with System#arraycopy as deep cloning is not required, and is also slower than arraycopy
- Add System#arraycopy when accessing chunk section data via history to ensure it is not altered whilst being written
- Renaming EMPTY to empty means it is not implied to be a static variable

Fixes https://github.com/IntellectualSites/FastAsyncWorldEdit/issues/1028
Fixes https://github.com/IntellectualSites/FastAsyncWorldEdit/issues/1025
Fixes https://github.com/IntellectualSites/FastAsyncWorldEdit/issues/1089
Fixes https://github.com/IntellectualSites/FastAsyncWorldEdit/issues/1091
Fixes https://github.com/IntellectualSites/FastAsyncWorldEdit/issues/1097


dordsor21:
Should address #1028 and #1025

NotMyFault:
When merging this pr, https://github.com/IntellectualSites/FastAsyncWorldEdit/pull/1023/commits/737606a791f436b4d893efaddfafdbf5b2008e57#diff-ea7f891c2aa11eb18c3c9a4ac6c613c3a6b4bfcf76972d97375c08a95cd42801R59 may need revisiting before merging too.

MattBDev:
So due to the heightmap and lightning process, if a chunk is submitted but gets trimmed after it was already submitted an NPE would occur? I just want to make I understand the bug being fixed.

dordsor21:
> So due to the heightmap and lightning process, if a chunk is submitted but gets trimmed after it was already submitted an NPE would occur? I just want to make I understand the bug being fixed.

It's the only place where the array can realistically be nullified. It's not directly caused by these, but it's because (basically onlu lighting) needs to load many chunks fast since it can't do it per-chunk, it needs to do large areas one y-level at a time. Therefore a lot of chunks end up getting submitted before lighting is finished with them, thus nullifying the arrays when they get trimmed.

dordsor21:
(I tested this a bunch and couldn't reproduce the errors the synchronisation created. The NPE error I'm attempting to fix here was also not reproduced).

worldedit-core/src/main/java/com/boydti/fawe/beta/implementation/blocks/CharBlocks.java
```diff
@@ -16,6 +16,13 @@
     protected static final Section FULL = new Section() {
         @Override
         public final char[] get(CharBlocks blocks, int layer) {
+            if (blocks.blocks[layer] == null) {
```
SirYwell, line 19: Does that still return the expected values? Can a non-empty section be trimmed? Because if that's the case, processing chunks will end up in wrong calculations.

worldedit-core/src/main/java/com/boydti/fawe/beta/implementation/blocks/CharBlocks.java
```diff
@@ -16,6 +16,13 @@
...
+                if (blocks.sections[layer].isFull()) {
+                    throw new IllegalStateException("Array cannot be null: " + blocks.getClass());
```
SirYwell, line 21: I cherry-picked your commits on the feature/tuinity-starlight branch and I throws this exception now when doing `//fillr green_wool 200`.
Stacktrace: 
https://paste.gg/p/anonymous/d32d21c5521b4545becbf0dd91ab3f4b

NotMyFault (approved):

SirYwell (changes requested):
Tried `//fillr` with the latest build from the CI (695) and a build from this branch. On the build from the CI, the NPEs are thrown and the result is somewhat broken (as mentioned in the bug reports). Using the build from this branch, the result is even more broken, not sent to the player correctly and changes aren't written to history properly (`//undo` fails). No exceptions are written to console.

dordsor21:
I would like to have more than the standard required two reviews on this, so I've disabled auto-merge

SirYwell (approved):

PR #1057, opened 2021-05-09
Reduced code duplication

terrarier2111:
## Overview
When looking at the code base I noticed some duplicated code in a class and i removed it.

## Description
I removed the duplicated code.

## Result
More readable code.


SirYwell:
I'd prefer to keep it similar to upstream.

NotMyFault:
Not speaking of that this wouldn't compile at all, I see no reason to diverge from upstream if not really necessary.
This will just make pulling upstream more difficult over time.

PR #1100, opened 2021-05-30
Ability to add multiple regions on a single FaweMask

DolphinGamerYT:
## Overview

No issue has been opened, since it's an addition.

## Description

I'm making a plugin for a client that masks the regions of GriefDefender and perms, etc. The point is, that a user can have multiple regions, and FaweMask just gives me 1 region to mask. I've added an extended class of FaweMask that supports multiple regions, and added a comparation on WEManager.java to check if the mask is an instance of that class, and add the multiple regions. It is currectly working perfectly with my plugin.

## Checklist


DolphinGamerYT:
@SirYwell told me about RegionIntersection.java, and it works the same way, but I think this method would have more sense using it the way I wanted, since the regions that are going to be masked, is not a single region, but multiple regions (Different GriefDefender claims). It's totally up to you.

NotMyFault:
@SirYwell Would we (still) want to add such a thing? Because this is infact covered by other methods 🤔

SirYwell:
I think that's covered and can be closed.

PR #1213, opened 2021-08-04
Fix major security bugs (3 brushes + superpickaxe)!

Aurelien30000:
## Overview
Fix major security bugs (3 brushes _(stacker, repl & cycler)_ + superpickaxe)!
Issues can be easily reproduced using one of the brushes above or ``//superpickaxe``, and PS road.

## Description
- Due to some recent changes, FAWE could edit everything in the world using a few tools, no matter other plugin protections such as PS or WG.
- Fix superpickaxe allow to bypass protections => Fix SurvivalModeExtent not taking into account protections plugins due to breaking blocks naturally to get drops.
---
- ``EditSession#disableBuffering()`` method shouldn't be used at all since it breaks _all other extents_, some were re-added by @NotMyFault in https://github.com/IntellectualSites/FastAsyncWorldEdit/commit/50ab8ad5c706630930d5e83712c5d566aa7a2dde
- Previously, **superpickaxe** dropped items on the ground by breaking naturally a block and without checking anything. I added a method ``World#getBlockDrops`` to return default block drops instead. **Superpickaxe** now drop them only if actor is able to break the targeted block.
~~_Note: WorldEdit upstream is likely affected by the same issue with **superpickaxe**._~~

## Checklist

- [x] I read and followed the [contribution guidelines](https://github.com/IntellectualSites/FastAsyncWorldEdit/blob/main/CONTRIBUTING.md)

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/command/tool/BlockDataCyler.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/command/tool/BlockDataCyler.java
@@ -90,6 +90,10 @@ public class BlockDataCyler implements DoubleActionBlockTool {
                 BlockState newBlock = block.with(objProp, currentProperty.getValues().get(index));
 
                 try (EditSession editSession = session.createEditSession(player)) {
+                    // FAWE start
+                    // editSession.disableBuffering();
+                    // FAWE end
+
                     try {
                         editSession.setBlock(blockPoint, newBlock);
                         player.print(Caption.of(
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/command/tool/BlockDataCyler.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/command/tool/BlockDataCyler.java
@@ -90,8 +90,6 @@ public class BlockDataCyler implements DoubleActionBlockTool {
                 BlockState newBlock = block.with(objProp, currentProperty.getValues().get(index));
 
                 try (EditSession editSession = session.createEditSession(player)) {
-                    editSession.disableBuffering();
-
                     try {
                         editSession.setBlock(blockPoint, newBlock);
                         player.print(Caption.of(
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/command/tool/BlockDataCyler.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/command/tool/BlockDataCyler.java
@@ -90,10 +90,6 @@ public class BlockDataCyler implements DoubleActionBlockTool {
                 BlockState newBlock = block.with(objProp, currentProperty.getValues().get(index));
 
                 try (EditSession editSession = session.createEditSession(player)) {
-                    // FAWE start
-                    // editSession.disableBuffering();
-                    // FAWE end
-
                     try {
                         editSession.setBlock(blockPoint, newBlock);
                         player.print(Caption.of(
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/extent/world/SurvivalModeExtent.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/extent/world/SurvivalModeExtent.java
@@ -19,13 +19,18 @@
 
 package com.sk89q.worldedit.extent.world;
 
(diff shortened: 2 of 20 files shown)


Aurelien30000:
_Since I changed a method behavior, should I keep it, add diff annotations or consider another way to fix it?_

worldedit-core/src/main/java/com/sk89q/worldedit/world/World.java
```diff
@@ -215,8 +216,9 @@ default int getBlockLightLevel(BlockVector3 position) {
      * Simulate a block being mined at the given position.
      *
      * @param position the position
+     * @return Block default drops
      */
-    void simulateBlockMine(BlockVector3 position);
+    Collection<BaseItemStack> simulateBlockMine(BlockVector3 position);
```
MattBDev, line 221: Missing FAWE Start/end comments. It needs to specify how it differs from upstream.

SirYwell: This would be an API breaking change, causing incompatibility with WE. You need to keep the old method and introduce a new one. It also needs to be a `default` method then to be fully compatible with any classes implementing `World`. (e.g. just call the void method and return an empty list/set)

worldedit-core/src/main/java/com/sk89q/worldedit/extent/world/SurvivalModeExtent.java
```diff
@@ -91,8 +96,21 @@ public void setStripNbt(boolean stripNbt) {
     @Override
     public <B extends BlockStateHolder<B>> boolean setBlock(BlockVector3 location, B block) throws WorldEditException {
         if (toolUse && block.getBlockType().getMaterial().isAir()) {
-            world.simulateBlockMine(location);
-            return true;
+            Collection<BaseItemStack> drops = world.simulateBlockMine(location);
+            boolean canSet = super.setBlock(location, block);
+            if (canSet) {
+                for (BaseItemStack stack : drops) {
```
SirYwell, line 102: Put the for loop in the `run` method instead.

SirYwell (changes requested):

Aurelien30000:
Ty for review, I'll address requests. 

**Important note: WorldEdit is exempt from all those issues, tried earlier, ``#disableBuffering`` doesn't cause protections to be _ignored_ and superpickaxe works perfectly!**

worldedit-core/src/main/java/com/sk89q/worldedit/extent/world/SurvivalModeExtent.java
Aurelien30000: @SirYwell Would it be appropriate to return a boolean with ``#simulateBlockMine`` method, without changing its behavior?

worldedit-core/src/main/java/com/sk89q/worldedit/world/World.java
```diff
@@ -218,6 +220,19 @@ default int getBlockLightLevel(BlockVector3 position) {
      */
     void simulateBlockMine(BlockVector3 position);
 
+    //FAWE start
+
+    /**
+     * Return block drops, as if it were mined.
```
NotMyFault, line 226: Better spelling? Using the superpickaxe on grass e.g. drops grass which isn't a block. Item > block imo.

NotMyFault (changes requested):

worldedit-core/src/main/java/com/sk89q/worldedit/command/tool/BlockDataCyler.java
```diff
@@ -90,7 +90,9 @@ private boolean handleCycle(
                 BlockState newBlock = block.with(objProp, currentProperty.getValues().get(index));
 
                 try (EditSession editSession = session.createEditSession(player)) {
-                    editSession.disableBuffering();
+                    // FAWE start
+                    // editSession.disableBuffering();
+                    // FAWE end
```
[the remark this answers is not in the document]
Aurelien30000: Oh ok!

NotMyFault (approved):

SirYwell (approved):

PR #1220, opened 2021-08-09
Fix reading of tile entities with "id" instead of "Id" tag

Gerrygames:
## Overview
If a schematic contains a tile entity with an "id" instead of an "Id" the previous code would delete the id. The changed code checks which one of the tags is set.

This schematic for example caused this exception:
https://i.imgur.com/qjrT6ut.png
http://paste.cytooxien.de/jowobiyile.bash

## Checklist

- [x] I read and followed the [contribution guidelines](https://github.com/IntellectualSites/FastAsyncWorldEdit/blob/main/CONTRIBUTING.md)

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/io/SpongeSchematicReader.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/io/SpongeSchematicReader.java
@@ -253,7 +253,13 @@ public class SpongeSchematicReader extends NBTSchematicReader {
                 values.put("x", new IntTag(pt.getBlockX()));
                 values.put("y", new IntTag(pt.getBlockY()));
                 values.put("z", new IntTag(pt.getBlockZ()));
-                values.put("id", values.get("Id"));
+                //FAWE start
+                if (!values.containsKey("id")) {
+                    values.put("id", requireTag(values, "Id", StringTag.class));
+                } else {
+                    requireTag(values, "id", StringTag.class);
+                }
+                //FAWE end
                 values.remove("Id");
                 values.remove("Pos");
                 if (fixer != null) {


dordsor21:
"Id" is the correct format based on the schematic format. https://github.com/SpongePowered/Schematic-Specification/blob/master/versions/schematic-2.md

If "id" is present, then the schematic is highly likely to have been created by someone that doesn't follow the format, and is therefore highly likely to be "corrupt".

Gerrygames:
But we/fawe uses id internally. So why should it not accept files with id? Also it did work fine in previous versions. And an obscure npe should not be expected behavior even with corrupted files. That's why I added the  requireTag.

dordsor21:
> But we/fawe uses id internally. So why should it not accept files with id? Also it did work fine in previous versions. And an obscure npe should not be expected behavior even with corrupted files. That's why I added the requireTag.

Minecraft uses "id" internally, so so does FAWE/WE. The Sponge schematic format details using "Id", so that is used. If "id" is saved to a schematic by either WE or FAWE then something has gone wrong, and it is likely to be corrupt.

Gerrygames:
The file is not corrupt. It works perfectly fine with an old version of FAWE and with this fix.

dordsor21:
> The file is not corrupt. It works perfectly fine with an old version of FAWE and with this fix.

FAWE used to write corrupt schematics by writing "id" rather than "Id" or omitting it altogether sometimes. This was corrected at the end of 2020

Gerrygames:
And those schematics are no longer supporter now? Nice solution.

NotMyFault:
> > The file is not corrupt. It works perfectly fine with an old version of FAWE and with this fix.
> 
> FAWE used to write corrupt schematics by writing "id" rather than "Id" or omitting it altogether sometimes. This was corrected at the end of 2020

Do you have an appx. duration in mind for how long that used to be a thing? Seeing that this is the first report after 8 months, I assume not for toolong or people simply redoing their schematics after half a year. Else we could consider some kind of legacy handing how we do it with TEs.

MattBDev:
@Gerrygames Was this tested loading the "broken" schematic? If so, did the schematic load successfully?


> > The file is not corrupt. It works perfectly fine with an old version of FAWE and with this fix.
> 
> FAWE used to write corrupt schematics by writing "id" rather than "Id" or omitting it altogether sometimes. This was corrected at the end of 2020

If there are still schematics out there that were written with the incorrect tag name then we should support loading them. However, we should notify people that these schematics are corrupted and need to be resaved (unless we do this automatically).

dordsor21 (changes requested):
I've made a change that "supports" tile entities in "corrupt" positions, thus I suppose it's logical this PR should also be accepted. It would, however, also need to make the same change in the FastSchematicReader

dordsor21:
Implemented in 6b5f21d67bfe6badd4d893fb58e2ece22e7bdcd8

PR #1256, opened 2021-08-26
Improve exceptions

dordsor21:
- Kick more exceptions further up the pipeline to be more likely to be shown to player
- Try to avoid lots of console spamming when it's the same error multiple times
- Allow parsing of FaweExceptions during commands to better give information to players


worldedit-core/src/main/java/com/fastasyncworldedit/core/extent/processor/MultiBatchProcessor.java
```diff
@@ -26,9 +31,14 @@
 
 public class MultiBatchProcessor implements IBatchProcessor {
 
-    private IBatchProcessor[] processors;
+    private static final Logger LOGGER = LogManagerCompat.getLogger();
+
     private final LoadingCache<Class<?>, Map<Long, Filter>> classToThreadIdToFilter =
             FaweCache.IMP.createCache((Supplier<Map<Long, Filter>>) ConcurrentHashMap::new);
+    private boolean[] faweExceptionReasonsUsed = new boolean[FaweException.Type.values().length];
```
SirYwell, line 38: This should be an (Enum)Set

worldedit-core/src/main/java/com/fastasyncworldedit/core/queue/implementation/ParallelQueueExtent.java
```diff
@@ -40,19 +45,30 @@
...
+    private int lastException = Integer.MIN_VALUE;
+    private int exceptionCount = 0;
 
     public ParallelQueueExtent(QueueHandler handler, World world, boolean fastmode) {
         super(handler.getQueue(world, new BatchProcessorHolder(), new BatchProcessorHolder()));
         this.world = world;
         this.handler = handler;
         this.processor = (BatchProcessorHolder) getExtent().getProcessor();
+        if (this.processor.getProcessor() instanceof MultiBatchProcessor) {
+            ((MultiBatchProcessor) this.processor.getProcessor()).setFaweExceptionArray(faweExceptionReasonsUsed);
```
SirYwell, line 66: Is it thread safe to share the array here?

worldedit-core/src/main/java/com/fastasyncworldedit/core/Fawe.java
```diff
@@ -346,4 +347,46 @@ public Thread setMainThread() {
...
+     *
+     * @param faweExceptionReasonsUsed boolean array that should be cached where this method is called from of length {@code
+     *                                 FaweException.Type.values().length}
+     * @param e                        {@link FaweException} to handle
+     */
+    public static void handleFaweException(boolean[] faweExceptionReasonsUsed, FaweException e) {
+        FaweException.Type type = e.getType();
+        switch (type) {
+            case OTHER:
+                LOGGER.warn("FaweException: " + e.getMessage());
```
SirYwell, line 363: Why do we only log the message instead of the actual exception?

SirYwell (changes requested):
The overall logging looks suboptimal to me, all the Throwable#printStackTrace() calls should be replaced with proper logging

worldedit-core/src/main/java/com/fastasyncworldedit/core/extent/processor/MultiBatchProcessor.java
dordsor21: I used an array because it's not going to throw errors with concurrent read/write, and it's not an important enough "feature" to need to worry about perfect synchronisation - the only thing that would happen is an error being displayed twice, but as long as it's not being spammed, it's fine.

worldedit-core/src/main/java/com/fastasyncworldedit/core/queue/implementation/ParallelQueueExtent.java
dordsor21: See above

worldedit-core/src/main/java/com/fastasyncworldedit/core/Fawe.java
dordsor21: This probably ought to be a stacktrace yeah

dordsor21:
> The overall logging looks suboptimal to me, all the Throwable#printStackTrace() calls should be replaced with proper logging

I suppose it could be useful to know where a stacktrace is actually being printed sometimes... But with exception handling now, it should be relatively trivial to work it out anyway? It would make stacktraces an even bigger block of text

SirYwell:
The logger prints stacktraces too, and it would be consistent with every other logging that way

dordsor21:
Are we (or upstream) actually doing it anywhere? Or would you want all stacktrace prints to be loggered. Would probably be best to give the logger instance from the class calling the new Fawe exception handler method

SirYwell:
There are a few places where it is done, but upstream does not have many places where (unexpected) Exceptions are caught anyways.

SirYwell (approved):

PR #1287, opened 2021-09-10
Deprecate sponge schematic reader/writer and tell users why

dordsor21:
Push people towards actually using FAWE and not OOMing


NotMyFault (approved):

MattBDev (approved):

IronApollo:
Are there any Issues opened that outline the dangers of using FastSchematicReader? I would be interested in knowing what the dangers are since this is being set to the default schematic reader.

dordsor21:
> Are there any Issues opened that outline the dangers of using FastSchematicReader? I would be interested in knowing what the dangers are since this is being set to the default schematic reader.

They're all closed now, but mostly it was an ooold issue from legacy FAWE that had incorrect tile entity positions saved, and issues with entities. Nothing new's appeared for a while now.

NotMyFault (approved):

PR #1293, opened 2021-09-12
Use an object array for synchronising on internal chunk sections rather than the sections array itself.

dordsor21:
- Synchronising on full sections synchronises on the global FULL instance (bad)
 - Synchronising on empty sections synchronises on the local empty instance (bad)
 - Leads to needless thread locking, and raises the possibility of cyclic locks considerably


MattBDev:
It seems like I still have more to learn on synchronization. Elements of an array can be locked instead of the entire array itself? I hope I asked that question right.

dordsor21:
> It seems like I still have more to learn on synchronization. Elements of an array can be locked instead of the entire array itself? I hope I asked that question right.

It's ultimately not much different to having a method that simply returns the object at an index in the array, and then synchronising onto that object (except technically not synchronising on a local variable)

NotMyFault (approved):

MattBDev (approved):

Issue #39, opened 2019-02-04
Different sets of disallowed blocks with permissions.


RealmportalMC:
for example: 

Default group would have access to basic blocks, but not things like cactus or tnt or lava  or command blocks / structure blocks etc.

But allow a 2nd group (trusted players) to have access to tnt lava and cactus but still not command blocks or structure blocks.

much the same way different block limits can be set.

(also a side note. is there a permission for allowing NBT copying?)

AeSix:
I feel that since this is part of WorldGuard, that there's no need for this to be part of FAWE.  The use of WorldGuard is quite usual for servers, and adds a great amount of benefit.  It should be used in lieu of adding this feature. 

That's my $0.02.

RealmportalMC:
Edit, im especially referring to which blocks can be set / replaced etc

boy0001:
There's a permission to bypass disallowedBlocks already.

N0tMyFaultOG:
No.

AeSix:
I see the benefit of having block group permissions for FAWE in regard to editing, as it would make it a cleaner experience for those users who don't have permissions for some block they're trying to set or replace.  With that said, I still believe this is a super low priority goal, if something to achieve at all.  

side note:  Perhaps stalebot can be set to ignore a specific label, such as "pinned" or "future" for which tickets tagged with won't be set to stale?

PR #1312, opened 2021-09-20
Add configurable disallowed states for limits, global disallowed blocks

dordsor21:


worldedit-core/src/main/java/com/fastasyncworldedit/core/configuration/Settings.java
```diff
@@ -192,7 +192,17 @@
...
         })
         public List<String> STRIP_NBT = new ArrayList<>();
-
+        @Comment({
+                "If the disallowed blocks listed in config-legacy.yml should be disallowed in all edits,",
+                "not just where blocks patterns are used.",
+                " - Can prevent blocks being pasted from clipboards, etc.",
+                " - If fast-placement is disabled, this may cause edits to be slower."
+        })
+        public boolean UNIVERSAL_DISALLOWED_BLOCKS = true;
```
NotMyFault, line 201: Instead/on top of of making WorldEdit's list global, I'd (rather) vouch for #39 too

dordsor21: Yeah just noticed that issue and implementing atm

worldedit-core/src/main/java/com/fastasyncworldedit/core/configuration/Settings.java
```diff
@@ -192,7 +192,21 @@
...
+        @Comment({
+                "List of blocks to deny use of",
+        })
+        public List<String> DISALLOWED_BLOCKS = new ArrayList<>();
```
NotMyFault, line 205: To show the format used, we could copy the 3 examples from `config-legacy.yml` here. I think, that would be easier than explaining over and over how to put sth into `[]`.

worldedit-core/src/main/java/com/fastasyncworldedit/core/configuration/Settings.java
```diff
@@ -192,7 +192,21 @@
...
+        @Comment({
+                "List of blocks to strip nbt from",
+        })
+        public List<String> DISALLOWED_STATES = new ArrayList<>();
```
NotMyFault, line 209: Looking of the description of the option, wouldn't that be the same like `STRIP_NBT`?

worldedit-core/src/main/java/com/sk89q/worldedit/extension/factory/parser/DefaultBlockParser.java
```diff
@@ -178,6 +182,20 @@ private String woolMapper(String string) {
...
+            disallowed = context
+                    .getActor()
+                    .getLimit().DISALLOWED_STATES
+                    .stream()
+                    .map(String::toLowerCase)
+                    .collect(Collectors.toSet());
+            checkState = !disallowed.isEmpty();
+        } else {
+            disallowed = Collections.emptySet();
+        }
```
NotMyFault, line 197: Fawe annotations

worldedit-core/src/main/resources/lang/strings.json
```diff
@@ -144,6 +144,7 @@
     "fawe.error.parse.unknown-transform": "Unknown transform: {0}, See: {1}",
     "fawe.error.parse.no-clipboard": "To use {0}, please first copy something to your clipboard",
     "fawe.error.parse.no-clipboard-source": "No clipboards found at given source: {0}",
+    "fawe.error.limit.disallowed-block": "Your limit disallows use of nlock '{0}'",
```
NotMyFault, line 147: Yes.

NotMyFault (changes requested):

worldedit-core/src/main/java/com/sk89q/worldedit/world/block/BlockTypesCache.java
```diff
@@ -296,4 +288,19 @@ private static BlockType register(final String id, int internalId, List<BlockSta
         return existing;
     }
 
+    public static Map<String, Property<?>> getAllProperties() {
```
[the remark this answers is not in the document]
dordsor21: The class isn't in upstream. Should probably be refactors outside this PR

dordsor21:
> 
> Why does it trigger the usage message if you have no permission for sth? Only the bottom part should be send to the player

because the pattern therefore doesn't apply and throws an exception so that gets printed alongside the pattern not parsing. Nothing new.

NotMyFault:
> > 
> > Why does it trigger the usage message if you have no permission for sth? Only the bottom part should be send to the player
> 
> because the pattern therefore doesn't apply and throws an exception so that gets printed alongside the pattern not parsing. Nothing new.

Oic, might be a coincidence of these messages being out of place if you have no active selection:

NotMyFault (changes requested):
- Adding `waterlogged` to `disallowed-states` and running `//set oak_slab[waterlogged=true]` after that replies with "Operation completed." and strips the water but doesn't drop a note about it, we may should add one.
- Very opinionated, but we either should make `minecraft:` optional or add it if it's missing, otherwise entries are ignored at all.

dordsor21:
> * Very opinionated, but we either should make `minecraft:` optional or add it if it's missing, otherwise entries are ignored at all.

https://github.com/IntellectualSites/FastAsyncWorldEdit/pull/1312/files#diff-bc0e072b834198de95f949925a122fd8384b86a12812be9f995737b9c38f417dR48-R50

NotMyFault:
> > * Very opinionated, but we either should make `minecraft:` optional or add it if it's missing, otherwise entries are ignored at all.
> 
> https://github.com/IntellectualSites/FastAsyncWorldEdit/pull/1312/files#diff-bc0e072b834198de95f949925a122fd8384b86a12812be9f995737b9c38f417dR48-R50
> 
> >  GitHub**[Add configurable disallowed states for limits, global disallowed blocks by dordsor21 · Pull Request #1312 · IntellectualSites/FastAsyncWorldEdit](https://github.com/IntellectualSites/FastAsyncWorldEdit/pull/1312)**Blazingly fast world manipulation for artists, builders and everyone else: https://www.spigotmc.org/resources/13932/ - Add configurable disallowed states for limits, global disallowed blocks by dordsor21 · Pull Request #1312 · IntellectualSites/FastAsyncWorldEdit

Oic, yet that works for the disallowed-blocks in the config-legacy.yml only.

dordsor21:
> Oic, yet that works for the disallowed-blocks in the config-legacy.yml only.

Fixed, along with better usage exception handling as just committed to main

NotMyFault (changes requested):
is only fired if you add a block in the format of `minecraft:bedrock`, using `bedrock` fails silently with "Operation completed." yet it cancels the operation.

dordsor21:
> 
> is only fired if you add a block in the format of `minecraft:bedrock`, using `bedrock` fails silently with "Operation completed." yet it cancels the operation.

Technically doesn't "fail silently" or cancel the operation. The operation completes, but without setting the disallowed blocks.

NotMyFault (approved):

worldedit-core/src/main/java/com/fastasyncworldedit/core/extent/DisallowedBlocksExtent.java
```diff
@@ -0,0 +1,163 @@
...
+        }
+        // ignore blocks in default state.
+        if (block.getOrdinalChar() == block.getBlockType().getDefaultState().getOrdinalChar()) {
+            return block;
+        }
+        BlockState def = block.getBlockType().getDefaultState();
+        for (Property<?> property : blockedStates) {
+            Object value = def.getState(property);
+            if (block.getState(property) != value) {
+                block = block.with(property.getKey(), (Object) def.getState(property));
```
SirYwell, line 100: the value is here the same as 2 lines above

SirYwell (changes requested):
I don't think the blocked states work the way they should.

For example, forbidding `waterlogged` results in forbidding `minecraft:conduit[waterlogged=true]` and `minecraft:conduit[waterlogged=false]` but using `minecraft:conduit` (which has `waterlogged=true` as default) is still fine.

In my opinion, it should be all block *state* based, so you can forbid exactly `minecraft:conduit[waterlogged=true]` (though writing `minecraft:conduit` should still forbid the block type alltogether by collecting all states).

dordsor21:
Perhaps using a "force-blockstates" list like
```
    forced-states:
        key: value
        key2: value
```
default being
`example: value` and "example" is ignored in code?

SirYwell:
Okay, I'm wondering if this actually solves the problem we want to solve.

For me, there are two ways one wants to deal with disallowing specific blocks:

#### 1. 
Don't place forbidden block types/states. This would be comprehensive to the user, the setup could be a simplified version of masks (block types and block states only, but the same syntax and the same behavior). 
This would only require one setting, as it works for both types and states. It would also be possible to allow block tags this way.

#### 2. 
Don't place forbidden block types in general, but modify specific block states to a different value. This might be useful for copy/pasting in particular. It requires a more complex setup, the mask (or general block state) syntax would be insufficient. To make it actually useful, there should be ways to transform a property `A` of a specific block type (!) from value `x` to value `y`. 

Default block states are likely not good enough (when considering the default for conduits is `waterlogged=true`), but if more than two values are possible, you probably only want to map *one* of them to another - and keep everything else the same (e.g. map `cobblestone_wall[north=low]` to `cobblestone_wall[north=none]` but don't change `cobblestone_wall[north=tall]`).

I doubt a combination of multiple block states would be relevant (e.g, disallow exactly `minecraft:note_block[instrument=guitar,note=5]` and map it to `minecraft:note_block[instrument=flute,note=8]`), but that would make it even more complex.

----

The second approach allows more granular settings, but is way more complex, especially for the user to set up properly. The first is less powerful but would work in a way people are used to already.

What of that do we want to achieve?

dordsor21:
If merging the states into the blocks config list, then every single instance of a particular property (e.g. waterlogged) would need to be added to the list/blocked which isn't very viable. Given the use-case of this is really only for blocking waterlogged, or other states that can cause lag there shouldn't be a need to add finer configurability anyway.

I see no reason for anyone to want to block a specific stair direction (for example) at all really.

dordsor21:
> For this, the mask syntax could be reused too, e.g. `//paste -m ^=[waterlogged=false]` only places blocks with that state defined (the meaning of what a mask does would be inverted to make sense when speaking about disallowed blocks).

The point would be to still allow blocks to be placed, but just not be waterlogged. Any more complex computation using masks is unneeded for the ultimate use case of the feature imo.

SirYwell:
> Any more complex computation using masks is unneeded for the ultimate use case of the feature imo.

Masks are way more simple here (for both the user, as it's comprehensive and convient, and the computation, as no new block state needs to be calculated), but just aren't powerful enough to cover what you want to achieve.

That said, I think there is a difference between *disallowing* block types/states and *remapping* specific properties, therefore it should also be addressed separately.

dordsor21:
> Masks are way more simple here (for both the user, as it's comprehensive and convient, and the computation, as no new block state needs to be calculated), but just aren't powerful enough to cover what you want to achieve.
> 
> That said, I think there is a difference between _disallowing_ block types/states and _remapping_ specific properties, therefore it should also be addressed separately.

Hmm okay, fair. If states are merged into types as a "blocked-blocks" that works, assuming a fuzzy "mask" for the blocked states. For remapping, configuration will need some thought. Something like
```yaml
remap-properties:
    - exampleBoolProp:true:false
    - exampleIntProp:7:1,8:2
```

SirYwell:
Yes, that would probably work. Another design would be
```yml
remap-properties:
  exampleBoolProp:
    true: 'false'
  exampleIntProp:
    7: '1'
    8: '2'
```
not sure if this is better or not though...

dordsor21:
My concern with something like that is actually being able to support it with the current config implementation

SirYwell:
I'm fine with either here, so just do whatever is easier to integrate I guess

PR #1364, opened 2021-10-17
Fix paperweight changes

PierreSchwang:
## Overview
Newest paperweight changes break FAWE logic. This PR fixes those issues.

## Description
- Moved v1_17_R1_2 adapter to correct package structure to be resolved by BukkitImplLoader
- Fixed obfuscated mappings / field names

## Checklist


PierreSchwang:
I've had the problem as well. This took me a few hours, but the cause of that is in the gradle.properties:
`org.gradle.caching=true`
You have to build the project using `./gradlew clean build --no-build-cache` as this skips the cache. I also wrote about this yesterday with SirYwell, apparently Gradle then throws some classes around and cloned them 3x all over the place for me.

NotMyFault:
> I've had the problem as well. This took me a few hours, but the cause of that is in the gradle.properties: `org.gradle.caching=true` You have to build the project using `./gradlew clean build --no-build-cache` as this skips the cache. I also wrote about this yesterday with SirYwell, apparently Gradle then throws some classes around and cloned them 3x all over the place for me.

That does the trick. Addressing the imports and we're good to go, I guess.

PR #1399, opened 2021-11-04
Fix incompatibility between WorldEdit and FAWE for LocalSession#getTool

### LINKED ISSUE #1398: FAWE breaks compatibility with WorldEdit on LocalSession#getTool

WorldEdit has only LocalSession#getTool(ItemType) and does not have any overload which matches any of the FAWE method signatures.

The original WorldEdit method signature should be readded again to avoid hacky workarounds when supporting FAWE and WorldEdit

Using reflections, which is not really nice and makes it annoying to support FAWE and WorldEdit

### Anything else?

_No response_

RainbowDashLabs:
## Overview
WorldEdit has only LocalSession#getTool(ItemType) and does not have any overload which matches any of the FAWE method signatures.

Fixes #1398

## Description
The original WorldEdit method signature is readded again to avoid hacky workarounds when supporting FAWE and WorldEdit

## Checklist

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/LocalSession.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/LocalSession.java
@@ -1120,6 +1120,24 @@ public class LocalSession implements TextureHolder {
         //FAWE end
     }
 
+    /**
+     * Get the tool assigned to the item.
+     *
+     * @param item the item type
+     * @return the tool, which may be {@code null}
+     */
+    @Nullable
+    @Deprecated
+    //FAWE start
+    //This method is here for byte code compatibility with WorldEdit.
+    //WorldEdit does not have any other getTool method with matching overloads.
+    public Tool getTool(ItemType item) {
+        synchronized (this.tools) {
+            return tools.get(item.getInternalId());
+        }
+        //FAWE end
+    }
+
     //FAWE start
     @Nullable
     public Tool getTool(Player player) {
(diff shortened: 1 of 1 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/LocalSession.java
```diff
@@ -1120,6 +1120,24 @@ public Tool getTool(ItemType item, Player player) {
...
+    /**
+     * Get the tool assigned to the item.
+     *
+     * @param item the item type
+     * @return the tool, which may be {@code null}
+     */
+    @Nullable
+    @Deprecated
+    //FAWE start
+    //This method is here for byte code compatibility with WorldEdit.
```
NotMyFault, line 1132: If that is an upstream method, the FAWE annotations are not needed. Yet, the deprecation note can point to the recommended Fawe method to use.

RainbowDashLabs: I thought they are needed since the method body differs from WorldEdit. World Edit uses a simple HashMap to store the tools.

I will add a deprecation comment to the method.

NotMyFault: > I thought they are needed since the method body differs from WorldEdit. World Edit uses a simple HashMap to store the tools.

That would then need to be disclosed within annotations as a one liner

dordsor21:
The annotations also ought to be within the method body, else it implies the method is added by FAWE

worldedit-core/src/main/java/com/sk89q/worldedit/LocalSession.java
RainbowDashLabs: I adjusted it and moved the annotations to the body.

It is not a one line since WorldEdit does not use a synchronized block. Should be fine now.

NotMyFault (approved):

dordsor21 (approved):

Issue #1372, opened 2021-10-20
Invalid config field with PS v4


FZDNSZYT:
### Server Implementation

Paper

### Server Version

1.17.1

### Describe the bug

[com.fastasyncworldedit.core.configuration.Config] Invalid config field: enabled-components.plotsquared-v4-hook for enabled-components
[07:51:37 ERROR]: [com.fastasyncworldedit.core.configuration.Config] Failed to set config option: enabled-components.plotsquared-v4-hook: true | com.fastasyncworldedit.core.configuration.Settings$ENABLED_COMPONENTS@3feeb857 | Settings.yml

### To Reproduce

This error will appear every time the server is restarted

### Expected behaviour

Update the ".jar" file can not be repaired

### Error log (if applicable)

[com.fastasyncworldedit.core.configuration.Config] Invalid config field: enabled-components.plotsquared-v4-hook for enabled-components [07:51:37 ERROR]: [com.fastasyncworldedit.core.configuration.Config] Failed to set config option: enabled-components.plotsquared-v4-hook: true | com.fastasyncworldedit.core.configuration.Settings$ENABLED_COMPONENTS@3feeb857 | Settings.yml

### Fawe Debugpaste

https://athion.net/ISPaster/paste/view/08a679de0a184fd8a65e8944031dabee

### Fawe Version

FastAsyncWorldEdit version 1.17-365;e7a6c3e

### Checklist

- [X] I have included a Fawe debugpaste.
- [X] I am using the newest build from https://ci.athion.net/job/FastAsyncWorldEdit-1.17/ and the issue still persists.

srnyx:
What is even "invalid" about it?

PR #1427, opened 2021-11-16
Fix PSv4 config error

NotMyFault:
## Description
Fixes https://github.com/IntellectualSites/FastAsyncWorldEdit/issues/1372

## Checklist

- [X] I tested my changes and approved their functionality.
- [X] I read and followed the [contribution guidelines](https://github.com/IntellectualSites/.github/blob/main/CONTRIBUTING.md)

### DIFF
--- a/worldedit-bukkit/src/main/java/com/fastasyncworldedit/bukkit/regions/plotsquaredv4/PlotSquaredFeature.java
+++ b/worldedit-bukkit/src/main/java/com/fastasyncworldedit/bukkit/regions/plotsquaredv4/PlotSquaredFeature.java
@@ -40,7 +40,7 @@ public class PlotSquaredFeature extends FaweMaskManager {
     public PlotSquaredFeature() {
         super("PlotSquared");
         LOGGER.info("Optimizing PlotSquared");
-        if (com.fastasyncworldedit.core.configuration.Settings.IMP.ENABLED_COMPONENTS.PLOTSQUARED_v4_HOOK) {
+        if (com.fastasyncworldedit.core.configuration.Settings.IMP.ENABLED_COMPONENTS.PLOTSQUARED_V4_HOOK) {
             Settings.Enabled_Components.WORLDEDIT_RESTRICTIONS = false;
             try {
                 setupBlockQueue();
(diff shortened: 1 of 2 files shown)


dordsor21:
Any particular reason for not using debug? I feel it would fit better here

dordsor21 (approved):

SirYwell:
Is there any specific reason why we don't just make the `v` uppercase here https://github.com/IntellectualSites/FastAsyncWorldEdit/blob/2c56e480c3a4171d71c71103e965ae4f376fb368/worldedit-core/src/main/java/com/fastasyncworldedit/core/configuration/Settings.java#L91?

NotMyFault:
> Is there any specific reason why we don't just make the `v` uppercase here
> 
> https://github.com/IntellectualSites/FastAsyncWorldEdit/blob/2c56e480c3a4171d71c71103e965ae4f376fb368/worldedit-core/src/main/java/com/fastasyncworldedit/core/configuration/Settings.java#L91
> 
> ?

Considering it's already marked for removal, I spend no further time with its implementation. Although, _my_ preferred solution would be dropping v4, since we dropped v5 when v6 arrived. Furthermore, the versions v4 supports no longer match with the versions Fawe supports.

SirYwell:
> Considering it's already marked for removal, I spend no further time with its implementation.

I guess it would make more sense removing it althogether rather than spending time on it with workarounds.

NotMyFault:
> > Considering it's already marked for removal, I spend no further time with its implementation.
> 
> I guess it would make more sense removing it althogether rather than spending time on it with workarounds.

If I remember correctly, @dordsor21 had some concerns about removing it, the last time we were debating about it.

dordsor21:
If changing v4 -> V4 actually fixes the issue, this is trivial. V4 support currently requires not extra effort as it's working as-is, and _can_ still be used with FAWE. No need to remove v4 support until there's no way to run the two together.

NotMyFault:
v4 used tp support 1.13 and 1.14, Fawe does no longer support these versions, so even if you are using v4, you can't use Fawe with it in the first place.

dordsor21:
Does v4 not work at all with 1.15/16? I see no reason why it shouldn't work?

NotMyFault:
I never gave it a try, as that used to be the time we dropped v5.

dordsor21:
Looking at v4 reviews from entitlee people on spigot, it doesn't work on 1.16 but does on 1.15

NotMyFault:
Oic, we have no overlapping statistic for that specific use case, but people using 1.15.2 are a small minority.

dordsor21:
Title no longer fits

NotMyFault:
> Title no longer fits

Smh, that only applies to repos using release drafter.

NotMyFault:
Rebasing on Jenkins was successful.

NotMyFault:
@SirYwell Named recommendation has been applied.

SirYwell (approved):
Looks good

PR #1483, opened 2021-12-21
Remove deprecations marked for removal that are not present in upstream

dordsor21:
2.0.0 allows us to remove all of this. If there's any particular reason anything shouldn't be removed, please say


MattBDev (approved):

NotMyFault (approved):
Looks good and doing that right now is a good opportunity.

PR #1502, opened 2021-12-28
Add back statically-set reserved IDs for air blocks only, make it clear they're "reserved"

dordsor21:
Also:
 - Ensure that reserved is never returned in GET block operations
 - "empty" thus doesn't exist in the GET update methods; remove the needless checks
 - Allow GET/SET chunks to determine their own default values for non-present blocks/sections


dordsor21:
Adapters for older versions can and should also be updated similarly when this is merged

NotMyFault (approved):

SirYwell (approved):

PR #1551, opened 2022-01-17
Only call `org.bukkit.World` height methods if they're present (1.17+)

dordsor21:


worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitWorld.java
```diff
@@ -82,10 +82,10 @@
 
 public class BukkitWorld extends AbstractWorld {
 
-    private static final Logger LOGGER = LogManagerCompat.getLogger();
+    public static final boolean HAS_3D_BIOMES;
+    public static final boolean HAS_MIN_Y;
```
NotMyFault, line 86: Missing Fawe annotations, but `HAS_3D_BIOMES` can stay private, no?

NotMyFault (approved):
Untested, due me not having such an old test setup atm, but looks good.

SirYwell (changes requested):
Wouldn't it make more sense to use the `getMinY()`|`getMaxY()` methods from the WE world? (or `getMinimumPoint().getY()`|`getMaximumPoint().getY()` from Extent if no world is present)

dordsor21:
> Wouldn't it make more sense to use the `getMinY()`|`getMaxY()` methods from the WE world? (or `getMinimumPoint().getY()`|`getMaximumPoint().getY()` from Extent if no world is present)

Changing to a BukkitWorld for regen stuff would require wrapping or the method signatures to all be changes to accept BukkitWorlds. For that case it's far much efficient to just access the static field

SirYwell (approved):

NotMyFault:
This PR is ready to be merged, we will merge it in ~24 hours if there is no negative feedback, as documented by the [merge process](https://github.com/IntellectualSites/IntellectualSites/blob/main/docs/MAINTAINERS_CONTRIBUTORS.adoc#merge-process) documentation.

PR #1554, opened 2022-01-19
Add fluid ticking and refactor post-processing a bit

dordsor21:


dordsor21:
I feel personally attacked by sonatype

NotMyFault:
> I feel personally attacked by sonatype

</3

Should be better by now, may need to add more rules that don't necessarily apply to us in the future.

worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/adapter/BukkitImplAdapter.java
```diff
@@ -350,5 +351,9 @@ default RelighterFactory getRelighterFactory() {
     default Map<String, List<Property<?>>> getAllProperties() {
         return Collections.emptyMap();
     }
+
+    default IBatchProcessor getTickingPostProcessor() {
```
NotMyFault, line 355: Can we get javadocs and a `@since TODO` here?

worldedit-core/src/main/java/com/fastasyncworldedit/core/queue/IBatchProcessor.java
```diff
@@ -23,7 +23,11 @@
      */
     IChunkSet processSet(IChunk chunk, IChunkGet get, IChunkSet set);
 
-    Future<IChunkSet> postProcessSet(IChunk chunk, IChunkGet get, IChunkSet set);
+    /**
+     * Post-process a chunk that has been edited. Set should NOT be modified here, changes will NOT be flushed to the world,
+     * but MAY be flushed to history. Defaults to nothing as most Processors will not use it.
+     */
+    default void postProcessSet(IChunk chunk, IChunkGet get, IChunkSet set) {}
```
NotMyFault, line 30:
```suggestion
     * but MAY be flushed to history. Defaults to nothing as most Processors will not use it.
     *
     * @since TODO
     */
    default void postProcessSet(IChunk chunk, IChunkGet get, IChunkSet set) {}
```

NotMyFault (commented):
I gave it a try this morning and it worked so far well for lava and water. Defaulting to false seems like a good idea, considering this will have performance downsides on larger scaled servers.

worldedit-core/src/main/java/com/fastasyncworldedit/core/queue/IBatchProcessor.java
dordsor21: I suppose the method returning a Future ought to be kept actually, for API shizzle dizzle

NotMyFault: Likely, can deprecate it though.

NotMyFault (commented):
Can you exchange the the JB annotations to javax? (FA)WE uses them, but that is no review blocker.

NotMyFault (approved):

NotMyFault:
@SirYwell mind taking a look? 👀

dordsor21:
Recalcblockcounts is called in the constructor we use for LevelChunkSection 1.18. We should just switch to using this method for ticking all placed blocks as it can no longer be avoided (server restarts/chunk IO reload will make all blocks tick anyway)

worldedit-core/src/main/java/com/fastasyncworldedit/core/configuration/Settings.java
```diff
@@ -567,7 +567,8 @@ public FaweLimit getLimit(Actor actor) {
         public int DISCARD_AFTER_MS = 60000;
 
         @Comment({
-                "When using fastmode also do not bother to fix existing ticking blocks"
+                "When using fastmode also do not bother to fix existing ticking blocks",
+                "Only works in versions up to 1.17.2"
```
NotMyFault, line 571: 1.17.2?

worldedit-core/src/main/java/com/fastasyncworldedit/core/configuration/Settings.java
```diff
@@ -625,16 +626,10 @@ public FaweLimit getLimit(Actor actor) {
         public boolean OTHER = false;
 
         @Comment({
-                "Allow blocks placed by WorldEdit to tick. This could cause the big lags.",
+                "Allow fluids placed by FAWE to tick (flow). This could cause the big lags.",
```
NotMyFault, line 629: Maybe worth to add a note that ticked blocks are not captured in history and therefore not available for undo/redo.

worldedit-core/src/main/java/com/fastasyncworldedit/core/util/MultiFuture.java
```diff
@@ -0,0 +1,55 @@
+package com.fastasyncworldedit.core.util;
+
+import javax.annotation.Nonnull;
+import java.util.List;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+public class MultiFuture implements Future<Object[]> {
```
SirYwell, line 10: Is there anything that can't be covered by using `CompletableFuture#allOf(...)`? The current implementation of `get` with timeout here is insufficient in cases where e.g. the first future is the only one that isn't done but requires more than `timeout  / size` to complete.

dordsor21: You can't make a completable future from futures. I did look at using it, but using completeable future isn't ideal

worldedit-core/src/main/java/com/fastasyncworldedit/core/configuration/Settings.java
dordsor21: 1.18 will always call recalcBlockCounts as it's in the constructor for LevelChunkSection thus we can't "disable" it

NotMyFault: 2. is wrong then nevertheless 👀

dordsor21: Why?

NotMyFault:
Doesn't conflict, but needs to be updated for 1.18.2

NotMyFault:
Currently testing with PlotSquared 6.6.0 on 1.18.2: https://paste.gg/p/anonymous/ff5390b86a2a40dc82b4932327c474b7/files/10da529e8048487ea2c5942f1bbc13c5/raw

NotMyFault (approved):
Tested on 1.18.2 as well, works great 🎉

NotMyFault:
Conflicts

NotMyFault (approved):

PR #1566, opened 2022-01-25
fix: Don't synchronize DelegateSemaphore

### LINKED ISSUE #1565: DelegateSemaphore implements synchronized incorrectly
### Server Implementation

Spigot

### Server Version

1.18.1

### Describe the bug

https://github.com/IntellectualSites/FastAsyncWorldEdit/blob/main/worldedit-bukkit/src/main/java/com/fastasyncworldedit/bukkit/adapter/DelegateSemaphore.java

Both acquire() and release() are both synchronized methods, however this fails to account for when two threads are both accessing the same lock.

This is mostly noticeable to us as I often perform some laggy world saving operations.

### To Reproduce

Thread $1 calls acquire(), and starts doing some work.
$2 calls acquire(), but has to wait as $1 has the underlying lock.
$1 finishes doing some work, moves to call release()
$1 waits for $2 to finish calling acquire(), because the method is synchronized and release() cannot be called while $2 is calling acquire()

So because $1 has the underlying lock, $2 cannot finish calling acquire(), because $2 has the synchronized lock, $1 cannot call release()

### Expected behaviour

$1 calls acquire(), $2 calls acquire(), $1 is not blocked from calling release()

### Screenshots / Videos

_No response_

### Error log (if applicable)

https://gist.github.com/ec9b489ef20fe333000a8eb646dd0c7e

### Fawe Debugpaste

N/A

### Fawe Version

2.0.1-SNAPSHOT-76;e8f72d6

### Checklist

PierreSchwang:
## Overview

Fixes #1565

## Description
I could reproduce the issue using the following code snippet: https://h.pschwang.eu/iyeduwexin.cs
The semaphore couldn't be released as the synchronization blocks after the 2nd acquire from another thread.

The method doesn't need to be marked as synchronized itself - removing that seems to have fixed the issue - at least with my provided code snippet as test

Adapter stuff works fine nevertheless (e.g. regen)

## Submitter Checklist

## Maintainer checklist

Before the changes are marked as `Ready for merge`: 

- [ ] There are at least 2 approvals from project developers or maintainers for the pull request and no outstanding requests for change
- [ ] Conversations in the pull request are over OR it is explicit that a reviewer does not block the change
- [ ] Changelog entries in the PR title are correct
- [ ] Proper changelog labels are set so that the changelog can be generated automatically
- [ ] If the change needs additional modification steps from users, e.g. for permission changes, make sure the entry is added to the pull request description so it can be appended to the changelog.

### DIFF
--- a/worldedit-bukkit/src/main/java/com/fastasyncworldedit/bukkit/adapter/DelegateSemaphore.java
+++ b/worldedit-bukkit/src/main/java/com/fastasyncworldedit/bukkit/adapter/DelegateSemaphore.java
@@ -13,7 +13,7 @@ public class DelegateSemaphore extends Semaphore {
 
     // this is bad
     @Override
-    public synchronized boolean tryAcquire() {
+    public boolean tryAcquire() {
         try {
             this.delegate.acquire();
             return true;
@@ -23,12 +23,12 @@ public class DelegateSemaphore extends Semaphore {
     }
 
     @Override
-    public synchronized void acquire() throws InterruptedException {
+    public void acquire() throws InterruptedException {
         this.delegate.acquire();
     }
 
     @Override
-    public synchronized void release() {
+    public void release() {
         this.delegate.release();
     }
 


dordsor21:
This is an incorrect fix for the issue. There should not be synchronisation on unlock, but the synchronisation should remain on the lock. FAWE synchronized upon the DelegateSemaphore to prevent it being locked, and calling an "until free" method within the synchronisation to force priority to itself, whilst preventing locking by the server. The release method should therefore not be synchronised, and a brute-forced "until free" method added.

PierreSchwang:
> @PierreSchwang What do you think about applying the suggested changes?

Yeah, mb - wasn't quite sure how to implement that behavior, I could try that tho

PR #1581, opened 2022-01-27
Support PlotSquared's expanded world heights

dordsor21:


NotMyFault:
`[won't build]` 🥴

dordsor21:
Meant for it to be draft

NotMyFault:
P2's release version is 6.6.0 that contains these changes, just FYI.

NotMyFault:
Don't need to force push everything, we can squash merge to truncate history too :P

dordsor21:
Was only small changes that fit in the original commit

worldedit-bukkit/src/main/java/com/fastasyncworldedit/bukkit/regions/plotsquared/FaweDelegateSchematicHandler.java
```diff
@@ -65,6 +69,19 @@ public void paste(
             final int zOffset,
             final boolean autoHeight,
             final RunnableVal<Boolean> whenDone
+    ) {
+        paste(schematic, plot , xOffset, yOffset, zOffset, autoHeight, null, whenDone);
+    }
+
+    public void paste(
```
NotMyFault, line 76: Is that API and needed for P2? Otherwise we could make it private.

dordsor21: P2 will use the undeprecated method when the PR is merged, then we can delete the method that delegates with a null actor (or leave it)

NotMyFault (approved):

SirYwell (approved):

PR #1619, opened 2022-02-16
Add null check to Heightmap processor "just in case"

### LINKED ISSUE #1592: Cannot load from char array because "getSection" is null
### Server Implementation

Paper

### Server Version

1.16.5

### Describe the bug

Randomly, when inserting the schematic, an error occurs that seems to be related to lighting processing.
I'm using WorldEdit API to insert the schematic and seem to be doing everything right using a single thread pool that does something like this:

```
try (EditSession es = WorldEdit.getInstance().newEditSession(w)) {
                Operation op = new ClipboardHolder(cb)
                        .createPaste(es)
                        .to(BlockVector3.at(loc.getX(), loc.getY(), loc.getZ()))
                        .ignoreAirBlocks(false)
                        .copyEntities(false)
                        .copyBiomes(false)
                        .build();

                Operations.complete(op);
            }
```
            
And everything works, however, from time to time, the schematic is inserted incorrectly and an error occurs.

### To Reproduce

The error happens randomly, there is no way to repeat it all the time.

### Expected behaviour

Correct insertion of schematics without errors

### Screenshots / Videos

_No response_

### Error log (if applicable)

https://pastebin.com/z5XhDgZd

### Fawe Debugpaste

https://athion.net/ISPaster/paste/view/509a03a0c2c9480bb04f91fcc4920819

### Fawe Version

2.0.2-SNAPSHOT-84;faa2477

### Checklist

### Anything else?

_No response_

dordsor21:
Fixes #1592

### DIFF
--- a/worldedit-core/src/main/java/com/fastasyncworldedit/core/extent/processor/heightmap/HeightmapProcessor.java
+++ b/worldedit-core/src/main/java/com/fastasyncworldedit/core/extent/processor/heightmap/HeightmapProcessor.java
@@ -86,8 +86,9 @@ public class HeightmapProcessor implements IBatchProcessor {
                         } else if (getSection == null) {
                             getSection = get.load(layer);
                             // skip empty layer
-                            if (Arrays.equals(getSection, FaweCache.INSTANCE.EMPTY_CHAR_4096)
-                                    || Arrays.equals(getSection, AIR_LAYER)) {
+                            if (getSection == null ||
+                                    Arrays.equals(getSection, FaweCache.INSTANCE.EMPTY_CHAR_4096) ||
+                                    Arrays.equals(getSection, AIR_LAYER)) {
                                 hasSectionGet = false;
                                 if (!hasSectionSet) {
                                     continue layer;


NotMyFault (approved):

worldedit-core/src/main/java/com/fastasyncworldedit/core/extent/processor/heightmap/HeightmapProcessor.java
```diff
@@ -86,8 +86,9 @@ public IChunkSet processSet(IChunk chunk, IChunkGet get, IChunkSet set) {
                         } else if (getSection == null) {
                             getSection = get.load(layer);
                             // skip empty layer
-                            if (Arrays.equals(getSection, FaweCache.INSTANCE.EMPTY_CHAR_4096)
-                                    || Arrays.equals(getSection, AIR_LAYER)) {
+                            if (getSection == null ||
```
SirYwell, line 89: `Arrays.equals` already covers that in this case, although it might make sense to have this check, as other overloads of that method would indeed throw a NPE. That said, it would probably make sense to have a comment about it in the code.

dordsor21: That's true. The NPE in the issue was probably a race condition

SirYwell (approved):

NotMyFault:
This PR is ready to be merged, we will merge it in ~24 hours if there is no negative feedback, as documented by the [merge process](https://github.com/IntellectualSites/IntellectualSites/blob/main/docs/MAINTAINERS_CONTRIBUTORS.adoc#merge-process) documentation.

dordsor21:
This isn't an actual fix for the issue in question

PR #1668, opened 2022-03-18
Remove hardcoding of heightmap compressed long size

dordsor21:


SirYwell (approved):

NotMyFault:
Does that affect 1.16.5 too? Considering we maintain that version still, would be nice to have it sorted there as well, if applicable.

dordsor21:
It's not applicable because there's no way to change world heights in 1.16

NotMyFault:
> It's not applicable because there's no way to change world heights in 1.16

Ah ok, I don't know when the datapack thing got around.

NotMyFault (approved):

Issue #504, opened 2020-06-12
Can't handle unloding worlds


janLindner:
# Bug report for FastAsyncWorldEdit 1.15.2

**[REQUIRED] FastAsyncWorldEdit Configuration Files:** 
https://athion.net/ISPaster/paste/view/215c17dda504430db13c62cb7c57349e

**[REQUIRED] FastAsyncWorldEdit Version Number:** 
FastAsyncWorldEdit-257

**[REQUIRED] Spigot/Paper Version Number:** 
git-Paper-348

**[REQUIRED] Description of the problem:**
It seems like fawe can't handle unloading and reloading worlds. It seems like this issue is linked to the garbage collector clearing the weak reference of the world ref: https://github.com/IntellectualSites/FastAsyncWorldEdit/blob/1.15/worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitWorld.java

Exception: https://hastebin.com/mefamosoqo.cs
Ingame Exception: ![image](https://user-images.githubusercontent.com/18546571/84548531-1ad81800-ad06-11ea-817c-396c8a40f1d9.png)

**How to replicate**:
- Edit something in a world, that is not the default "world"
(Undo the operation)
- Unload the world
- Wait until the Garbage Collector clears the weak reference of the world
- Load the world again
- Edit something in the world --> exception in the log
(Try to undo the operation --> exception in log and chat)

**Plugins being used on the server:**
Arceon, dynmap, FAWE, goBrush, goPaint, HeadDatabase, ProtocolLib, Vault, VoxelSniper, WorldEdit

**Checklist**:

- [x] I included all information required in the sections above
- [x] I made sure there are no duplicates of this report [(Use Search)](https://github.com/IntellectualSites/FastAsyncWorldEdit/issues?q=is%3Aissue)
- [x] I made sure I am using an up-to-date version of [FastAsyncWorldEdit for 1.15.2](https://ci.athion.net/job/FastAsyncWorldEdit-1.15/)
- [x] I made sure the bug/error is not caused by any other plugin

PR #1712, opened 2022-04-20
Implement getNameUnsafe method to allow an unloaded world's name to be accessed

### LINKED ISSUE #1671: History (e.g. //undo) sometimes stops working completely when previous worlds are unloaded
### Server Implementation

Paper

### Server Version

1.18.2

### Describe the bug

When using a plugin to manage multiple worlds (e.g. MultiVerse or [BuildSystem](https://www.spigotmc.org/resources/buildsystem-1-8-1-18.60441/)) which unloads worlds when not in use, sporadically `//undo` stops working.
The circumstances to evoke this bug might seem a bit convoluted (see steps to reproduce) but the bug has occured multiple times in practice.

### To Reproduce

1. Setup a PaperMC 1.18.2 Server with FAWE and MultiVerse
2. Create a new world e.g. "test" with `/mv create test NORMAL -t FLAT` (generation options are not relevant)
3. Go into the newly generated world (`/mv tp test`)
4. Do any worldedit operation that is saved in the history (e.g. `//pos1`, `//pos2`, `//set stone`)
5. Go back to the default world and unload the test world (`/mv tp world`, **`/mv unload test`**)
6. The problem now is that FAWE [caches the world object with a WeakReference](https://github.com/IntellectualSites/FastAsyncWorldEdit/blob/465c81d193ceeb7a2568677e76903bc611c1d4a3/worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitWorld.java#L118) before trying to get the world by name via the Bukkit API so we need to:
6.1. **Either** force garbage collection a few times for example with [VisualVM](https://visualvm.github.io/), clutter the heap by flying around, loading chunks, wait a few minutes and pray to your favorite god(s)
6.2. **Or** replace [this line](https://github.com/IntellectualSites/FastAsyncWorldEdit/blob/main/worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitWorld.java#L177) with `World tmp = null;` for testing
7. Do another worldedit operation that impacts history (an exception should be thrown in the console but the player won't notice anything is wrong yet)
8. Try to `//undo` this last operation. This will fail with an exception (`java.lang.NullPointerException: The world was unloaded and the reference is unavailable`) and the changes won't be undone.

### Expected behaviour

No exceptions should be thrown and //undo should undo the latest changes.

### Screenshots / Videos

_No response_

### Error log (if applicable)

https://gist.github.com/Schuwi/072b8abd3b4861d21c89165963835c97

### Fawe Debugpaste

https://athion.net/ISPaster/paste/view/a26d3e931462486c87a96737b4ade9f5

### Fawe Version

FastAsyncWorldEdit version 2.1.1-SNAPSHOT-134;465c81d

### Checklist

### Anything else?

_No response_

dordsor21:
- Fixes #1671 and #504

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/world/World.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/world/World.java
@@ -63,11 +63,21 @@ public interface World extends Extent, Keyed, IChunkCache<IChunkGet> {
 //FAWE end
 
     /**
-     * Get the name of the world.
+     * Get the name of the world. This will error if world has been unloaded by the server.
      *
      * @return a name for the world
      */
     String getName();
+
+    //FAWE start - allow history to read an unloaded world's name
+    /**
+     * Get the name of the world. If the world referenced has been unloaded, this will still return the name.
+     *
+     * @return a name for the world
+     */
+    String getNameUnsafe();
+    //FAWE end
+
 
     /**
      * Get the folder in which this world is stored. May return null if unknown
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/world/World.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/world/World.java
@@ -74,6 +74,7 @@ public interface World extends Extent, Keyed, IChunkCache<IChunkGet> {
      * Get the name of the world. If the world referenced has been unloaded, this will still return the name.
      *
      * @return a name for the world
+     * @since TODO
      */
     String getNameUnsafe();
     //FAWE end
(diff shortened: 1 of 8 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/world/World.java
```diff
@@ -63,12 +63,22 @@ public interface World extends Extent, Keyed, IChunkCache<IChunkGet> {
...
      *
      * @return a name for the world
      */
     String getName();
 
+    //FAWE start - allow history to read an unloaded world's name
+    /**
+     * Get the name of the world. If the world referenced has been unloaded, this will still return the name.
+     *
+     * @return a name for the world
```
NotMyFault, line 76:
```suggestion
     * @return a name for the world
     * @since TODO
```

NotMyFault (approved):

SirYwell (approved):

PR #1715, opened 2022-04-21
Some fixes and improvements to internal queuing

### LINKED ISSUE #1664: Large scale replace (either replacenear or replace) generating weird results with an angle mask
### Server Implementation

Paper

### Server Version

1.18.2

### Describe the bug

When replacing something on a larger scale with an angle mask, some weird patterns arise. Seen in the images provided below, lines of the block to be replaced appear on the chunk borders. This issue didn't seem to happen in 1.17 FAWE.

### To Reproduce

1. Gmask an angle, I tried #angle and /[][] (/gmask /[0d][55d])
2. Replace anything with anything, in my case (//r 19 green_wool)

### Expected behaviour

The blocks will replace to the new block, but streaks of the old block will persist around the chunk borders.

### Screenshots / Videos

image
image

### Error log (if applicable)

N/A

### Fawe Debugpaste

https://athion.net/ISPaster/paste/view/1a68a9a07e744361bce12bc22f281aba

### Fawe Version

FastAsyncWorldEdit 2.1.1-SNAPSHOT-132

### Checklist

### Anything else?

_No response_

### LINKED ISSUE #1700: //fixwater causing terrain damage
### Server Implementation

Paper

### Server Version

1.18.2

### Describe the bug

When using fixwater in a large area, it can cause weird issues with cutting large holes in the terrain.

### To Reproduce

Use //fixwater with a large radius (500 is what I used)

### Expected behaviour

Solid blocks should never be affected by this command

### Screenshots / Videos

image
image
This is the result I've been seeing on my test server.

### Error log (if applicable)

_No response_

### Fawe Debugpaste

https://athion.net/ISPaster/paste/view/6f7f6f440832443888997caa6ef6d723

### Fawe Version

FastAsyncWorldEdit-153

### Checklist

### Anything else?

_No response_

dordsor21:
- The set array given in NMSAdapter should have get data written to it
 - Having target size >= 2* parallel threads allows for adjacent chunks to be loaded with issues
 - "empty" chunk section doesn't need to be localised to the chunk and may be static
 - Switch to slightly more performant stream method for testing for non-empty sections
 - Implement lock into ChunkHolder preventing any modification to occur whilst the edit is being applied to the world (when ChunkHolder is called)

 - Fixes #1664
 - May fix #1700

### DIFF
--- a/worldedit-core/src/main/java/com/fastasyncworldedit/core/queue/implementation/chunk/ChunkHolder.java
+++ b/worldedit-core/src/main/java/com/fastasyncworldedit/core/queue/implementation/chunk/ChunkHolder.java
@@ -1,6 +1,7 @@
 package com.fastasyncworldedit.core.queue.implementation.chunk;
 
 import com.fastasyncworldedit.core.FaweCache;
+import com.fastasyncworldedit.core.concurrent.ReentrantWrappedStampedLock;
 import com.fastasyncworldedit.core.configuration.Settings;
 import com.fastasyncworldedit.core.extent.filter.block.ChunkFilterBlock;
 import com.fastasyncworldedit.core.extent.processor.EmptyBatchProcessor;
@@ -25,8 +26,6 @@ import java.util.Map;
 import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.Future;
-import java.util.concurrent.locks.Lock;
-import java.util.concurrent.locks.ReentrantLock;
 
 /**
  * An abstract {@link IChunk} class that implements basic get/set blocks.
@@ -44,7 +43,7 @@ public class ChunkHolder<T extends Future<T>> implements IQueueChunk<T> {
         return POOL.poll();
     }
 
-    private final Lock calledLock = new ReentrantLock();
+    private final ReentrantWrappedStampedLock calledLock = new ReentrantWrappedStampedLock();
 
     private IChunkGet chunkExisting; // The existing chunk (e.g. a clipboard, or the world, before changes)
     private IChunkSet chunkSet; // The blocks to be set to the chunkExisting
@@ -79,8 +78,13 @@ public class ChunkHolder<T extends Future<T>> implements IQueueChunk<T> {
      * If the chunk is currently being "called", this method will block until completed.
      */
     private void checkAndWaitOnCalledLock() {
-        if (calledLock.tryLock()) {
-            calledLock.unlock();
+        if (calledLock.isLocked()) {
+            synchronized (this) {
+                if (calledLock.isLocked()) {
+                    calledLock.lock();
+                    calledLock.unlock();
+                }
+            }
         }
     }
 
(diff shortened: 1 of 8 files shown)


NotMyFault (approved):

worldedit-core/src/main/java/com/fastasyncworldedit/core/queue/implementation/chunk/ChunkHolder.java
```diff
@@ -64,66 +68,87 @@ public void init(IBlockDelegate delegate) {
...
 
     public synchronized IBlockDelegate getDelegate() {
         return delegate;
     }
 
+    /**
+     * If the chunk is currently being "called", this method will block until completed.
+     */
+    private void checkAndWaitOnCalledLock() {
+        if (calledLock.tryLock()) {
```
SirYwell, line 82: `tryLock` does not block, am I missing something?

worldedit-core/src/main/java/com/fastasyncworldedit/core/queue/implementation/chunk/ChunkHolder.java
```diff
@@ -992,9 +1038,19 @@ public synchronized <V extends IChunk> void init(IQueueExtent<V> extent, int chu
...
             chunkSet.setBitMask(bitMask);
-            return this.call(chunkSet, this::recycle);
+            try {
+                return this.call(chunkSet, () -> {
+                    recycle();
+                    calledLock.unlock();
+                });
+            } catch (Throwable t) {
+                calledLock.unlock();
+                t.printStackTrace();
```
SirYwell, line 1051: If the exception is rethrown, we shouldn't print it. 
How can we make sure that the callback/finalizer is actually called (on the same thread) to actually release the lock?

SirYwell (changes requested):

worldedit-core/src/main/java/com/fastasyncworldedit/core/queue/implementation/chunk/ChunkHolder.java
```diff
@@ -79,8 +78,13 @@ public synchronized IBlockDelegate getDelegate() {
...
     private void checkAndWaitOnCalledLock() {
-        if (calledLock.tryLock()) {
-            calledLock.unlock();
+        if (calledLock.isLocked()) {
+            synchronized (this) {
+                if (calledLock.isLocked()) {
+                    calledLock.lock();
+                    calledLock.unlock();
+                }
+            }
```
SirYwell, line 87: can't we just do lock unlock here?

dordsor21: yeah not sure why I tried to have some sort of optimisation here. I'll leave the first `isLocked` call there though as it's faster than locking and unlocking, and `checkAndWaitOnCalledLock` is called frequently

SirYwell (approved):

PR #1770, opened 2022-06-05
Add block break effect to area pickaxe

Owen1212055:
## Overview
Adds block break effect to area pickaxe.

## Description
Adds block break effect to area superpickaxe to match upstream/recursive pickaxe behavior.

I'm not sure what to do here as the commits that were responsible for adding this were reverted on a merge a _while_ back.
Added in: https://github.com/IntellectualSites/FastAsyncWorldEdit/commit/96e56bdd0cbf5d7640fa30984f78c27a084b6d05
Reverted in: https://github.com/IntellectualSites/FastAsyncWorldEdit/commit/66744bfaa4154569176e677ebb71a36470147d8b

## Submitter Checklist

- [X ] Make sure you are opening from a topic branch (**/feature/fix/docs/ branch** (right side)) and not your main branch.
- [] New public fields and methods are annotated with `@since TODO`.

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/command/tool/AreaPickaxe.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/command/tool/AreaPickaxe.java
@@ -72,13 +72,13 @@ public class AreaPickaxe implements BlockTool {
                 for (int x = ox - range; x <= ox + range; ++x) {
                     for (int y = oy - range; y <= oy + range && y <= maxY; ++y) {
                         for (int z = oz - range; z <= oz + range; ++z) {
-                            BlockVector3 pos = BlockVector3.at(x, y, z);
                             if (!initialType.equals(editSession.getBlock(x, y, z).getBlockType())) {
                                 continue;
                             }
 
                             editSession.setBlock(x, y, z, BlockTypes.AIR.getDefaultState());
 
+                            BlockVector3 pos = BlockVector3.at(x, y, z);
                             ((World) clicked.getExtent()).queueBlockBreakEffect(server, pos, initialType,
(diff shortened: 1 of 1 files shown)


NotMyFault:
> I'm not sure what to do here as the commits that were responsible for adding this were reverted on a merge a while back.

If there's no downside of adding it again and no associated issue, I guess we're good to go to reintegrate that?
Either it has been removed on purpose or on accident when local changes were different to upstream changes.

Owen1212055:
Yeah, it looked like it may have been removed by accident.

NotMyFault (approved):

worldedit-core/src/main/java/com/sk89q/worldedit/command/tool/AreaPickaxe.java
```diff
@@ -76,6 +78,9 @@ public boolean actPrimary(Platform server, LocalConfiguration config, Player pla
 
                             editSession.setBlock(x, y, z, BlockTypes.AIR.getDefaultState());
 
+                            BlockVector3 pos = BlockVector3.at(x, y, z);
+                            ((World) clicked.getExtent()).queueBlockBreakEffect(server, pos, initialType,
+                                    clicked.toVector().toBlockPoint().distanceSq(pos));
```
dordsor21, line 83: This should probably only be done if the setBlock method returns true

Owen1212055: It should be noted upstream doesn't check for this, but sure.

dordsor21: In that case also ensure to wrap in //FAWE start and end comments

PR #1778, opened 2022-06-07
build: Switch to bom

NotMyFault:


worldedit-bukkit/src/main/java/com/fastasyncworldedit/bukkit/regions/plotsquared/FaweQueueCoordinator.java
```diff
@@ -182,6 +188,16 @@ public void setLightingMode(@Nullable LightingMode mode) {
 
     }
 
+    @Override
+    public @org.checkerframework.checker.nullness.qual.Nullable SideEffectSet getSideEffectSet() {
+        return null;
+    }
+
+    @Override
+    public void setSideEffectSet(@org.checkerframework.checker.nullness.qual.Nullable final SideEffectSet sideEffectSet) {
```
dordsor21, line 197: We can just switch to javax here

PR #1781, opened 2022-06-10
Fix DelegateSemaphore synchronisation issues on Spigot

### LINKED ISSUE #1565: DelegateSemaphore implements synchronized incorrectly

dordsor21:
- Also effectively nullify it on paper - the synchronisation on the object is enough
 - Fixes #1565 
 - Supersedes #1566

### DIFF
--- a/worldedit-bukkit/src/main/java/com/fastasyncworldedit/bukkit/adapter/DelegateSemaphore.java
+++ b/worldedit-bukkit/src/main/java/com/fastasyncworldedit/bukkit/adapter/DelegateSemaphore.java
@@ -1,8 +1,6 @@
 package com.fastasyncworldedit.bukkit.adapter;
 
-import java.util.Collection;
 import java.util.concurrent.Semaphore;
-import java.util.concurrent.TimeUnit;
 
 public class DelegateSemaphore extends Semaphore {
 


SirYwell (approved):

NotMyFault (approved):
Looks good so far, the same change needs to be applied onto `feat/main/1.19` for the 1.19 adapter stuff too.

NotMyFault:
Ignoring the 24h cooldown here to prevent possible merge conflicts with the 1.19 branch.

PR #1789, opened 2022-06-12
Only implement cache to MaskingExtent when off main thread

### LINKED ISSUE #1681: `//gmask 0` causes `ArrayIndexOutOfBoundsException`
### Server Implementation

Paper

### Server Version

1.18.2

### Describe the bug

I have created a custom command, similar to `//forest` except it uses schematics to create trees, not ones provided by vanilla.
The command works perfectly fine in small regions, but whenever the region is larger than **123x123**, the command no longer works and a `java.lang.ArrayIndexOutOfBoundsException` is thrown.

### To Reproduce

1. The command (`//schemforest`) is created with the following classes:
* **SchematicForestCommand:** https://paste.gg/p/anonymous/eff55070a7274ccebc694468c8a39549
* **SchematicForestGenerator:** https://paste.gg/p/anonymous/495396e9f4e44f5cb1daee52dfd59c77
* **ThreadLocalRandomNoise:** https://paste.gg/p/anonymous/de2ab1cddcd34ff9be7ec7ee977b24ae

2. Select a large region (anything greater than 123x123)

3. Use the command `//schemforest <base> <schematic>`
* **\<base>:** Block type the schematic will be pasted on (e.g. `grass_block`)
* **\<schematic>:** The schematic to paste

4. Error is thrown and edit session will not paste
 

### Expected behaviour

No error is thrown and the schematics are pasted

### Screenshots / Videos

_No response_

### Error log (if applicable)

https://paste.gg/p/anonymous/c0e32b38ac1c4b008e6e6ad5cc3a2277

### Fawe Debugpaste

https://athion.net/ISPaster/paste/view/850861b448d44c44bdc4f71c8a0f1cd0

### Fawe Version

FastAsyncWorldEdit version 2.1.1-SNAPSHOT-139;4970db0

### Checklist

### Anything else?

_No response_

dordsor21:
- It's possible for a [Chunk/Char]FilterBlock to be used multiple times in the same "tree" of method calls meaning the mutable paramets (particularly index) get increased within a loop, causing AIOOBs and other issues
 - This is only possible on the main thread due to the handling of submissions in SingleThreadQueueExtent, as it ensures all operation remains on the main thread to prevent deadlocks
 - Therefore safe usage of FilterBlocks requires that main-threaded operation create a new [Chunk/Char]FilterBlock instance each time
 - Further fixes #1681

### DIFF
--- a/worldedit-core/src/main/java/com/fastasyncworldedit/core/FaweCache.java
+++ b/worldedit-core/src/main/java/com/fastasyncworldedit/core/FaweCache.java
@@ -18,6 +18,7 @@ import com.fastasyncworldedit.core.util.collection.CleanableThreadLocal;
 import com.google.common.cache.CacheBuilder;
 import com.google.common.cache.CacheLoader;
 import com.google.common.cache.LoadingCache;
+import com.plotsquared.core.util.AnnotationHelper;
 import com.sk89q.jnbt.ByteArrayTag;
 import com.sk89q.jnbt.ByteTag;
 import com.sk89q.jnbt.CompoundTag;
@@ -135,6 +136,32 @@ public enum FaweCache implements Trimable {
                 return withInitial.apply(key);
             }
         });
+    }
+
+    /**
+     * Create a new cache aimed to act as a thread-cache that is safe to the main server thread. If the method is called away
+     * from the main thread, it will return a {@link Function} referencing the {@link LoadingCache} returned by
+     * {@link FaweCache#createCache(Supplier)}. If it is called from the main thread, it will return a {@link Function} that
+     * will always return the result of the given {@link Supplier}. It is designed to prevent issues caused by
+     * interally-mutable and resettable classes such as {@link com.fastasyncworldedit.core.extent.filter.block.CharFilterBlock}
+     * from causing issues when used in edits on the main thread.
+     *
+     * @param withInitial The supplier used to determine the initial value if a thread cache is created, else to provide the
+     *                    a new instance of the class being cached if on the main thread.
+     * @return a {@link Function} referencing a cache, or the given {@link Supplier}
+     * @since TODO
+     */
+    @AnnotationHelper.ApiDescription(info = "Designed for specific internal use.")
+    public <V> Function<Long, V> createMainThreadSafeCache(Supplier<V> withInitial) {
+        return new com.google.common.base.Function<>() {
+            private final LoadingCache<Long, V> loadingCache = Fawe.isMainThread() ? null : FaweCache.INSTANCE.createCache(
+                    withInitial);
+
+            @Override
+            public V apply(final Long input) {
+                return loadingCache != null ? loadingCache.getUnchecked(input) : withInitial.get();
+            }
+        };
     }
 
     /*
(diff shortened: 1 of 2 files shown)


NotMyFault (approved):

worldedit-core/src/main/java/com/fastasyncworldedit/core/FaweCache.java
```diff
@@ -137,6 +138,32 @@ public V load(@Nonnull T key) {
         });
     }
 
+    /**
+     * Create a new cache aimed to act as a thread-cache that is safe to the main server thread. If the method is called away
+     * from the main thread, it will return a {@link Function} referencing the {@link LoadingCache} returned by
+     * {@link FaweCache#createCache(Supplier)}. If it is called from the main thread, it will return a {@link Function} that
+     * will always return the result of the given {@link Supplier}. It is designed to prevent issues caused by
+     * interally-mutable and resettable classes such as {@link com.fastasyncworldedit.core.extent.filter.block.CharFilterBlock}
```
NotMyFault, line 146:
```suggestion
     * internally-mutable and resettable classes such as {@link com.fastasyncworldedit.core.extent.filter.block.CharFilterBlock}
```

worldedit-core/src/main/java/com/fastasyncworldedit/core/FaweCache.java
```diff
@@ -137,6 +138,32 @@ public V load(@Nonnull T key) {
...
+     * from causing issues when used in edits on the main thread.
+     *
+     * @param withInitial The supplier used to determine the initial value if a thread cache is created, else to provide the
```
NotMyFault, line 149:
```suggestion
     * @param withInitial The supplier used to determine the initial value if a thread cache is created, else to provide
```

worldedit-core/src/main/java/com/fastasyncworldedit/core/FaweCache.java
```diff
@@ -137,6 +138,32 @@ public V load(@Nonnull T key) {
...
+     *                    a new instance of the class being cached if on the main thread.
+     * @return a {@link Function} referencing a cache, or the given {@link Supplier}
+     * @since TODO
+     */
+    @AnnotationHelper.ApiDescription(info = "Designed for specific internal use.")
```
NotMyFault, line 154:
```suggestion
    @AnnotationHelper.ApiDescription(info = "Designed for specific internal use only.")
```

NotMyFault (commented):
Non blocking post approval feedback.

worldedit-core/src/main/java/com/fastasyncworldedit/core/FaweCache.java
dordsor21: I left out "only" has it could be a useful method for API use, however, it's advising that it's not really a general API method

worldedit-core/src/main/java/com/fastasyncworldedit/core/FaweCache.java
```diff
@@ -137,6 +138,32 @@ public V load(@Nonnull T key) {
...
+     * from causing issues when used in edits on the main thread.
+     *
+     * @param withInitial The supplier used to determine the initial value if a thread cache is created, else to provide a new
+     *                    instance of the class being cached if on the main thread.
+     * @return a {@link Function} referencing a cache, or the given {@link Supplier}
+     * @since TODO
+     */
+    @AnnotationHelper.ApiDescription(info = "Designed for specific internal use.")
+    public <V> Function<Long, V> createMainThreadSafeCache(Supplier<V> withInitial) {
+        return new com.google.common.base.Function<>() {
```
SirYwell, line 156: Is there are reason for using a google Function here? Also, can't we use a `LongFunction` instead?

SirYwell (changes requested):

worldedit-core/src/main/java/com/fastasyncworldedit/core/FaweCache.java
dordsor21: I had meant for it to be a java.util function, but LongFunction would make more sense

SirYwell (approved):

Issue #1729, opened 2022-05-07
//img max load time / max image size


ayunami2000:
### What feature do you want to see added?

I would like //img to not load forever. Very large images essentially crash/hang worldedit forever, making it literally unusable. If there could be an image size (e.g. able to set it to 2K max res) or an amount of time to timeout the img command, then this plugin would be a lot more stable for allowing players to freely use the //img command on their own.

### Are there any alternatives?

No alternatives.

PR #1790, opened 2022-06-12
Implement limits to image size and load times

dordsor21:
- Prevents issues caused by users attempting to load large images
 - Implements #1729


NotMyFault (approved):

worldedit-core/src/main/java/com/sk89q/worldedit/command/GenerationCommands.java
```diff
@@ -587,11 +594,31 @@ public void image(
...
-                    RenderingHints.VALUE_INTERPOLATION_BILINEAR, false
-            );
+        ExecutorService executor = Executors.newSingleThreadExecutor();
+        Future<BufferedImage> future = executor.submit(() -> {
+            BufferedImage image = MainUtil.readImage(url);
+            if (dimensions != null) {
+                checkCommandArgument(
+                        (long) dimensions.getX() * dimensions.getZ() <= Settings.settings().WEB.MAX_IMAGE_SIZE,
+                        Caption.of("fawe.error.image-dimensions", Settings.settings().WEB.MAX_IMAGE_SIZE)
+                );
```
SirYwell, line 604: This check could be done before starting a new thread, however it doesn't prevent someone from loading a large image without a dimension argument.

dordsor21: The image itself is checked in the MainUtil method

SirYwell: I see, missed that.

SirYwell (approved):

PR #1818, opened 2022-06-15
Don't limit split size for extra block data (signs, spawner, etc.)

dordsor21:


NotMyFault (approved):
Untested, but looks fine.

SirYwell (approved):

PR #1822, opened 2022-06-16
Bring stack in line with upstream using offset rather than direction

### LINKED ISSUE #1349: //stack has no option for offset

FAWE's /stack implementation lacks an option for an offset between the stacked copies of the selection, while in plain WorldEdit a x,y,z vector is allowed for the direction parameter, which allows for offsetting the copy by a certain number of blocks.

Example command: //stack 1 512,0,0 will stack on x axis with offset=512 (includes selection)

Workarounds exist but this should be implemented as it's present upstream.

### Anything else?

_No response_

dordsor21:
- Fixes #1349


NotMyFault:
Looks good, but the build fails due a javadoc.io outage.

SirYwell (approved):

PR #1859, opened 2022-06-29
docs: Use rootProject.name over project.name for javadocs title

NotMyFault:
As the title says, our javadocs are now named "FastAsyncWorldEdit + version", as appropriate.

### DIFF
--- a/buildSrc/src/main/kotlin/CommonJavaConfig.kt
+++ b/buildSrc/src/main/kotlin/CommonJavaConfig.kt
@@ -67,6 +67,7 @@ fun Project.applyCommonJavaConfiguration(sourcesJar: Boolean, banSlf4j: Boolean 
                     "https://jd.papermc.io/paper/1.18/",
                     "https://intellectualsites.github.io/fastasyncworldedit-javadocs/worldedit-core/"
             )
+            docTitle = "${rootProject.name}" + " " + "${rootProject.version}"
         }
     }
 


buildSrc/src/main/kotlin/CommonJavaConfig.kt
```diff
@@ -67,6 +67,7 @@ fun Project.applyCommonJavaConfiguration(sourcesJar: Boolean, banSlf4j: Boolean
                     "https://jd.papermc.io/paper/1.18/",
                     "https://intellectualsites.github.io/fastasyncworldedit-javadocs/worldedit-core/"
             )
+            docTitle = "${rootProject.name}-${project.description}" +  " " + "${rootProject.version}"
```
SirYwell, line 70: Is there a reason why we're mixing string interpolation and string concatenation?

SirYwell (changes requested):

buildSrc/src/main/kotlin/CommonJavaConfig.kt
NotMyFault: That's the same format used everywhere else, e.g. see the pom generator.

SirYwell: I see. I guess it would make sense to clean that up at some point, but that can be done separately

SirYwell (approved):

PR #1881, opened 2022-07-20
fix: use servers removeGameEventListener instead of reimpl

peaches94:
## Overview

there is an issue on servers that might have a different type for game event listeners (https://github.com/Bloom-host/Petal/issues/6)
this fix is also good for fawe because its less maintenance of internal code that can change. the generic stuff doesnt matter for reflection

## Description
replaces the copy of removeGameEventListener with calling it directly

## Submitter Checklist


NotMyFault (commented):
If applicable, this logic should be applied retroactively to older versions we support as well.

peaches94:
@NotMyFault 😊

NotMyFault (approved):
Thanks for addressing the outstanding concerns

dordsor21 (approved):

PR #1884, opened 2022-07-23
feature: support paper chunk system

SirYwell:
## Description

This PR adds support for the new chunk system in paper.

Current state: Feel free to compile and test it yourself.

### Issues (and whether they are fixed)
- [x] Changes aren't sent to the client due to a UnsupportedOperationException
- [x] Copying/pasting entities fails due to missing field 

## Submitter Checklist


anonyco:
By the look of it, all this does is remove the `Optional` object holder and remove the call to `chunkHolder.getTickingChunkFuture().getNow(ChunkHolder.UNLOADED_LEVEL_CHUNK)`, which is what actually loads in the chunk.

So, the error you are getting about `UnsupportedOperationException` is actually just FAWE trying to operate on the chunk that it couldn't load in.

SirYwell:
> By the look of it, all this does is remove the `Optional` object holder and remove the call to `chunkHolder.getTickingChunkFuture().getNow(ChunkHolder.UNLOADED_LEVEL_CHUNK)`, which is what actually loads in the chunk.

No, it does not load the chunk as we call `getNow`. It's basically our current way to get the chunk without loading it on spigot. The method isn't implemented in the new chunk system at the moment, see https://github.com/PaperMC/Paper/blob/4ec2e6e64fea5fdad52e156625252031cff1aef5/patches/server/0929-Rewrite-chunk-system.patch#L14553-L14554.

We also don't really need it, as `getChunkAtIfLoadedImmediately` does exactly what we want on paper. We can also probably just use `getTickingChunk()` directly, but I'd need to look into that first.

SirYwell:
I'd like to integrate this in the next release.

NotMyFault (approved):
Works fine based on a quick test.

NotMyFault:
Expediting merge due to no negative feedback.

PR #1896, opened 2022-08-03
Update plugin.yml

### LINKED ISSUE #1814: Loaded class com.sk89q.worldguard.protection.association.RegionAssociable from WorldGuard v7.0.7+216b061 which is not a depend or softdepend of this plugin
### Server Implementation

Paper

### Server Version

1.19

### Describe the bug

The following warning is displayed on server start:

`[16:20:00 WARN]: [FastAsyncWorldEdit] Loaded class com.sk89q.worldguard.protection.association.RegionAssociable from WorldGuard v7.0.7+216b061 which is not a depend or softdepend of this plugin.`

Incomplete existing issue #865

### To Reproduce

1. Download the plugins:
a. [Paper](https://api.papermc.io/v2/projects/paper/versions/1.19/builds/17/downloads/paper-1.19-17.jar)
b. [FAWE](https://ci.athion.net/job/FastAsyncWorldEdit/lastSuccessfulBuild/artifact/artifacts/FastAsyncWorldEdit-Bukkit-2.3.1-SNAPSHOT-208.jar)
c. [WorldGuard](https://dev.curseforge.com/projects/worldguard/files/3677516/download)
2. Run server

### Expected behaviour

No warning is printed

### Screenshots / Videos

_No response_

### Error log (if applicable)

https://paste.gg/p/anonymous/0d3ae6730a794d488b44fc85e56aba3f

### Fawe Debugpaste

https://athion.net/ISPaster/paste/view/001f6c5153be4eb0b3ea5c3d0ca59581

### Fawe Version

FastAsyncWorldEdit version 2.3.1-SNAPSHOT-208;cb6e200

### Checklist

### Anything else?

_No response_

Firetop:
## Overview

Update softdepend list to include WorldGuard to stop the Loaded class error message: [ Vault, WorldGuard ]

## Submitter Checklist


Firetop:
test this please

NotMyFault:
How did you test this out to make sure it doesn't affect dependencies negatively in circular and arbitrary environments?
The bukkit plugin loader system isn't the greatest handling those cases.

Firetop:
> How did you test this out to make sure it doesn't affect dependencies negatively in circular and arbitrary environments? The bukkit plugin loader system isn't the greatest handling those cases.

All I did was modify the plugin.yml in FAWE and load it in paper on a Windows machine to confirm the expected functionality of both plugins (FAWE and WorldGuard) and that the Class Loaded Error message didn't appear. I didn't do anymore testing than that.

NotMyFault:
Looks fine to me.

Firetop:
No rush of course; just a suggestion as I know people will poke about the warning if they see it.

dordsor21:
soft-depend shouldn't cause circular dependencies unless something goes very very wrong as it has no impact on load order. The trouble with load order where `provides` is used (i.e. in FAWE) has also been fixed iirc?

Firetop:
I'm surprised this hasn't been fixed yet, not sure what we are waiting for.

The softdepend is very well documented with Spigot and Bukkit, not sure what else would be expected of this report, FAWE devs just needs to do a bit of Googling.

I've been running this change on my own production servers since I originally reported this months ago without issues.

SirYwell:
1. `softdepend` actually affects plugin loading order. If A softdepends on B, B is loaded before A (if B exists). The difference to `depend` is that A will load even if B is not present.
2. The only reason this *currently* works is because the circular dependency check does not understand the `provides` section. 	
3. There are a lot of other factors playing into this, like OS, file system, JDK version and other things we have no control over which could make this fail at any point. Just because it works *right now* and *for you* does not mean it is correct.
4. The actual issue is a *warning*, not an *error*. It is meant to help with finding causes of a bug, but it isn't a bug itself.

I don't want to rely on the absolutely flawed mess of the current plugin loading system of spigot. Especially as the functionality only can get worse with this change.

Owen1212055:
It should be noted that paper has plans to streamline the plugin loading process, see (https://github.com/PaperMC/Paper/pull/8108).

This will hopefully reduce any weird behavior but as a result plugin loading will no longer support circular plugin dependencies.

If I'm understanding this correctly, do you want to be able to add dependencies like this whilst not effecting the load order?

SirYwell:
The problem is that we actually *have* a circular dependency. WorldGuard depends on WorldEdit (or FAWE providing WE), and FAWE optionally depends on WorldGuard (and also depends on it at compile time). 

I hadn't the time to look into your Paper PR much, but I think the underlying problem is that plugins can randomly load classes from other plugins. In best case, this could be controlled, and no warnings for "unintended" class loading from other plugins are needed anymore.

Owen1212055:
I see, so for FAWE it requires it during compile time therefore has to mark it as a dependency.

I defo see the use case when you need two plugins to be able to "integrate" with each other. 

I think that eliminating this circular dependency behavior then might be problematic for plugins that do stuff like this, obviously in your case you rely on it as well.

Do you think then you'd benefit from another dependency-like configuration?

Current: hard depends,depends, and loads before (all effects loading order)

What if, for example, FAWE could use a new "Soft Utilizes" (idk a better name) option where FAWE would be able to say that it "utilizes" WorldGuard during runtime. This however wouldn't effect load order, so it would allow plugins to have a relationship like this. Is this something you'd benefit by? Or does FAWE rely on having worldguard on startup? 


Would appreciate your feedback here. 😄

Issue #1902, opened 2022-08-07
Macro Command - IllegalStateException


BlockyTheDev:
### Server Implementation

Paper

### Server Version

1.18.2

### Describe the bug

An Exception occures and an information in the chat with the value "Please report this error: [See console]"

### To Reproduce

1. Run command `//macro` without any arguments
2. Look into chat and console

### Expected behaviour

No Exception

### Screenshots / Videos

<img width="1442" alt="image" src="https://user-images.githubusercontent.com/86119630/183313453-f87f7be8-d4ff-4cef-a5be-2ccf563fd28c.png">

### Error log (if applicable)

https://gist.github.com/BlockyTheDev/0edad899411389b329fc99636d71b5c0

### Fawe Debugpaste

https://athion.net/ISPaster/paste/view/f94e0cf66147415c8c53ba5e70b44c5a

### Fawe Version

FastAsyncWorldEdit version 2.4.3

### Checklist

- [X] I have included a Fawe debugpaste.
- [X] I am using the newest build from https://ci.athion.net/job/FastAsyncWorldEdit/ and the issue still persists.

### Anything else?

While i wanted to run a command from an other plugin i wrote one `/` too much and then found the exception on a 1.16.5 server. So I went on an testing server with 1.18.2 and tried it out there too, which worked.

NotMyFault:
The command is currently not implemented. What is it supposed to do?

BlockyTheDev:
> The command is currently not implemented. What is it supposed to do?

I don't know. I executed it by mistake while tabbing through commands and saw the error and thaught I am going to report it because the message said it.

BlockyTheDev:
I think it was removed by upstream at any point or added by FAWE, because I can't find the command in worldedit in the last commit's code.

PR #1911, opened 2022-08-19
fix: Lazy fix macro stacktrace

NotMyFault:
The command is commented out since 2019, given there is no forseeable change anytime soon, we don't need to keep it up as-is.

Fixes https://github.com/IntellectualSites/FastAsyncWorldEdit/issues/1902


SirYwell (approved):

NotMyFault:
Expediting further reviews to cut a release.

PR #1922, opened 2022-08-23
 fix: avoid recursive LegacyMapper initialization

SirYwell:
## Description

When using a FAWE version that doesn't support the server version, FAWE started causing issues and spamming the logs because the LegacyMapper wasn't initialized correctly, and while initializing, the parse method was used, causing another initialization (and so on). While it will still fail to load correctly, it won't spam the logs anymore over and over again.

The parsing logic in BlockTypes should probably removed altogether, but that needs some investigation on what's needed, so using the parser context seems to be a quick and easy solution.

## Submitter Checklist


NotMyFault (approved):
I have no such world at hand at the moment, but we can surely verify the integrity within the next MC release.

dordsor21 (approved):

PR #1944, opened 2022-09-12
Fix #saturate-Pattern

### LINKED ISSUE #1933: #saturate not working after looking for wrong json files in the textures folder
### Server Implementation

Spigot

### Server Version

1.19.2

### Describe the bug

Using #saturate in any instance where it would be applicable, such as replacing some existing terrain with a saturation effect, (//r #saturate[156][100][0][120]), results in an error. After looking at console, there were various blocks that the plugin tried to check for in fawe's texture folder and it could not find apparently. These can be seen in the debugpaste I've provided. I double-checked that the jar I provided was the latest, and it was (1.19.2.jar). It appears that what all of the blocks share in common is that they all have some form of block state. I looked into the jar file through winrar to see if some of these block states were in the blockstates folder (1.19.2.jar\assets\minecraft\blockstates) and they were.

I suppose either my textures folder I uploaded was incorrect, although I obtained it from my own .minecraft/versions folder, or the plugin is just looking for the incorrect json files for the proper saturation. To counter the former, I deleted the textures jar I provided and let the plugin download the jar from mojang (https://athion.net/ISPaster/paste/view/5ccd411609ff47a3825b4afa1aaa0b1b). The same error seemed to persist.

### To Reproduce

- Select an area
- Try and apply some saturation pattern to it (//r #saturate[156][100][0][120])
- Receive error

### Expected behaviour

The saturation effect should work as normal, instead of giving an error.

### Screenshots / Videos

_No response_

### Error log (if applicable)

_No response_

### Fawe Debugpaste

https://athion.net/ISPaster/paste/view/3f1b1ddc83eb44779f10f29eda6a31ce

### Fawe Version

FastAsyncWorldedit version 2.4.5-SNAPSHOT-275;5da558e

### Checklist

### Anything else?

_No response_

PierreSchwang:
## Overview
Fixes #1933 

## Description
1. Updated the asset url to the last release
2. Search through item models as well (No idea why, but many block state models are located in models/item)
3. Change BlockType access by using the BlockTypeCache directly - as the used ids are based on the indexes in the values array of BlockTypeCache. No idea if this changes anything, was just more logical to me.

Example for `#saturate[255][0][120][255]`
grafik
grafik

Based on my quick testing the changes works so far - Should logs and leaves be replaced as well? For me they don't, but I've never worked with that pattern before, so not that sure what the expected outcome is tbh

## Submitter Checklist

### DIFF
--- a/worldedit-core/src/main/java/com/fastasyncworldedit/core/util/TextureUtil.java
+++ b/worldedit-core/src/main/java/com/fastasyncworldedit/core/util/TextureUtil.java
@@ -30,7 +30,6 @@ import java.io.BufferedInputStream;
 import java.io.BufferedReader;
 import java.io.File;
 import java.io.FileNotFoundException;
-import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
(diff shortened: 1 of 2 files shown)


worldedit-core/src/main/java/com/fastasyncworldedit/core/util/TextureUtil.java
```diff
@@ -810,7 +810,7 @@ public void loadModTextures() throws IOException {
                 new File(Fawe.platform().getDirectory() + "/" + Settings.settings().PATHS.TEXTURES + "/")
                         .mkdirs();
                 try (BufferedInputStream in = new BufferedInputStream(
-                        new URL("https://piston-data.mojang.com/v1/objects/c0898ec7c6a5a2eaa317770203a1554260699994/client.jar")
+                        new URL("https://piston-data.mojang.com/v1/objects/055b30d860ead928cba3849ba920c88b6950b654/client.jar")
```
NotMyFault, line 813: Is this 1.19.2?

PierreSchwang: Yep

NotMyFault: Thanks!

NotMyFault (approved):
Thanks for addressing!

worldedit-core/src/main/java/com/fastasyncworldedit/core/util/TextureUtil.java
SirYwell, line 813: We already downloaded the client jar at this point if I see this right, couldn't we therefore get rid of this somehow?

PierreSchwang: Gonna rebase the pr, yes - Alex cherry picked the client jar download from a commit I've made in this PR

SirYwell (approved):
Looks good

PR #1947, opened 2022-09-15
Retrieve the latest asset release version automatically

NotMyFault:
(cherry picked from commit a8885a349a567849f6db29565cc65b14e3dab155)

### DIFF
--- a/worldedit-core/src/main/java/com/fastasyncworldedit/core/util/TextureUtil.java
+++ b/worldedit-core/src/main/java/com/fastasyncworldedit/core/util/TextureUtil.java
@@ -40,7 +40,7 @@ import java.net.URL;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
-import java.security.DigestInputStream;
+import java.security.DigestOutputStream;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 import java.util.ArrayList;
@@ -648,10 +648,8 @@ public class TextureUtil implements TextureHolder {
     private static String calculateSha1(Path path) throws NoSuchAlgorithmException, IOException {
         MessageDigest digest = MessageDigest.getInstance("SHA-1");
         try (final BufferedInputStream stream = new BufferedInputStream(Files.newInputStream(path));
-             final DigestInputStream digestInputStream = new DigestInputStream(stream, digest)) {
-            // noinspection StatementWithEmptyBody - update digest with bytes from passed input stream
-            while (digestInputStream.read() != -1) {
-            }
+             final DigestOutputStream digestOutputStream = new DigestOutputStream(OutputStream.nullOutputStream(), digest)) {
+            stream.transferTo(digestOutputStream);
             return HexFormat.of().formatHex(digest.digest());
         }
     }
--- a/worldedit-core/src/main/java/com/fastasyncworldedit/core/util/TextureUtil.java
+++ b/worldedit-core/src/main/java/com/fastasyncworldedit/core/util/TextureUtil.java
@@ -29,6 +29,7 @@ import java.awt.image.BufferedImage;
 import java.io.BufferedInputStream;
 import java.io.BufferedReader;
 import java.io.File;
+import java.io.FileInputStream;
 import java.io.FileNotFoundException;
 import java.io.FileOutputStream;
 import java.io.IOException;
@@ -36,7 +37,14 @@ import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.lang.reflect.Type;
 import java.net.URL;
+import java.nio.channels.Channels;
+import java.nio.channels.FileChannel;
+import java.nio.channels.ReadableByteChannel;
 import java.nio.charset.StandardCharsets;
+import java.nio.file.Path;
+import java.security.DigestInputStream;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
@@ -327,6 +335,7 @@ public class TextureUtil implements TextureHolder {
             new BiomeColor(255, "Unknown Biome", 0.8f, 0.4f, 0x92BD59, 0x77AB2F)};
 
     private static final String VERSION_MANIFEST = "https://piston-meta.mojang.com/mc/game/version_manifest.json";
+    private static final byte[] HEX_ARRAY = "0123456789abcdef".getBytes(StandardCharsets.US_ASCII);
     private final BlockType[] layerBuffer = new BlockType[2];
     protected int[] blockColors = new int[BlockTypes.size()];
     protected long[] blockDistance = new long[BlockTypes.size()];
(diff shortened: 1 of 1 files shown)


PierreSchwang (approved):
Not sure how I should review my code, so I'd say LGTM? :D

NotMyFault:
I outsourced this change, to allow independent feedback and reviews from the other PR.

worldedit-core/src/main/java/com/fastasyncworldedit/core/util/TextureUtil.java
```diff
@@ -352,17 +358,22 @@ public TextureUtil(File folder) throws FileNotFoundException {
...
+                    try (BufferedInputStream in = new BufferedInputStream(new URL(url).openStream());
+                         FileOutputStream fileOutputStream = new FileOutputStream(
+                                 Fawe.platform().getDirectory() + "/" + Settings.settings().PATHS.TEXTURES +
+                                         "/" + metadata.version() + ".jar")) {
+                        byte[] dataBuffer = new byte[1024];
+                        int bytesRead;
+                        while ((bytesRead = in.read(dataBuffer, 0, 1024)) != -1) {
+                            fileOutputStream.write(dataBuffer, 0, bytesRead);
+                        }
+                        LOGGER.info("Asset jar down has been downloaded successfully.");
```
NotMyFault, line 375: I see you copied that from the line below, but grammar wise, the sentence doesn't make much sense.

I have no idea what I meant by that.

worldedit-core/src/main/java/com/fastasyncworldedit/core/util/TextureUtil.java
```diff
@@ -321,6 +325,8 @@ public class TextureUtil implements TextureHolder {
             new BiomeColor(253, "Unknown Biome", 0.8f, 0.4f, 0x92BD59, 0x77AB2F),
             new BiomeColor(254, "Unknown Biome", 0.8f, 0.4f, 0x92BD59, 0x77AB2F),
             new BiomeColor(255, "Unknown Biome", 0.8f, 0.4f, 0x92BD59, 0x77AB2F)};
+
+    private static final String VERSION_MANIFEST = "https://piston-meta.mojang.com/mc/game/version_manifest.json";
```
NotMyFault, line 329: Given this is now a hard-coded variable, I'd be in favor of verifying the SHA1 of the client jar we download.

worldedit-core/src/main/java/com/fastasyncworldedit/core/util/TextureUtil.java
```diff
@@ -352,17 +367,47 @@ public TextureUtil(File folder) throws FileNotFoundException {
...
+                            Fawe.platform().getDirectory().getPath(),
+                            Settings.settings().PATHS.TEXTURES,
+                            metadata.version() + ".jar"
+                    ).toString();
+                    // Copy resource to local fs
+                    try (final InputStream stream = new URL(resource.resource()).openStream();
+                         final ReadableByteChannel inChannel = Channels.newChannel(stream);
+                         final FileOutputStream out = new FileOutputStream(outPath);
+                         final FileChannel outChannel = out.getChannel()) {
+                        outChannel.transferFrom(inChannel, 0, Long.MAX_VALUE);
```
SirYwell, line 386:
```suggestion
                        stream.transferTo(out);
```

That allows to also remove the Channel resources.

worldedit-core/src/main/java/com/fastasyncworldedit/core/util/TextureUtil.java
```diff
@@ -555,6 +600,85 @@ private static long getDistance(BufferedImage image, int c1) {
...
+    }
+
+    /**
+     * Transforms the bytes into a readable hexadecimal string (
+     * <a href="https://stackoverflow.com/a/9855338/12620913">Stackoverflow</a>)
+     * @param bytes The bytes to convert into hex
+     * @return The hexadecimal representation of the bytes
+     * @since TODO
+     */
+    private static String toHexString(byte[] bytes) {
```
SirYwell, line 672: We can simply use [`HexFormat`](https://docs.oracle.com/en/java/javase/18/docs/api/java.base/java/util/HexFormat.html#formatHex(byte%5B%5D)) here

worldedit-core/src/main/java/com/fastasyncworldedit/core/util/TextureUtil.java
```diff
@@ -555,6 +600,85 @@ private static long getDistance(BufferedImage image, int c1) {
...
+     */
+    private static String calculateSha1(File file) throws NoSuchAlgorithmException, IOException {
+        MessageDigest digest = MessageDigest.getInstance("SHA-1");
+        try (final BufferedInputStream stream = new BufferedInputStream(new FileInputStream(file));
+             final DigestInputStream digestInputStream = new DigestInputStream(stream, digest)) {
+            // noinspection StatementWithEmptyBody - update digest with bytes from passed input stream
+            while (digestInputStream.read() != -1) {
+            }
+            return toHexString(digest.digest());
+        }
```
SirYwell, line 662:
```suggestion
        return toHexString(digest.digest(Files.readAllBytes(file.toPath())));
```
In general, using Path instead of File where possible would be great.

worldedit-core/src/main/java/com/fastasyncworldedit/core/util/TextureUtil.java
```diff
@@ -352,17 +367,47 @@ public TextureUtil(File folder) throws FileNotFoundException {
...
-                    int bytesRead;
-                    while ((bytesRead = in.read(dataBuffer, 0, 1024)) != -1) {
-                        fileOutputStream.write(dataBuffer, 0, bytesRead);
+
+                try {
+                    VersionMetadata metadata = getLatestVersion();
+                    LOGGER.info("Latest release version is {}", metadata.version());
+                    HashedResource resource = getLatestClientJarUrl(metadata);
+
+                    String outPath = Path.of(
```
SirYwell, line 376: We can use Path here and `Files.newOutputStream(outPath)` below.

SirYwell (changes requested):

worldedit-core/src/main/java/com/fastasyncworldedit/core/util/TextureUtil.java
PierreSchwang: The reason for the usage of Channel was the significant file size of the client jar. By using channel the file may be copied directly, without buffering the whole jar file (even in chunks) into the memory. 

> This method is potentially much more efficient than a simple loop that reads from the source channel and writes to this channel. Many operating systems can transfer bytes directly from the source channel into the filesystem cache without actually copying them.

SirYwell: That mainly applies to two file channels on one system, and even then the proper stream implementations just delegate to channels.

worldedit-core/src/main/java/com/fastasyncworldedit/core/util/TextureUtil.java
SirYwell: This still uses the DigestInputStream, but from a second look, reading the bytes lazy instead all at once might be better - but I think we can also use DigestOutputStream and use transferTo here too? This would at least avoid the while loop.

SirYwell (approved):
Looks good.

NotMyFault (commented):
Approved. 

Thanks!

PR #1693, opened 2022-04-05
Fix memory leak by Citizen NPCs

PierreSchwang:
## Description
Fixes a memory leak that attaches and caches PermissionAttachments to NPCs. Citizens uses version 2 UUIDs, which are not used for regular player uuids. Those v2 uuids are now excluded.

## Submitter Checklist

- [x] Make sure you are opening from a topic branch (**/feature/fix/docs/ branch** (right side)) and not your main branch.
- [x] Ensure that the pull request title represents the desired changelog entry.
- [x] I tested my changes and approved their functionality.
- [x] I ensured my changes do not break other parts of the code
- [x] I read and followed the [contribution guidelines](https://github.com/IntellectualSites/.github/blob/main/CONTRIBUTING.md)

PR #1969, opened 2022-10-02
Don't error when NPC-annotated "entity" leaves the server

### LINKED ISSUE #1966: IllegalArgumentException exception for Quitting NPC's (Given attachment is not part of Permissible object CraftPlayer{name=Dispenser-Block})
### Server Implementation

Paper

### Server Version

1.19.2

### Describe the bug

When Players with the Metadata `NPC` leave the game it causes the PermissibleBase to throw IllegalArgumentException

### To Reproduce

1. Set metadata "NPC" to true for your player
2. Leave the game
3. Observe result

(I have also tested this with a Regular player entity (myself))

### Expected behaviour

For FAWE to handle quit events of NPC players without error.

### Screenshots / Videos

_No response_

### Error log (if applicable)

https://paste.gg/p/anonymous/272029b3304047a3ae1ee1f38c0e3d2e

### Fawe Debugpaste

https://athion.net/ISPaster/paste/view/b21c2fc395e84377a2d1134fdb8db9a3

### Fawe Version

FastAsyncWorldEdit version 2.4.7-SNAPSHOT-282;87f6806

### Checklist

### Anything else?

I'm not exactly sure why you expect the Player's attachments to contain your new attachment, as nowhere in your code adds the attachment from com.fastasyncworldedit.bukkit.BukkitPermissionAttachmentManager#getOrAddAttachment

My only solution would be to check the caller in the quit event and patch it on my end by removing the NPC meta when passed to WorldEdit. But this seems to be more WE's issue than mine.

Cheers.

PierreSchwang:
## Overview
Fixes #1966

## Description
The Noop-PermissibleAttachment is never added to the entity - but rather shared by reference - therefore can't be removed and throws error in certain edge cases (#1966)

## Submitter Checklist

### DIFF
--- a/worldedit-bukkit/src/main/java/com/fastasyncworldedit/bukkit/BukkitPermissionAttachmentManager.java
+++ b/worldedit-bukkit/src/main/java/com/fastasyncworldedit/bukkit/BukkitPermissionAttachmentManager.java
@@ -36,10 +36,6 @@ public class BukkitPermissionAttachmentManager {
         if (p == null) {
             return;
         }
-        if (p.hasMetadata("NPC") && noopAttachment != null) {
-            p.removeAttachment(noopAttachment);
-            return;
-        }
         PermissionAttachment attach = attachments.remove(p);
         if (attach != null) {
             p.removeAttachment(attach);


NotMyFault (commented):
This change was part of addressing a memory leak with Citizens, see https://github.com/IntellectualSites/FastAsyncWorldEdit/pull/1693.

Is this no longer the case?

PierreSchwang:
It is still the case, but has a logic flaw. The attachment shouldn't be removed, as it isn't added in the first place.
grafik
It's just the noop-attachment which is shared by reference, but never added to the entity in first place.

NotMyFault (approved):

dordsor21 (approved):

PR #1975, opened 2022-10-08
fix: Add blocked blocks to the correct set

dordsor21:


worldedit-core/src/main/java/com/fastasyncworldedit/core/extent/DisallowedBlocksExtent.java
```diff
@@ -137,8 +134,11 @@ public IChunkSet processSet(final IChunk chunk, final IChunkGet get, final IChun
             char[] blocks = Objects.requireNonNull(set.loadIfPresent(layer));
             it:
             for (int i = 0; i < blocks.length; i++) {
-                char block = blocks[i];
-                BlockState state = states[block];
+                char block;
+                if ((block = blocks[i]) == BlockTypesCache.ReservedIDs.__RESERVED__) {
```
SirYwell, line 138: I'd prefer to avoid this assignment expression and initialize `block` directly instead

SirYwell (changes requested):

SirYwell (approved):

NotMyFault (approved):

PR #1992, opened 2022-10-23
fix: bring default min/max y in line with upstream in AbstractRegion

### LINKED ISSUE #1987: FAWE unable to select WorldGuard regions above Y:320 or below Y:-64.
### Server Implementation

Paper

### Server Version

1.19.2

### Describe the bug

When using a datapack that allows for increased build height, FAWE is perfectly capable of interacting with blocks and even allows the creation of WorldGuard regions that exist beyond the vanilla borders, but cannot properly select them when prompted. When attempting to select a WorldGuard region above Y:320 for example, the plugin will instead select the vanilla height limit at the same X and Z coordinates. This would not come as much of a surprise to me if not for the fact that WorldEdit is able to do this just fine.

### To Reproduce

1) Install WorldGuard, FAWE, and a region visualizer of your choice to help with the test.
2) Use a plugin or datapack that allows for increased build height. 
3) Select a region above the vanilla height limit and save it as a WorldGuard region.
4) Unselect the region and then attempt to select it again through WorldGuard.
5) You cannot.
6) Switch out FAWE for WorldEdit.
7) Repeat steps 3 and 4 to see the difference.

### Expected behaviour

I expected the WorldGuard region I'd created above the vanilla height limit to be selectable while using FAWE, but it is not. It is selectable, however, when using WorldEdit.

### Screenshots / Videos

This is me attempting to select my WorldGuard region at Y:1024 with FAWE: https://streamable.com/xpffmg

This is me doing it again, but with WorldEdit: https://streamable.com/0iqv4b

### Error log (if applicable)

_No response_

### Fawe Debugpaste

https://athion.net/ISPaster/paste/view/4bde1412bb774c4cb075375099c4b780

### Fawe Version

FastAsyncWorldEdit-293

### Checklist

### Anything else?

While I do not know for certain, I should mention that I believe this difference between FAWE and WorldEdit may be caused by a single setting that exists within WorldEdit's configuration file, but not FAWE's.
image
Simply pasting the option into FAWE's worldedit-config.yml understandably does nothing whatsoever.
image

dordsor21:
- Fixes #1987

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/regions/AbstractRegion.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/regions/AbstractRegion.java
@@ -223,23 +223,11 @@ public abstract class AbstractRegion extends AbstractSet<BlockVector3> implement
     // Sub-class utilities
 
     protected final int getWorldMinY() {
-        //FAWE start > Server default based on version
-        return world == null ? WorldEdit
-                .getInstance()
-                .getPlatformManager()
-                .queryCapability(Capability.WORLD_EDITING)
-                .versionMinY() : world.getMinY();
-        //FAWE end
+        return world == null ? Integer.MIN_VALUE : world.getMinY();
     }
 
     protected final int getWorldMaxY() {
-        //FAWE start > Server default based on version
-        return world == null ? WorldEdit
-                .getInstance()
-                .getPlatformManager()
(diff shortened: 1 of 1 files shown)


SirYwell (approved):

PierreSchwang (approved):

NotMyFault (approved):
Untested

PR #2018, opened 2022-11-15
Add CF upload task

NotMyFault:
Copied from https://github.com/IntellectualSites/fastasyncvoxelsniper/blob/cf45f0b4aabeb9f2c0091908b60e743689ba261c/.github/workflows/build.yml#L71


PierreSchwang (approved):
LGTM - worked on FAVS, why shouldn't it here 🕵🏽‍♂️

PierreSchwang:
> > LGTM - worked on FAVS, why shouldn't it here 🕵🏽‍♂️
> 
> YOLO, testing in prod then in a bit. Can always revert

grafik

PR #2021, opened 2022-11-19
Show affected biomes in `//copy -b`

### LINKED ISSUE #1999: "{0} biomes affected" is not / wrongly counted
### Server Implementation

Paper

### Server Version

1.19.2

### Describe the bug

Copying biomes with copy -b (and paste -b) won't work on my server. It won't work anywhere in any world. I tried few blocks and a lot of blocks. Every time it says 0 biomes affected. It works perfectly for copying entities.

### To Reproduce

I tried using normal Worldedit and it works nicely.  I also tried fresh fresh paper server with only FAWE plugin and it still didn't work. I also tried few different plugin versions.

### Expected behaviour

Since copying biomes works normally with normal Worldedit it should work just fine with FAWE.

### Screenshots / Videos

_No response_

### Error log (if applicable)

_No response_

### Fawe Debugpaste

https://athion.net/ISPaster/paste/view/fe00cd7769f142a4b4eb8b47ec6d1b7f

### Fawe Version

FastAsyncWorldEdit-Bukkit-2.4.9-SNAPSHOT-297

### Checklist

### Anything else?

_No response_

PierreSchwang:
## Overview
Fixes #1999 

## Description
- Fixes the counting for affected biomes
- Align RegionMaskingFilter with upstream, as FAWE changes are not used at all
- 	Fix runServer configuration for gradle plugin

tbh, I'm not sure if that's correct. The biome for every block is counted - is that the case for normal WE as well?
grafik

## Submitter Checklist

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -75,8 +75,8 @@ applyCommonConfiguration()
 
 tasks {
     runServer {
-        minecraftVersion("1.19")
-        pluginJars(project(":worldedit-bukkit").file("build/libs/FastAsyncWorldEdit-Bukkit-$version.jar"))
+        minecraftVersion("1.19.2")
+        pluginJars(project(":worldedit-bukkit").file("build/libs/FastAsyncWorldEdit-Bukkit-${rootVersion}-SNAPSHOT.jar"))
 
     }
 }
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/function/RegionMaskingFilter.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/function/RegionMaskingFilter.java
@@ -35,24 +35,32 @@ public class RegionMaskingFilter implements RegionFunction {
 
     private final RegionFunction function;
     private final Mask mask;
-    //FAWE start
-    private final Extent extent;
-    //FAWE end
 
     /**
      * Create a new masking filter.
      *
      * @param mask     the mask
      * @param function the function
+     * @deprecated Extent is not used
      */
-    //FAWE start - Extent
+    //FAWE start - Extent (unused -> to be removed)
     public RegionMaskingFilter(Extent extent, Mask mask, RegionFunction function) {
         checkNotNull(function);
         checkNotNull(mask);
-        //FAWE start
-        checkNotNull(extent);
-        this.extent = extent;
-        //FAWE end
+        this.mask = mask;
+        this.function = function;
+    }
+    // FAWE end - Extent
+
+    /**
+     * Create a new masking filter.
+     *
+     * @param mask the mask
+     * @param function the function
+     */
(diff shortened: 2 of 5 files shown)


build.gradle.kts
```diff
@@ -75,8 +75,8 @@ applyCommonConfiguration()
 
 tasks {
     runServer {
-        minecraftVersion("1.19")
-        pluginJars(project(":worldedit-bukkit").file("build/libs/FastAsyncWorldEdit-Bukkit-$version.jar"))
+        minecraftVersion("1.19.2")
+        pluginJars(project(":worldedit-bukkit").file("build/libs/FastAsyncWorldEdit-Bukkit-${rootVersion}-SNAPSHOT.jar"))
```
dordsor21, line 79: hmm?

worldedit-core/src/main/java/com/sk89q/worldedit/function/RegionMaskingFilter.java
```diff
@@ -35,24 +35,32 @@ public class RegionMaskingFilter implements RegionFunction {
...
     /**
      * Create a new masking filter.
      *
      * @param mask     the mask
      * @param function the function
+     * @deprecated Extent is not used
      */
-    //FAWE start - Extent
+    //FAWE start - Extent (unused -> to be removed)
     public RegionMaskingFilter(Extent extent, Mask mask, RegionFunction function) {
```
dordsor21, line 47: Add a `@deprecated` tag with forRemoval = true

worldedit-core/src/main/java/com/sk89q/worldedit/function/operation/ForwardExtentCopy.java
```diff
@@ -394,17 +394,23 @@ public Operation resume(RunContext run) throws WorldEditException {
...
         }
 
+        if (copyingBiomes && (source.isWorld() || region instanceof FlatRegion)) {
+            BiomeCopy biomeCopy = new BiomeCopy(source, finalDest);
+            RegionFunction biomeFunction = sourceMask == Masks.alwaysTrue() ?
+                    biomeCopy : new RegionMaskingFilter(sourceMask, biomeCopy);
+            RegionVisitor regionVisitor = new RegionVisitor(region, biomeFunction, source);
+            Operations.completeBlindly(regionVisitor);
+            this.affectedBiomeCols += regionVisitor.getAffected();
+        }
```
dordsor21, line 412: Is it not possible to combine the biome and block copy into one? This implementation will require all chunks to be loaded twice and thus could make the copy take up to 2x as long for a large region

dordsor21:
There's no particuylarly graceful way of doing this. I think it should just be closed in favour of #2359 until a more robust solution is completed (tbh for such a minor thing, I don't think there's much need to put much effort in)

PR #2123, opened 2023-02-28
Fix FAWE using exclusive build height from PlotSquared

### LINKED ISSUE #2120: FAWE doesn't respect P2 build heights
### Server Implementation

Paper

### Server Version

1.19.3

### Describe the bug

In a plot world with defined heights of 0 - 255, FAWE will place blocks at Y=256. P2 does not allow the player to break these blocks because the height is exceeded. The screenshot below illustrates the problem.

AWE+WE respects the build height and only places blocks up to Y=255.

PlotSquared/config/worlds.yml contains the following:
```
    world:
      max_height: 256
      gamemode: creative
      min_height: 1
      border: true
      max_gen_height: 255
      min_gen_height: 0
```
This allows players to place/break blocks up to Y=255. Changing max_height did not affect FAWE's incorrect block placement.

### To Reproduce

**It's important to run this as a normal user, not as op or with special permissions.**
1. Create a server with P2+FAWE. Set up P2's worlds.yml as shown above.
2. Assuming ground level is at Y=64, select two points on the ground (I use //hpos1 and //hpos2). Then do `//expand 256 up` and `//set stone`. This will cause FAWE to attempt to place blocks up to Y=320.
3. Go to the top of the stone column and check the build heights via the client's F3 debug screen.
4. Observe that the uppermost block is at Y=256 and P2 will not allow you to break the block. P2 will allow you to break the block at Y=255.

### Expected behaviour

The stone column in the above example should only go up to Y=255.

### Screenshots / Videos

FAWE_bug

### Error log (if applicable)

_No response_

### Fawe Debugpaste

https://athion.net/ISPaster/paste/view/52208fd9a0e4438ea02e1c4bc22a5075

### Fawe Version

FastAsyncWorldEdit version 2.5.3-SNAPSHOT-368;c0dc42e

### Checklist

### Anything else?

_No response_

SirYwell:
## Overview

Fixes #2120 

## Description

(FastAsync)WorldEdit uses inclusive heights maximums whereas PlotSquared uses exclusive values. Using the values provided by PlotSquared therefore requires a subtraction by one.

## Submitter Checklist

### DIFF
--- a/worldedit-bukkit/src/main/java/com/fastasyncworldedit/bukkit/regions/plotsquared/PlotSquaredFeature.java
+++ b/worldedit-bukkit/src/main/java/com/fastasyncworldedit/bukkit/regions/plotsquared/PlotSquaredFeature.java
@@ -158,8 +158,10 @@ public class PlotSquaredFeature extends FaweMaskManager {
 
         Region maskedRegion;
         if (regions.size() == 1) {
-            int min = area != null ? area.getMinBuildHeight() : player.getWorld().getMinY();
-            int max = area != null ? Math.min(player.getWorld().getMaxY(), area.getMaxBuildHeight()) : player.getWorld().getMaxY();
+            final World world = player.getWorld();
+            int min = area != null ? area.getMinBuildHeight() : world.getMinY();
+            // PlotSquared uses exclusive max height, WorldEdit uses inclusive max height -> subtract 1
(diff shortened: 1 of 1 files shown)


NotMyFault (approved):

Bobcat00:
Looks good. Thanks, guys!

PR #2127, opened 2023-03-03
Use unsynchronized buffered OutputStream for history

SirYwell:
## Description

`BufferedOutputStream` has a fair amount of overhead as it's thread safe. However, we don't require thread safety here, so we can use an alternative implementation from fastutil which performs better. In my tests, the time spent writing history is halved.

## Submitter Checklist

### DIFF
--- a/worldedit-core/src/main/java/com/fastasyncworldedit/core/internal/io/FaweOutputStream.java
+++ b/worldedit-core/src/main/java/com/fastasyncworldedit/core/internal/io/FaweOutputStream.java
@@ -20,6 +20,13 @@ public class FaweOutputStream extends DataOutputStream {
         return parent;
     }
 
+    // overwritten to un-synchronized
+    @Override
+    public void write(final int b) throws IOException {
+        out.write(b);
+        written++;
+    }
+
     public void write(int b, int amount) throws IOException {
         for (int i = 0; i < amount; i++) {
             write(b);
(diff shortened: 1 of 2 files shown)


dordsor21:
It's probably worth adding a note somewhere that it's changed to a non-thread-safe implementation

SirYwell:
> It's probably worth adding a note somewhere that it's changed to a non-thread-safe implementation

Where would you think is the best place for such comment? I was thinking about the `processSet` method in AbstractChangeSet as that's the point that synchronizes atm. However, the whole design isn't thread safe, writing to the stream from different threads would have caused issues even before...

dordsor21:
Only because it's a public method in a util class, so probably a javadoc note on getCompressedOS

dordsor21 (approved):

SirYwell:
I saw that `DataOutputStream#write` is synchronized. Overriding and reimplementing that method brings even more improvements.

Overall, with `send-before-history: false` this brings down a `//set stone` on a region `(0, 0, 0) -> (1023, 319, 1023)` from >70s to <20s.

Because flamegraphs are cool, here a diff comparing the baseline to the changes on this branch (green: in baseline but not in patch, red: in patch but not in baseline. Unrelated calls are hidden)
flamegraph

NotMyFault (approved):

Issue #1928, opened 2022-09-01
Mixing up positions when using //pos1 & //pos2 from console


bobhenl:
### Server Implementation

Paper

### Server Version

1.18.2

### Describe the bug

Hello, so I've tried using plugin CommandTimer & MyCommand to create scheduled task where I want to reset the inner place of area where players can build. The reset is planned every 15 minutes and I have two arenas. With one arena everything was OK, but when I'm setting two different selections from console it's mixing up.
Images of arenas
![image](https://user-images.githubusercontent.com/18662086/187910671-91ff8bdf-2ad3-4377-864e-df5ff14be26f.png)
![image](https://user-images.githubusercontent.com/18662086/187910723-235bde0d-fb53-4061-9711-1d92ca05dd6d.png)

**First arena selection:** 
  - '//pos1 13,74,80'
  - '//pos2 -13,65,43'
**Second arena selection:**
  - '//pos1 10,69,18'
  - '//pos2 44,71,36'

In MyCommand file I'm using this sequence of commands with delay variable (1 minute delay) to be sure that it cannot combine, but it's still combined somehow.
```
 ResetArena:
  command: /resetarena
  type: RUN_CONSOLE
  runcmd:
  - '//world world'
  - '$delay$'
  - '//pos1 13,74,80'
  - '//pos2 -13,65,43'
  - '//pos1 13,74,80'
  - '//pos2 -13,65,43'
  - '$delay$'
  - '//pos1 13,74,80'
  - '//pos2 -13,65,43'
  - '$delay$'
  - '//cut'
  - '$delay$'
  - '$delay$'
  - '$delay$'
  - '//world world'
  - '$delay$'
  - '//pos1 10,69,18'
  - '//pos2 44,71,36'
  - '//pos1 10,69,18'
  - '//pos2 44,71,36'
  - '$delay$'
  - '//pos1 10,69,18'
  - '//pos2 44,71,36'
  - '$delay$'
  - '//cut'
  cost: 0
  delaytimer: 60
  delaytimer_format: SECONDS
  permission-required: true
  register: true
```
 
As you can see I've even tried to set multiple pos1 & pos2 same selections to solve this issue, but it still isn't. So, this is what's happening:

LOOK AT THE beginning on the coords of the first selection and second selection and it's POS1 & POS2.

![image](https://user-images.githubusercontent.com/18662086/187911307-d73e6272-8128-4450-99e5-68a3feaf903f.png)
[...]

PR #2131, opened 2023-03-05
Simplify command dispatch to ensure command order

### LINKED ISSUE #2130: Command aliases executed in wrong order
### Server Implementation

Paper

### Server Version

1.19.3

### Describe the bug

If a Bukkit command alias is defined (in commands.yml) with a sequence of FAWE commands, the commands are executed in an arbitrary order which changes every time.

AWE+WE does not have this problem.

### To Reproduce

1. Place the following in commands.yml:
```
aliases:
  squarene:
  - /pos1
  - /pos2
  - /shift 1 down
  - /expand $$1 n
  - /contract 1 s
  - /expand $$1 e
  - /contract 1 w
  - /walls glowstone
```
The intent of this command to to place a glowstone square of user-specified size in the ground, with the southwest corner of the square located under the player's feet.

2. Go to a location that has a suitably large section of flat ground. I used a plot world.

3. Enter the command `/squarene 16`. This should create a 16x16 square. You can use `//undo` to remove the glowstone and execute the command again. You'll see the commands are processed in a different order each time.

### Expected behaviour

The commands should be executed in the order in which they are defined. It should be deterministic.

### Screenshots / Videos

The chat area shows that the commands are executed in the wrong order.

First, here is the chat when the 8 commands are issued **manually**, showing the correct order of execution. Positions set, regions shifted, expand/contract, expand/contract, and finally 60 blocks are set.
FAWE_command_1

The following two screenshots show what happens when the command `/squarene 16` is issued. Note the commands are not executed in the correct order, and the sequence changes every time. FAWE is manipulating the region before the positions are set, and setting the blocks before the final region is defined.
FAWE_command_2

FAWE_command_3

### Error log (if applicable)

_No response_

### Fawe Debugpaste

https://athion.net/ISPaster/paste/view/e7eadac353f94e9198b76919ef0020de

### Fawe Version

FastAsyncWorldEdit version 2.5.3-SNAPSHOT-374;84e1886

### Checklist

### Anything else?

_No response_

SirYwell:
## Overview

Fixes #2130

I suspect #1928 was the same problem.

## Description

Before, FAWE submitted a task to the CraftScheduler just to actually handle it on a FAWE thread pool anyways. We can simplify that by skipping the first thread switch.

For commands that skip the queue, we still use the old behavior of submitting a task to the CraftScheduler.

I also removed the thread renaming part. It was pretty much useless this way, so I don't see any benefit in keeping it.

## Submitter Checklist

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/extension/platform/PlatformCommandManager.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/extension/platform/PlatformCommandManager.java
@@ -678,31 +678,26 @@ public final class PlatformCommandManager {
 
         Actor actor = event.getActor();
         String args = event.getArguments();
-        TaskManager.taskManager().taskNow(() -> {
-            if (!Fawe.isMainThread()) {
-                Thread.currentThread().setName("FAWE Thread for player: " + actor.getName());
+        int space0 = args.indexOf(' ');
+        String arg0 = space0 == -1 ? args : args.substring(0, space0);
+        Optional<Command> optional = commandManager.getCommand(arg0);
+        if (optional.isEmpty()) {
+            return;
+        }
+        Command cmd = optional.get();
+        PermissionCondition queued = cmd.getCondition().as(PermissionCondition.class).orElse(null);
+        if (queued != null && !queued.isQueued()) {
+            TaskManager.taskManager().taskNow(() -> handleCommandOnCurrentThread(event), Fawe.isMainThread());
+            return;
+        } else {
+            actor.decline();
+        }
+        actor.runAction(() -> {
+            SessionKey key = actor.getSessionKey();
+            if (key.isActive()) {
+                PlatformCommandManager.this.handleCommandOnCurrentThread(event);
             }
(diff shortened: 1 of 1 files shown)


Bobcat00:
I tested this PR, and it solves the problem I was having.

SirYwell:
> Or maybe not. I got this exception 10 minutes after doing the WE operations: [pastebin.com/KzGJ8g0x](https://pastebin.com/KzGJ8g0x)
> 
> I'll note that I compiled FAWE myself for this test.

From a first look, I suspect this is unrelated, although confusing. Sadly, the Exception doesn't contain any information how gson reached that field...

Bobcat00:
I can get it to happen, but not reliably. I can't narrow down the circumstances that cause it. It could very well be unrelated.

Bobcat00:
I've been running with this PR for 6 days. Several people have used FAWE, and I haven't see the above error again.

Bobcat00:
Hey, just wondering if this will be merged soon?

TheMeinerLP:
If you will. I can testing it locally.

dordsor21 (approved):

PR #2135, opened 2023-03-07
Drop Residence integration

NotMyFault:
This PR is a follow-up to https://github.com/IntellectualSites/FastAsyncWorldEdit/pull/2081:

Residence uses a non-standardized approach to build and distribute binaries, incompatible with the industry standard.
There have been community efforts on Residence's end to maven-ize the project, see `https://github.com/Zrips/Residence/issues/42`. Given there's been no progress in the past 6 years, I see no reason why we should continue maintaining our integration relying on our artifact, if the maintainer shows lack of motivation to move the community effort forward.

As shown by [telemetry data](https://bstats.org/plugin/bukkit/FastAsyncWorldEdit/1403#residence), the impact of removal is surveyable. To be fair, I wouldn't take the 2% affected as gospel truth because our integration works if Fawe and Residence are both installed, which does not imply that everyone using Fawe and Residence alongside necessarily use Fawe's Residence features.
Nevertheless, I'll add an entry to the upcoming changelog to communicate this change and the reasoning.


dordsor21:
Tbh unless it becomes a work-required feature, there's no need to remove residence support. If wherever we're pulling it in as a dependency has a compatible version with the latest residence releases then it's all good imo

OneLiteFeather:
About which Residence we talk? I found a fork on spigot that seemed active https://github.com/Zrips/Residence
You can find it on the same page the original came from.

dordsor21:
The issue is not how active it is, it's that it's to using any sort of build tooling, has no maven repository provided, etc. I suggest only removing residence at such a time the current integration no longer works, unless residence decides to become a useable open source project

PR #2138, opened 2023-03-07
Allow edits inside of allowed region when executing outside of region.

### LINKED ISSUE #2137: Working on plot from outside of it
**PlotSquared Feature Suggestion:**
Are we able to set this up so players can world edit inside their own plots while standing outside of it? This is important for 16x16 block plots as working from outside the plot is kinda necessary :slight_smile:

**How should it be implemented?**
Players should be able to run worldedit and Voxelsniper commands on areas inside their plots while standing outside of their plots.

**Additional context**
https://imgur.com/a/cbUebJE

**Checklist**:

- [x] I made sure there are no duplicates of my suggestion [(Use search)](https://github.com/IntellectualSites/PlotSquaredSuggestions/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+)
- [x] I made sure my suggestion does not exist yet [(Check the wiki)](https://github.com/IntellectualSites/PlotSquared/wiki)
- [x] I made sure my suggestion is useful for the majority of PlotSquared users

PierreSchwang:
## Overview

Fixes #2137

## Description

Allows to pass a custom base location to the EditSessionBuilder and various Mask-methods to determine the allowed regions for remote tools like brushes, which should not be based on the player location itself but the target location.

Pending:
- [ ] Blacklisted Regions
- [ ] How should //set / //replace operations be handled, if the command is executed outside the selected region (the selection intersects with an allowed region, but the players stands in a forbidden region -> fails)

## Submitter Checklist

### DIFF
--- a/worldedit-bukkit/src/main/java/com/fastasyncworldedit/bukkit/regions/GriefDefenderFeature.java
+++ b/worldedit-bukkit/src/main/java/com/fastasyncworldedit/bukkit/regions/GriefDefenderFeature.java
@@ -8,6 +8,7 @@ import com.sk89q.worldedit.bukkit.BukkitAdapter;
 import com.sk89q.worldedit.internal.util.LogManagerCompat;
 import com.sk89q.worldedit.math.BlockVector3;
 import com.sk89q.worldedit.regions.CuboidRegion;
+import com.sk89q.worldedit.util.Location;
 import org.apache.logging.log4j.Logger;
 import org.bukkit.entity.Player;
 import org.bukkit.event.Listener;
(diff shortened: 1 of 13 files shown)


PierreSchwang:
@IntellectualSites/fastasyncworldedit-team I'm not really sure, why a method documented as `Get the player's current allowed WorldEdit regions` only returns the allowed regions the player's location intersects with - but it would be quite resource heavy for various integrations to fetch all allowed regions (esp. PS).
Due to that, I've added a few methods to pass a custom location as the position to search for allowed regions. That works just fine with Brushes (other tools are missing yet) but actions for a selection are still an issue (see the 2nd open checklist entry). In that case I guess we shouldn't do any preprocessing checks but instead directly return the FaweMask for the extent itself. Not happy with that either, but the other way we would have to iterate over the the whole selection twice (once for the preprocessing checks and lastly for setting the blocks itself...)

dordsor21:
This could still be useful tbh, my PR requires people to enter the region first still

PierreSchwang:
> This could still be useful tbh, my PR requires people to enter the region first still

Possibly, yea. 
I guess another clean PR would be better, as masks as a whole thing have to be changed in a certain way. This PR hardly fixed brushes, but other actions may have problems as well.

PR #2150, opened 2023-03-14
Bump run paper to 1.19.3 and replace path with better solution

TheMeinerLP:
## Overview
Currently FAWE use for run paper 1.19 and a "static" path for get the jar file. In some cases that is not working.

## Description
I replaced the old way to get the jar after shadow with a better one that selects in many cases the jar for out of the box running/debugging.
## Submitter Checklist


NotMyFault (approved):
Thanks!

PR #2157, opened 2023-03-21
fix: invalidate cached FaweMask when a plot is unlinked

dordsor21:
- Fixes IntellectualSites/PlotSquared#3998

### DIFF
--- a/worldedit-bukkit/src/main/java/com/fastasyncworldedit/bukkit/regions/plotsquared/PlotSquaredFeature.java
+++ b/worldedit-bukkit/src/main/java/com/fastasyncworldedit/bukkit/regions/plotsquared/PlotSquaredFeature.java
@@ -23,6 +23,7 @@ import com.sk89q.worldedit.world.World;
 import org.apache.logging.log4j.Logger;
 import org.bukkit.Bukkit;
 
+import java.lang.ref.WeakReference;
 import java.util.List;
 import java.util.Locale;
 import java.util.Set;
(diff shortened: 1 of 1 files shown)


worldedit-bukkit/src/main/java/com/fastasyncworldedit/bukkit/regions/plotsquared/PlotSquaredFeature.java
```diff
@@ -176,15 +176,28 @@ public FaweMask getMask(Player player, MaskType type, boolean isWhitelist) {
...
-            @Override
-            public boolean isValid(Player player, MaskType type) {
-                if (Settings.Done.RESTRICT_BUILDING && DoneFlag.isDone(finalPlot)) {
-                    return false;
-                }
-                return isAllowed(player, finalPlot, type);
+        return new PlotSquaredMask(maskedRegion, finalPlot);
+    }
+
+    private final class  PlotSquaredMask extends FaweMask {
```
PierreSchwang, line 182: Duplicate space between class and PlotSquaredMask 😡

dordsor21: Yeah the whole thing kinda messy haha

SirYwell: heh
huh

but on a serious note, while I don't expect the hashCode to collide, it might be bad to rely on it - maybe it is fine to rely on reference equality of the sets? While that's not specified anywhere, we could add a note to the method in PlotSquared...

dordsor21: I suppose wrapping the set in a weak reference removes any possibility of memory "leak" as if the connected plots is reset we don't need the set anyway, we can just check the reference

SirYwell (approved):
Looks good

Ducki0210:
Hello, the bug seems to have been fixed but you can always cut the bridge using //undo

dordsor21:
> Hello, the bug seems to have been fixed but you can always cut the bridge using //undo

Do you have `restrict-history-to-regions` set to false in the config?

dordsor21:
Then set it to true.

NotMyFault (approved):
Nice catch

PR #2176, opened 2023-04-11
Fix wrong bit flipping of FAST_PLACEMENT limit for fawe

### LINKED ISSUE #2121: FAWE does not write into extent of edit session, but into world of extent
### Server Implementation

Paper

### Server Version

1.19.3

### Describe the bug

First: This bug does not occur when using WorldEdit. It only occurs when using FAWE

When creating an ImmutableWorld Extent which blocks setting of a block and modifying the world in general, FAWE still writes into the world itself instead of the extend provided for the edit session.

None of the setBlock methods of the extent is called. I suspect that FAWE directly writes into the chunks which would makes sense considering how fawe works, but this causes that the extent is used to retrieve the chunks but not to write the changes.

Since the original extent is wrapped into an SingleThreadQueueExtent and this implements how the chunk setters are retrieved, there is no real way to prevent writing into the chunks from api side.

**Case 1**
EditSession with normal world.
PasteOperation into EditSession as Extent
**Result:** Schematic gets pasted into the world (Expected)

**Case 2**
EditSession with ImmutableWorld.
PasteOperation with ImmutableWorld as Extent
**Result:** Schematic does NOT get paste into the world (Expected)

**Case 3**
EditSession with ImmutableWorld.
PasteOperation with EditSession as Extent
**Result:** Schematic gets pasted into the world (Unexpected)

### To Reproduce

I created a minimal working example:

https://github.com/eldoriarpg/we-playground

1. Install the plugin
2. Execute `/wetest <case>` to execute one of the cases described above.

### Expected behaviour

That FAWE writes into the provided extent and not into the chunks of the extent as it seems.

### Screenshots / Videos

_No response_

### Error log (if applicable)

_No response_

### Fawe Debugpaste

https://athion.net/ISPaster/paste/view/973182bf68674dfc91006cc50d54d6c4

### Fawe Version

FastAsyncWorldEdit-Bukkit-2.5.3-SNAPSHOT-370

### Checklist

### Anything else?

_No response_

TheMeinerLP:
## Overview
Fixes #2121 

## Description
In both cases when `FAST_PLACEMENT` is true or false in the config. The boolean will never be false. Of this and fastmode is the cause while every time place chunks in [EditSessionBuilder.compile()](https://github.com/IntellectualSites/FastAsyncWorldEdit/blob/da4d966d9e54873ad4b255dce4729fce0739c002/worldedit-core/src/main/java/com/sk89q/worldedit/EditSessionBuilder.java#L464) will be triggered.

## Submitter Checklist

### DIFF
--- a/worldedit-core/src/main/java/com/fastasyncworldedit/core/configuration/Settings.java
+++ b/worldedit-core/src/main/java/com/fastasyncworldedit/core/configuration/Settings.java
@@ -152,7 +152,7 @@ public class Settings extends Config {
                 );
                 limit.INVENTORY_MODE = Math.min(limit.INVENTORY_MODE, newLimit.INVENTORY_MODE);
                 limit.SPEED_REDUCTION = Math.min(limit.SPEED_REDUCTION, newLimit.SPEED_REDUCTION);
-                limit.FAST_PLACEMENT |= newLimit.FAST_PLACEMENT;
+                limit.FAST_PLACEMENT &= newLimit.FAST_PLACEMENT;
                 limit.CONFIRM_LARGE &= newLimit.CONFIRM_LARGE;
                 limit.RESTRICT_HISTORY_TO_REGIONS &= newLimit.RESTRICT_HISTORY_TO_REGIONS;
                 if (limit.STRIP_NBT == null) {
--- a/worldedit-core/src/main/java/com/fastasyncworldedit/core/configuration/Settings.java
+++ b/worldedit-core/src/main/java/com/fastasyncworldedit/core/configuration/Settings.java
@@ -152,7 +152,7 @@ public class Settings extends Config {
                 );
                 limit.INVENTORY_MODE = Math.min(limit.INVENTORY_MODE, newLimit.INVENTORY_MODE);
                 limit.SPEED_REDUCTION = Math.min(limit.SPEED_REDUCTION, newLimit.SPEED_REDUCTION);
-                limit.FAST_PLACEMENT &= newLimit.FAST_PLACEMENT;
+                limit.FAST_PLACEMENT |= newLimit.FAST_PLACEMENT;
                 limit.CONFIRM_LARGE &= newLimit.CONFIRM_LARGE;
                 limit.RESTRICT_HISTORY_TO_REGIONS &= newLimit.RESTRICT_HISTORY_TO_REGIONS;
                 if (limit.STRIP_NBT == null) {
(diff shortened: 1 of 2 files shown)


TheMeinerLP:
> I don't think this is the correct fix for this issue. The FAST_PLACEMENT limit should be enabled if _any_ limit applying to the player has FAST_PLACEMENT enabled, not if _all_ have it enabled.
> 
> However, the fastMode option should probably override this limit. But from a quick look, this might affect larger parts of functionality, so the behavior must be analyzed deeply to understand issues that might arise from this.

I discussed with @rainbowdashlabs and checked what coded references to this fast placement. 
This boolean is only for handle setting of if chunk placing enabled or not. 

On the end. The setting boolean is in current version of fawe not working.

Anytime i checked the boolean on that method that i referenced. It was false.
*I used the debugger*

TheMeinerLP:
After @SirYwell and me discussed privately. I will change the PR to if the fastMode is false. Then FAST_PLACEMENT will be ignored.

dordsor21:
This is entirely the wrong fix for the issue. The issue is that fast placement often bypasses the setBlock methods specifically and simply uses the processSet, fillBlocks, etc. in order to fill chunks with a pattern when the edit is "basic". There should be API notes added to this nature. See https://github.com/IntellectualSites/FastAsyncWorldEdit/blob/main/worldedit-core/src/main/java/com/fastasyncworldedit/core/queue/implementation/ParallelQueueExtent.java

Tbh, I'd be inclined to remove the fast placement option from the limit altogether and put it into queue. It makes no sense in limit as it's effectively a switch for going from block-by-block placement to chunk-by-chunk (efficient) placement.

TheMeinerLP:
> This is entirely the wrong fix for the issue. The issue is that fast placement often bypasses the setBlock methods specifically and simply uses the processSet, fillBlocks, etc. in order to fill chunks with a pattern when the edit is "basic". There should be API notes added to this nature. See https://github.com/IntellectualSites/FastAsyncWorldEdit/blob/main/worldedit-core/src/main/java/com/fastasyncworldedit/core/queue/implementation/ParallelQueueExtent.java
> 
> Tbh, I'd be inclined to remove the fast placement option from the limit altogether and put it into queue. It makes no sense in limit as it's effectively a switch for going from block-by-block placement to chunk-by-chunk (efficient) placement.

How else can i manage to use the right extent ?

You think i must be reworked the ParallelQueueExtent to allow "custom" ones ?

dordsor21:
No, there needs to be API notes added somewhere telling people that processSet should be overridden alongside the normal setBlock methods. The issue is effectively still that people are not overriding (all) methods

dordsor21:
Also, Worlds should basically never be used in place of an actual extent, which is the root issue of #2121 looking into the issue further

TheMeinerLP:
How we can go here future ?

PR #2181, opened 2023-04-17
Apply DataArray patch and update Bukkit adapters.

### LINKED ISSUE #1938: Change FAWE to use an int[] or long[] based ordinal system instead of a char[] ordinal system in order to support more than 65535 different blockstates

Hello, I am currently working on a Forge adapter update for FastAsyncWorldEdit, and during my work I have come across an issue that prevents me from submitting a pull request for it at this time. FAWE currently uses a char[] based ordinal system when it performs its operations. Due to this, the system is currently limited to supporting a maximum of 65535 blockstates since that is the maximum value of the char primitive data type. As such, I would like to suggest converting the system to be based off of either integer arrays or long arrays. Doing so would allow support for significantly more blockstates added by mods. I have implemented the change to an int[] based ordinal system on my own local version and have my Forge adapter update in a working and nearly complete state. 

I would be happy to submit this change myself as a PR if it would be desired.

Currently there are no alternatives for supporting more that 65535 blockstates in FAWE's ordinal system.

### Anything else?

_No response_

JayemCeekay:
## Overview

Fixes #1938

## Description

This pull request implements the patch for the use of type ambiguous DataArrays created by @SirYwell. This is necessary for the upcoming implementation of updated Forge and Fabric adapters in order to allow the registration and use of more than 65535 blockstates.

## Submitter Checklist

### DIFF
--- a/worldedit-bukkit/adapters/adapter-1_20_2/src/main/java/com/sk89q/worldedit/bukkit/adapter/impl/fawe/v1_20_R2/PaperweightGetBlocks.java
+++ b/worldedit-bukkit/adapters/adapter-1_20_2/src/main/java/com/sk89q/worldedit/bukkit/adapter/impl/fawe/v1_20_R2/PaperweightGetBlocks.java
@@ -11,6 +11,7 @@ import com.fastasyncworldedit.core.math.BitArrayUnstretched;
 import com.fastasyncworldedit.core.math.IntPair;
 import com.fastasyncworldedit.core.nbt.FaweCompoundTag;
 import com.fastasyncworldedit.core.queue.IChunkSet;
+import com.fastasyncworldedit.core.queue.implementation.blocks.DataArray;
 import com.fastasyncworldedit.core.util.MathMan;
 import com.fastasyncworldedit.core.util.NbtUtils;
 import com.fastasyncworldedit.core.util.collection.AdaptedMap;
@@ -405,7 +406,7 @@ public class PaperweightGetBlocks extends AbstractBukkitGetBlocks<ServerLevel, L
                                 );
                                 LevelChunkSection newSection = PaperweightPlatformAdapter.newChunkSection(
                                         layerNo,
-                                        new char[4096],
+                                        DataArray.createEmpty(),
                                         adapter,
                                         biomeRegistry,
                                         biomeData
@@ -418,7 +419,7 @@ public class PaperweightGetBlocks extends AbstractBukkitGetBlocks<ServerLevel, L
                                         newSection,
                                         getSectionIndex
                                 )) {
-                                    updateGet(nmsChunk, levelChunkSections, newSection, new char[4096], getSectionIndex);
+                                    updateGet(nmsChunk, levelChunkSections, newSection, DataArray.createEmpty(), getSectionIndex);
                                     continue;
                                 } else {
                                     existingSection = levelChunkSections[getSectionIndex];
@@ -451,9 +452,7 @@ public class PaperweightGetBlocks extends AbstractBukkitGetBlocks<ServerLevel, L
 
                 // setArr is modified by PaperweightPlatformAdapter#newChunkSection. This is in order to write changes to
                 // this chunk GET when #updateGet is called. Future dords, please listen this time.
-                char[] tmp = set.load(layerNo);
-                char[] setArr = new char[tmp.length];
-                System.arraycopy(tmp, 0, setArr, 0, tmp.length);
+                DataArray setArr = DataArray.createCopy(set.load(layerNo));
 
                 // synchronise on internal section to avoid circular locking with a continuing edit if the chunk was
                 // submitted to keep loaded internal chunks to queue target size.
@@ -469,15 +468,12 @@ public class PaperweightGetBlocks extends AbstractBukkitGetBlocks<ServerLevel, L
                         }
                     }
 
-                    if (createCopy) {
-                        char[] tmpLoad = load(layerNo);
-                        char[] copyArr = new char[4096];
-                        System.arraycopy(tmpLoad, 0, copyArr, 0, 4096);
-                        copy.storeSection(getSectionIndex, copyArr);
-                        if (biomes != null && existingSection != null) {
-                            copy.storeBiomes(getSectionIndex, existingSection.getBiomes());
+                        if (createCopy) {
+                            copy.storeSection(getSectionIndex, DataArray.createCopy(load(layerNo)));
+                            if (biomes != null && existingSection != null) {
+                                copy.storeBiomes(getSectionIndex, existingSection.getBiomes());
+                            }
                         }
-                    }
 
                     if (existingSection == null) {
                         PalettedContainer<Holder<Biome>> biomeData = biomes == null ? new PalettedContainer<>(
@@ -535,10 +531,9 @@ public class PaperweightGetBlocks extends AbstractBukkitGetBlocks<ServerLevel, L
                             } else if (existingSection != getSections(false)[getSectionIndex]) {
                                 this.sections[getSectionIndex] = existingSection;
                                 this.reset();
-                            } else if (!Arrays.equals(update(getSectionIndex, new char[4096], true), load(layerNo))) {
+                            } else if (!update(getSectionIndex, DataArray.createEmpty(), true)
+                                    .equals(load(layerNo))) {
                                 this.reset(layerNo);
-                        /*} else if (lock.isModified()) {
-                            this.reset(layerNo);*/
                             }
                         } finally {
                             sectionLock.writeLock().unlock();
@@ -768,7 +763,7 @@ public class PaperweightGetBlocks extends AbstractBukkitGetBlocks<ServerLevel, L
             LevelChunk nmsChunk,
             LevelChunkSection[] chunkSections,
             LevelChunkSection section,
-            char[] arr,
+            DataArray arr,
             int layer
     ) {
         try {
@@ -812,16 +807,14 @@ public class PaperweightGetBlocks extends AbstractBukkitGetBlocks<ServerLevel, L
      */
     @Override
     @SuppressWarnings("unchecked")
-    public char[] update(int layer, char[] data, boolean aggressive) {
+    public DataArray update(int layer, DataArray data, boolean aggressive) {
         LevelChunkSection section = getSections(aggressive)[layer];
         // Section is null, return empty array
         if (section == null) {
-            data = new char[4096];
-            Arrays.fill(data, (char) BlockTypesCache.ReservedIDs.AIR);
-            return data;
+            return DataArray.createFilled(BlockTypesCache.ReservedIDs.AIR);
         }
-        if (data == null || data == FaweCache.INSTANCE.EMPTY_CHAR_4096 || data.length != 4096) {
-            data = new char[4096]; // new array, will be populated below
+        if (data == null || data == FaweCache.INSTANCE.EMPTY_DATA) {
+            data = DataArray.createFilled(BlockTypesCache.ReservedIDs.AIR);
         }
         Semaphore lock = PaperweightPlatformAdapter.applyLock(section);
         synchronized (lock) {
@@ -834,7 +827,7 @@ public class PaperweightGetBlocks extends AbstractBukkitGetBlocks<ServerLevel, L
                 final BitStorage bits = (BitStorage) PaperweightPlatformAdapter.fieldStorage.get(dataObject);
 
                 if (bits instanceof ZeroBitStorage) {
-                    Arrays.fill(data, adapter.adaptToChar(blocks.get(0, 0, 0))); // get(int) is only public on paper
+                    data.setAll(adapter.adaptToOrdinal(blocks.get(0, 0, 0))); // get(int) is only public on paper
                     return data;
                 }
 
@@ -854,14 +847,14 @@ public class PaperweightGetBlocks extends AbstractBukkitGetBlocks<ServerLevel, L
                     return data;
                 }
 
-                char[] paletteToOrdinal = FaweCache.INSTANCE.PALETTE_TO_BLOCK_CHAR.get();
+                int[] paletteToOrdinal = FaweCache.INSTANCE.PALETTE_TO_BLOCK.get();
                 try {
                     if (num_palette == 1) {
-                        char ordinal = ordinal(palette.valueFor(0), adapter);
-                        Arrays.fill(data, ordinal);
+                        int ordinal = ordinal(palette.valueFor(0), adapter);
+                        data.setAll(ordinal);
                     } else {
                         for (int i = 0; i < num_palette; i++) {
-                            char ordinal = ordinal(palette.valueFor(i), adapter);
+                            int ordinal = ordinal(palette.valueFor(i), adapter);
                             paletteToOrdinal[i] = ordinal;
                         }
                         adapter.mapWithPalette(data, paletteToOrdinal);
@@ -879,11 +872,11 @@ public class PaperweightGetBlocks extends AbstractBukkitGetBlocks<ServerLevel, L
         }
     }
 
-    private char ordinal(BlockState ibd, PaperweightFaweAdapter adapter) {
+    private int ordinal(BlockState ibd, PaperweightFaweAdapter adapter) {
         if (ibd == null) {
             return BlockTypesCache.ReservedIDs.AIR;
         } else {
-            return adapter.adaptToChar(ibd);
+            return adapter.adaptToOrdinal(ibd);
         }
     }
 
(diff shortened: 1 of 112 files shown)


SirYwell:
I updated this branch to reflect the latest changes on main

SirYwell:
Retargeted to v3 and rebased, though I need to re-review the changes myself because there were some ugly merge conflicts that probably broke something.

worldedit-bukkit/adapters/adapter-1_17_1/src/main/java/com/sk89q/worldedit/bukkit/adapter/impl/fawe/v1_17_R1_2/PaperweightGetBlocks.java
```diff
@@ -894,9 +881,9 @@ public char[] update(int layer, char[] data, boolean aggressive) {
                 } else {
                     // The section's palette is the global block palette.
                     for (int i = 0; i < 4096; i++) {
-                        char paletteVal = data[i];
+                        char paletteVal = (char) data.getAt(i);
```
dordsor21, line 884: I think this (and the equivalent in all other adapters) should use int rather than char, else, the method breaks down when using an IntDataArray. Methods that return char should also be changed to int in adapters (e.g. `adaptToChar`, `adaptToChar` etc. The cached char[] array for `ibdToStateOrdinal` can also just be int[] - the memory impact is pretty negligible (in the 100s of KB)

worldedit-bukkit/adapters/adapter-1_20_2/src/main/java/com/sk89q/worldedit/bukkit/adapter/impl/fawe/v1_20_R2/PaperweightPostProcessor.java
```diff
@@ -115,39 +116,39 @@ public ProcessorScope getScope() {
         return ProcessorScope.READING_SET_BLOCKS;
     }
 
-    private boolean wasAdjacentToWater(char[] get, char[] set, int i, int x, int y, int z) {
+    private boolean wasAdjacentToWater(DataArray get, DataArray set, int i, int x, int y, int z) {
```
dordsor21, line 119: this can also be moved to the method in the super `PostProcessor` class?

worldedit-bukkit/adapters/adapter-1_20/src/main/java/com/sk89q/worldedit/bukkit/adapter/impl/fawe/v1_20_R1/PaperweightGetBlocks_Copy.java
```diff
@@ -236,7 +237,7 @@ public <T extends Future<T>> T call(IChunkSet set, Runnable finalize) {
     public char get(int x, int y, int z) {
         final int layer = (y >> 4) - getMinSectionPosition();
         final int index = (y & 15) << 8 | z << 4 | x;
-        return blocks[layer][index];
+        return (char) blocks[layer].getAt(index);
```
dordsor21, line 240: likewise move to int for this method in all adapters too

worldedit-bukkit/src/main/java/com/fastasyncworldedit/bukkit/adapter/NMSAdapter.java
```diff
@@ -17,14 +18,14 @@ public static int createPalette(
...
-            char[] set,
+            DataArray set,
             CachedBukkitAdapter adapter,
             short[] nonEmptyBlockCount
     ) {
         short nonAir = 4096;
         int num_palette = 0;
         for (int i = 0; i < 4096; i++) {
-            char ordinal = set[i];
+            char ordinal = (char) set.getAt(i);
```
dordsor21, line 28: This should all be moved to int as well

worldedit-bukkit/src/main/java/com/fastasyncworldedit/bukkit/adapter/PostProcessor.java
```diff
@@ -0,0 +1,78 @@
...
+    @Override
+    public Extent construct(final Extent child) {
+        throw new UnsupportedOperationException("Processing only");
+    }
+
+    protected boolean wasAdjacentToWater(DataArray get, DataArray set, int i, int x, int y, int z) {
+        if (set == null || get == null) {
+            return false;
+        }
+        char ordinal;
```
dordsor21, line 30: use int

worldedit-core/src/main/java/com/fastasyncworldedit/core/FaweCache.java
```diff
@@ -256,8 +269,12 @@ public V apply(final long input) {
     public final CleanableThreadLocal<char[]> PALETTE_TO_BLOCK_CHAR = new CleanableThreadLocal<>(
             () -> new char[Character.MAX_VALUE + 1], a -> {
         Arrays.fill(a, Character.MAX_VALUE);
-    }
-    );
+    });
+
+    public final CleanableThreadLocal<int[]> PALETTE_TO_BLOCK_INT = new CleanableThreadLocal<>(
+            () -> new int[Character.MAX_VALUE + 1], a -> {
```
dordsor21, line 275: This should probably be initialised to block types cache size

worldedit-core/src/main/java/com/fastasyncworldedit/core/FaweCache.java
```diff
@@ -290,21 +307,10 @@ public static final class Palette {
...
      */
-    public Palette toPalette(int layerOffset, char[] blocks) {
-        return toPalette(layerOffset, null, blocks);
-    }
-
-    /**
-     * Convert raw int array to palette
-     * @return palette
-     */
-    public Palette toPalette(int layerOffset, int[] blocks) {
```
dordsor21, line 303: These methods being removed should be deprecated for removal and overloaded

worldedit-core/src/main/java/com/fastasyncworldedit/core/queue/implementation/blocks/ChunkSectionedChunk.java
```diff
@@ -0,0 +1,45 @@
...
+import com.sk89q.worldedit.world.biome.BiomeType;
+
+public class ChunkSectionedChunk {
+
+    public Object[] sectionLocks;
+    protected int minSectionPosition;
+    protected int maxSectionPosition;
+    protected int sectionCount;
+
+    static BiomeType getBiomeType(
```
dordsor21, line 12: I'm not a particular fan of having a random biome method here, I'm not really sure that this class is needed and I think it ends up adding more clutter than it aims to remove

worldedit-core/src/main/java/com/fastasyncworldedit/core/queue/implementation/blocks/ThreadUnsafeCharBlocks.java
```diff
@@ -26,8 +25,8 @@
 import java.util.UUID;
 
 /**
- * Equivalent to {@link CharSetBlocks} without any attempt to make thread-safe for improved performance.
- * This is currently only used as a "copy" of {@link CharSetBlocks} to provide to
+ * Equivalent to {@link DataArraySetBlocks} without any attempt to make thread-safe for improved performance.
+ * This is currently only used as a "copy" of {@link DataArraySetBlocks} to provide to
  * {@link com.fastasyncworldedit.core.queue.IBatchProcessor} instances for processing without overlapping the continuing edit.
  *
  * @since 2.6.2
```
dordsor21, line 32: I suppose this should be renamed to ThreadUnsaveDataArrayBlocks

worldedit-core/src/main/java/com/sk89q/worldedit/regions/Region.java
```diff
@@ -485,6 +477,23 @@ default IChunkSet processSet(IChunk chunk, IChunkGet get, IChunkSet set, boolean
         }
     }
 
+    private boolean isProcessExtra(IChunkSet set, boolean processExtra, int layer, DataArray arr) {
```
dordsor21, line 480: shouldProcessExtra?

dordsor21 (changes requested):
I kinda wonder if we should just move entirely to int based arrays. The amount of extra method calls for setting blocks required it quite substantial and will have a performance impact. I don't really see the memory impact being particularly large.

Also, clipboards will need to be investigated as they are very char based (and moving to an int-based system will halve the capacity of disk-based solutions)

SirYwell:
The casts to `char` should only be present in the bukkit module, but I'm fine with changing it to int there too.

> I kinda wonder if we should just move entirely to int based arrays. The amount of extra method calls for setting blocks required it quite substantial and will have a performance impact. I don't really see the memory impact being particularly large.

I don't expect the methods to have measurable overhead over direct array access (we have one additional memory indirection until we have value classes, the methods are small, the JVM can inline them easily). Memory overhead of always using `int[]` might be fine, but that also mean more cache misses. It also might be interesting in future to explore Foreign Memory based approaches, in which case the `DataArray` abstraction would be useful too. So I think it's not worth to *not* have the `DataArray` abstraction.


> Also, clipboards will need to be investigated as they are very char based (and moving to an int-based system will halve the capacity of disk-based solutions)

I started investigating that in https://github.com/IntellectualSites/FastAsyncWorldEdit/tree/feature/disk-based-clipboard, but there are more things to consider.

dordsor21:
> The casts to `char` should only be present in the bukkit module, but I'm fine with changing it to int there too.

I'm not sure that's something we can 100% assume with datapacks, etc?

> I don't expect the methods to have measurable overhead over direct array access (we have one additional memory indirection until we have value classes, the methods are small, the JVM can inline them easily). Memory overhead of always using `int[]` might be fine, but that also mean more cache misses. It also might be interesting in future to explore Foreign Memory based approaches, in which case the `DataArray` abstraction would be useful too. So I think it's not worth to _not_ have the `DataArray` abstraction.

I'm not sure if an increased cache miss likelihood is really a large performance hit? I already assume that large edits that cannot be done per-chunk (large clipboard operations, any recursive operations) are already missing L1 (and very possibly L2 and L3 cache too) as the number of chunks being loaded and re-loaded is quite large. And then if a whole chunk's block arrays (max 200 KiB with char, 400 with int) are missing L3 cache we should be attempting to do something about that. Besides, I doubt that most servers have sole access to a single CPU to be able to be making full using the L1-3 cache anyway, and I would expect lots of cache misses due to running websites, other servers, services etc anyway. Switching to all int[] is more maintainable too

> I started investigating that in [`feature`/disk-based-clipboard](https://github.com/IntellectualSites/FastAsyncWorldEdit/tree/feature/disk-based-clipboard?rgh-link-date=2023-11-19T15%3A11%3A00Z), but there are more things to consider.

AH yeah I'd forgotten about that, looks to make it work

SirYwell:
> I'm not sure that's something we can 100% assume with datapacks, etc?

AFAIK it's still not possible to have additional block types/states. Custom biomes are possible though.

> I'm not sure if an increased cache miss likelihood is really a large performance hit

Most likely not, my point was more that we'd need to measure to really know the impact here (and even then, there are many parts that probably aren't true anymore in the next Java release).
I'd really like to keep the `DataArray` abstraction, there are a few things that are far easier to understand with that approach (e.g. avoiding System.arraycopy in the middle of already complex code, filling arrays, etc). It generally is more future-proof and will potentially allow for a bunch of optimizations in future.

SirYwell:
I made this PR compatible with the Vector API changes now - somewhat hacky, but seems to work.

worldedit-core/src/main/java/com/sk89q/worldedit/regions/Region.java
dordsor21: Should probably actually call the method sometihng that implies it's doing something to the data at all

dordsor21 (commented):
Most of previous comments still apply

SirYwell:
I significantly reworked this PR now.

1. It is based on Java 25 now.
2. It makes use of `MemorySegment`.
3. It currently implements only 4 byte per block storage, but additional configurations can be added easily.
4. The memory segment is currently backed by a `long` array. This can be changed too later on.

Why not switch to `int[]`? Having control over the length already makes quite a few things easier and more concise.

Why not use `MemorySegment` directly? Same as for `int[]`, but also access to memory segments is unified this way.

There are also potential future improvements and features that would depend on this:
- Native memory: we could use native memory, however this only makes sense if we can properly track the lifetime of the memory segments. 
- Making the lifetimes more explicit would likely help cleaning up code.
- Precise memory usage tracking: we could limit the memory an edit is allowed to use, this would depend on using native memory to make sense.
- Besides 2 bytes per block configuration, we could experiment with 3 bytes per block.
- We could experiment with copy-on-write constructs.
- We could experiment with more read-only access.

I didn't do any proper benchmarks yet, but I expect this to be fine. Having value classes would be better, but having only one implementation should already make devirtualization and inlining pretty simple for the JIT.

worldedit-bukkit/adapters/adapter-1_20_2/src/main/java/com/sk89q/worldedit/bukkit/adapter/impl/fawe/v1_20_R2/PaperweightGetBlocks.java
```diff
@@ -469,15 +470,12 @@ protected <T extends Future<T>> T internalCall(
...
-                        System.arraycopy(tmpLoad, 0, copyArr, 0, 4096);
-                        copy.storeSection(getSectionIndex, copyArr);
-                        if (biomes != null && existingSection != null) {
-                            copy.storeBiomes(getSectionIndex, existingSection.getBiomes());
+                        if (createCopy) {
+                            copy.storeSection(getSectionIndex, DataArray.createCopy(load(layerNo)));
+                            if (biomes != null && existingSection != null) {
+                                copy.storeBiomes(getSectionIndex, existingSection.getBiomes());
+                            }
                         }
```
dordsor21: Formatting

worldedit-bukkit/adapters/adapter-1_20_2/src/main/java/com/sk89q/worldedit/bukkit/adapter/impl/fawe/v1_20_R2/PaperweightGetBlocks_Copy.java
```diff
@@ -283,7 +282,7 @@ public <T extends Future<T>> T call(IQueueExtent<? extends IChunk> owner, IChunk
     public char get(int x, int y, int z) {
         final int layer = (y >> 4) - getMinSectionPosition();
         final int index = (y & 15) << 8 | z << 4 | x;
-        return blocks[layer][index];
+        return (char) blocks[layer].getAt(index);
```
dordsor21: Should be int? (and other adapters)

worldedit-bukkit/src/main/java/com/fastasyncworldedit/bukkit/adapter/PostProcessor.java
```diff
@@ -0,0 +1,78 @@
...
+import com.sk89q.worldedit.extent.Extent;
+import com.sk89q.worldedit.world.block.BlockState;
+import com.sk89q.worldedit.world.block.BlockTypesCache;
+
+import javax.annotation.Nullable;
+
+/**
+ * Common code for post-processing on different versions
+ */
+public abstract class PostProcessor implements IBatchProcessor {
```
dordsor21: Is this necessarily bukkit-only?

worldedit-core/src/main/java/com/fastasyncworldedit/core/queue/implementation/blocks/ThreadUnsafeCharBlocks.java
```diff
@@ -184,18 +183,12 @@ public char get(int x, int y, int z) {
             return defaultOrdinal;
         }
         final int index = (y & 15) << 8 | z << 4 | x;
-        return blocks[layer - minSectionPosition][index];
+        return (char) blocks[layer - minSectionPosition].getAt(index);
```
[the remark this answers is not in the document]
dordsor21: ^

worldedit-core/src/main/java/com/fastasyncworldedit/core/queue/implementation/blocks/ThreadUnsafeDataArrayBlocks.java
```diff
@@ -465,11 +458,12 @@ public boolean hasBiomes(int layer) {
...
     public IChunkSet createCopy() {
-        char[][] blocksCopy = new char[sectionCount][];
+        DataArray[] blocksCopy = new DataArray[sectionCount];
         for (int i = 0; i < sectionCount; i++) {
-            blocksCopy[i] = new char[FaweCache.INSTANCE.BLOCKS_PER_LAYER];
             if (blocks[i] != null) {
-                System.arraycopy(blocks[i], 0, blocksCopy[i], 0, FaweCache.INSTANCE.BLOCKS_PER_LAYER);
+                blocksCopy[i] = DataArray.createCopy(blocks[i]);
+            } else {
+                blocksCopy[i] = DataArray.createEmpty();
```
dordsor21: todo optimise to null? We can probably support null in the copy

SirYwell:
Tried to get rid of `char` usages where relevant, but I might have missed something. I'll address the other stuff separately.

PR #2194, opened 2023-04-28
Remove weak reference and replace with a simple not equal

### LINKED ISSUE #2188: Invalid region mask: PlotSquaredMask (unnecessary/slightly spammy message)
### Server Implementation

Paper

### Server Version

1.19.4

### Describe the bug

After updating from FAWE 2.6.0 to 2.6.1 players on my server complained about an error appearing in chat every time they use a worldedit command. The operation still works but it gives an error in chat, also the first operation does not give this error but any operation after that will give the error.

First operation gives no error:

Every operation after the first gives this error in chat "Invalid region mask: PlotSquaredMask":

### To Reproduce

1. Make sure you are not OP
2. Select a region with the wand and do `//set 0`
3. Do `//set stone` on the same region and you will get an error in chat

### Expected behaviour

There should be no error in chat

### Screenshots / Videos

_No response_

### Error log (if applicable)

_No response_

### Fawe Debugpaste

https://athion.net/ISPaster/paste/view/487f28fdfaa74188acd4ba24a4a22ecd

### Fawe Version

FastAsyncWorldEdit version 2.6.2-SNAPSHOT-408;9f3c909

### Checklist

### Anything else?

_No response_

TheMeinerLP:
## Overview
Fixes #2188 

## Description
Since 2.6.1 added some issue for plot squared mask with unlinked plot mask. Now without weak reference the code work again as expected.
Side information. I tested it locally

## Submitter Checklist

### DIFF
--- a/worldedit-bukkit/src/main/java/com/fastasyncworldedit/bukkit/regions/plotsquared/PlotSquaredFeature.java
+++ b/worldedit-bukkit/src/main/java/com/fastasyncworldedit/bukkit/regions/plotsquared/PlotSquaredFeature.java
@@ -23,7 +23,6 @@ import com.sk89q.worldedit.world.World;
 import org.apache.logging.log4j.Logger;
 import org.bukkit.Bukkit;
 
-import java.lang.ref.WeakReference;
 import java.util.List;
 import java.util.Locale;
 import java.util.Set;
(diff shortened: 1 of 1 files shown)


PierreSchwang:
Have you tested, that the change does not break the recently implemented behavior? https://github.com/IntellectualSites/FastAsyncWorldEdit/commit/e4158ac08401c293f98fe5ce85b7a64a1d299aef

TheMeinerLP:
I have tested with P2

SirYwell:
I don't see how this changes the actual behavior other than potentionally creating a memory leak. How is the weak ref the issue here?

TheMeinerLP:
Before everytime there was a message on Fawe for the wea mask from p2 this prevents now the exception and fixes the linked issue. A weak reference is here not needed

SirYwell:
The weak ref is used to avoid memory leaks, and I don't see the semantic difference here. That's why I'd like to know *how* this change fixes the issue.

TheMeinerLP:
The hashset what we get is a strong one. Also in there original get logic from connected plots is a cache implemented. When we now use a weak reference the content gets never updated or to late.

SirYwell:
The behavior of the contents doesn't change here, we call the getConnectedPlots method at the exact same places. The only difference is that if the cache is invalidated by PlotSquared, this solution still holds a strong reference to the *outdated* set that would otherwise be garbage collected with the existing solution.

dordsor21:
It's actually a bug that the message _wouldn't_ show (i.e. the message showing is correct - the weak reference has done its job). That implies that the mask has been invalidated by FAWE, and thus recalculated. Perhaps the message isn't needed and would server better as a debug-related message.

TheMeinerLP:
Okay for that i will replace it with your original code and replace the message with simple `logger.debug(FILLER)` ?
Thats right

dordsor21:
> Okay for that i will replace it with your original code and replace the message with simple `logger.debug(FILLER)` ? Thats right

Not sure if we do it elsewhere, but the message would be more suited to sending to the player still

TheMeinerLP:
I prefer to do that better in the future. Then my suggestion. Let switch to debug logging for like that

dordsor21:
Closing in favour of #2238 and IntellectualSites/PlotSquared#4046

PR #2198, opened 2023-04-30
Add mangrove tree type

### LINKED ISSUE #2197: //brush mangrove

Currently, the tree brush does not support mangrove trees. 

I would like this to support mangrove trees.

Not that I know of, I checked other issues I couldn't find any with the content "mangrove". 

### Anything else?

_No response_

TheMeinerLP:
## Overview
Fixes #2197 

## Description
Add some debug messages for exceptions. Also add mangrove as a tree type

## Submitter Checklist

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/util/TreeGenerator.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/util/TreeGenerator.java
@@ -49,6 +49,7 @@ public final class TreeGenerator {
         REDWOOD("Spruce tree", "spruce", "redwood", "sequoia", "sequoioideae"),
         TALL_REDWOOD("Tall spruce tree", "tallspruce", "bigspruce", "tallredwood", "tallsequoia", "tallsequoioideae"),
         MEGA_REDWOOD("Large spruce tree", "largespruce", "megaredwood"),
+        MANGROVE("Mangrove", "mangrove"),
         RANDOM_REDWOOD("Random spruce tree", "randspruce", "randredwood", "randomredwood", "anyredwood") {
             @Override
             public boolean generate(EditSession editSession, BlockVector3 pos) throws MaxChangedBlocksException {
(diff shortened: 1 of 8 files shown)


NotMyFault:
Tests are failing.

SirYwell (changes requested):
Catching unchecked exceptions should be a last resort workaround if we can't control the code. In this case, however, it looks like we can properly deal with it instead of catching exceptions?

TheMeinerLP:
> Catching unchecked exceptions should be a last resort workaround if we can't control the code. In this case, however, it looks like we can properly deal with it instead of catching exceptions?

There is currently no other way for handle thats type. 
I tried to check it before we register that tree type. But everytime it fails.
Now this solution is the friendliest one for every user of fawe.

SirYwell:
The NPE catches can clearly be replaced with null checks earlier. Registering the mapping can be guarded by a version check.

TheMeinerLP:
> The NPE catches can clearly be replaced with null checks earlier. Registering the mapping can be guarded by a version check.

The NPE comes from Bukkit it self. We can't handle it simple null checks

SirYwell:
I assume the NPE comes from `TreeType bukkitType = BukkitWorld.toBukkitTreeType(treeType);` being null? If so, we can simply return directly there

TheMeinerLP:
> I assume the NPE comes from `TreeType bukkitType = BukkitWorld.toBukkitTreeType(treeType);` being null? If so, we can simply return directly there

The NPE comes from internal of Bukkit not from our code. `bukkitWorld.generateTree(<PLACEHOLDER>, type)` throws a exception when we give over the tree type

SirYwell:
Which type? It will be null if the type doesn't exist on that version -> NPE

TheMeinerLP:
> Which type? It will be null if the type doesn't exist on that version -> NPE

I will try to handle it with null check and add a comment why.
Also i will add the debug again

NotMyFault (commented):
This should really be in upstream.

TheMeinerLP:
I can do that 😄

worldedit-core/src/main/java/com/sk89q/worldedit/util/TreeGenerator.java
```diff
@@ -49,6 +49,7 @@ public enum TreeType {
         REDWOOD("Spruce tree", "spruce", "redwood", "sequoia", "sequoioideae"),
         TALL_REDWOOD("Tall spruce tree", "tallspruce", "bigspruce", "tallredwood", "tallsequoia", "tallsequoioideae"),
         MEGA_REDWOOD("Large spruce tree", "largespruce", "megaredwood"),
+        MANGROVE("Mangrove", "mangrove"),
```
MattBDev, line 52:
```suggestion
        MANGROVE("Mangrove tree", "mangrove"),
```

MattBDev (changes requested):
I am requesting some changes since this will be a commit will have to PR to EngineHub/WorldEdit. 
In addition to the recommendations I commented on, I would also like to see Tall Mangrove tree's added to this PR before I approve.

NotMyFault:
Upstreamed by dordsor, due to lack of action by the PR creator.

PR #2210, opened 2023-05-05
Added upstream OffsetsMask classes

MattBDev:
## Overview
Fixes one of the API incompatibilities we have with upstream by adding these new masks to FAWE.
 
## Description
Adds two new masks to FAWE from upstream and ports a few FAWE changes into them.

## Submitter Checklist
- Not Applicable: New public fields and methods are annotated with `@since TODO`.

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/function/mask/OffsetMask.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/function/mask/OffsetMask.java
@@ -30,7 +30,10 @@ import static com.google.common.base.Preconditions.checkNotNull;
 /**
  * Checks whether another mask tests true for a position that is offset
  * a given vector.
+ *
+ * @deprecated Use {@link OffsetsMask#single}
  */
+@Deprecated
 public class OffsetMask extends AbstractMask {
 
     //FAWE start - ignore resultant position outside world height range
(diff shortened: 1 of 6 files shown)


MattBDev:
The checks seem to have failed but the code does compile and works. Not sure what is the cause of the failure to resolve dependencies and the CodeQL issue. I don’t even touch dependencies in this PR.

worldedit-core/src/main/java/com/sk89q/worldedit/function/mask/OffsetMask.java
```diff
@@ -30,7 +30,10 @@
 /**
  * Checks whether another mask tests true for a position that is offset
  * a given vector.
+ *
+ * @deprecated Use {@link OffsetsMask#single}
  */
+@Deprecated
```
dordsor21, line 36: Is it forRemoval as well?

MattBDev: That's a question for @me4502 to answer. Hopefully, she can clarify for us.

me4502: Most likely, generally deprecations are for removal in the next major version

there’s not much point having two classes that do the same thing here

PR #2211, opened 2023-05-05
Fix reading of tile entities with "id" instead of "Id" tag

Foorcee:
# Regression of #1220

## Overview
If a schematic contains a tile entity with an "id" instead of an "Id" the previous code would delete the id. The changed code checks which one of the tags is set.

This schematic for example caused this exception:
http://paste.cytooxien.de/ezezeqapek.bash

## Submitter Checklist


TheMeinerLP:
Can you maybe add a check for fallback ?
That could prevent breaking changes.

PierreSchwang (approved):
I'm not really sure why we support "old and corrupted" schematics in first place, as a lowercase Id key was never part of the Sponge Schematic Specification - but as this was already part of the logic before your PR I'm gonna approve that.

NotMyFault:
I'm with Matt and @dordsor21's previous comment.

dordsor21 (approved):

PR #2226, opened 2023-05-15
Avoid many threads blocking on AbstractChangeSet#processSet

SirYwell:
## Description

Previously, write tasks for the history were submitted to the thread pool separately, causing different threads to process the tasks. However, the `processSet` method is synchronized to ensure only one thread at a time actually writes to history.
This results in many threads being blocked and competing for the monitor.

With this change, we collect the tasks in a queue and only process the tasks on one thread instead.
After enqueueing a task, we check if another thread is already processing the queue.

This does not bring immediate performance improvements in form of speed, however it reduces CPU usage due to reduced contention. Additionally, it doesn't take up threads in the thread pool that block most of the time, allowing better performance when working on multiple FAWE actions concurrently.

Alternatives considered:
- Wait for virtual threads, create one vthread per task, switch to locks to prevent pinning -> maybe in future?
- Have a *new* thread for writing -> would further increase the already high number of FAWE threads

## Submitter Checklist

### DIFF
--- a/worldedit-core/src/main/java/com/fastasyncworldedit/core/history/changeset/AbstractChangeSet.java
+++ b/worldedit-core/src/main/java/com/fastasyncworldedit/core/history/changeset/AbstractChangeSet.java
@@ -31,8 +31,6 @@ import com.sk89q.worldedit.world.biome.BiomeType;
 import com.sk89q.worldedit.world.block.BaseBlock;
 import com.sk89q.worldedit.world.block.BlockState;
 import com.sk89q.worldedit.world.block.BlockTypesCache;
-import jdk.jfr.Category;
-import jdk.jfr.Event;
 import org.apache.logging.log4j.Logger;
 
 import java.io.IOException;
@@ -71,13 +69,6 @@ public abstract class AbstractChangeSet implements ChangeSet, IBatchProcessor {
     }
 
     public void closeAsync() {
-        @Category("FAWE")
-        class CloseCall extends Event {
-            public boolean closed;
-        }
-        final CloseCall closeCall = new CloseCall();
-        closeCall.closed = closed;
-        closeCall.commit();
         if (closed) {
             return;
         }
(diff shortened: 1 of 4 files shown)


dordsor21:
Currently postProcessSet (having send-before-history true) solves the issue does it not?

SirYwell:
No, but it's rather only `postProcessSet` where we can fix it this way. `addWriteTask` doesn't make sure that the tasks are processed on one thread only, and as new tasks arrive faster than they are processed (at least for me), other threads will start processing the tasks just to run into the synchronized method.

dordsor21:
Eh yeah that's fair, cleanup of the number of threads FAWE creates isn't a bad thing

worldedit-core/src/main/java/com/fastasyncworldedit/core/history/changeset/AbstractChangeSet.java
```diff
@@ -372,25 +364,58 @@ public Future<?> addWriteTask(final Runnable writeTask, final boolean completeNo
...
+            return task;
+        }
+    }
+
+    private void triggerWorker() {
+        if (workerSemaphore.availablePermits() == 0) {
+            return; // fast path to avoid additional tasks: a worker is already draining the queue
+        }
+        // create a new worker to drain the current queue
+        Fawe.instance().getQueueHandler().submit(() -> drainQueue(false));
```
dordsor21, line 392: We should probably have a synchronized block here, with a check before and after synchronizing to definitely prevent multiple submissions of drainQueue

worldedit-core/src/main/java/com/fastasyncworldedit/core/history/changeset/AbstractChangeSet.java
```diff
@@ -372,25 +364,58 @@ public Future<?> addWriteTask(final Runnable writeTask, final boolean completeNo
...
+                    Thread.currentThread().interrupt();
+                }
+            } else {
+                return; // another thread is draining the queue already, ignore
+            }
+        }
+        try {
+            while (true) {
+                var next = queue.poll();
+                if (next == null) {
```
dordsor21, line 412: Why do we not just have the above in the while loop?

SirYwell: We do not really profit from threads competing for processing the tasks from the queue

dordsor21: This doesn't solve that though, it's the same as if it were just in the while itself no?

SirYwell: I didn't find a good way to pull it into the loop, but I also don't see any benefit from that - did I miss something?

dordsor21: Define next outside the loop and `while ((next = queue.poll()) != null)` (unless I'm missing something). No particular benefit other than fewer lines really and avoiding a while(true)

SirYwell: I see, then I just totally misunderstood what you meant, lol. Will update the code.

dordsor21 (approved):

PR #2243, opened 2023-05-23
feat: configurable image hosts

dordsor21:


worldedit-core/src/main/java/com/fastasyncworldedit/core/util/image/ImageUtil.java
```diff
@@ -229,7 +230,9 @@ public static URI getImageURI(String arg) throws InputParseException {
                 if (arg.contains("imgur.com") && !arg.contains("i.imgur.com")) {
                     arg = "https://i.imgur.com/" + arg.split("imgur.com/")[1] + ".png";
                 }
-                return new URL(arg).toURI();
+                URL url = new URL(arg);
```
SirYwell, line 233: This constructor is deprecated in Java 20, maybe it makes sense to direclty construct the URI and also let the checkImageHost take a URI?

PierreSchwang:
Maybe we should directly support glob patterns and not limit on hosts directly?

dordsor21:
> Maybe we should directly support glob patterns and not limit on hosts directly?

Could do, but not sure if that's more effort than it's worth tbh. Could count on one hand the number of people specifically requesting this haha

worldedit-core/src/main/java/com/fastasyncworldedit/core/util/image/ImageUtil.java
dordsor21: OIC, yeah probably worth that then

NotMyFault (commented):
I don't think adding, even if configurable, settings like these are a great idea. Configured improperly, there are many possible drawbacks.

dordsor21:
It's more with the mind of build team/whitelist servers where it might be quite beneficial to be able to use images from anywhere, rather than having to reupload to imgur every time

OneLiteFeather:
You could also self host an upload solution like an S3 Storage or "picture" server so you could use your own link for that, right?

NotMyFault:
> It's more with the mind of build team/whitelist servers

In such a case, the implementation should be API only, allowing the people needed to extend it on their own.

dordsor21:
Tbh that's too much work for the user, especially given most users don't touch their configs at all in the first place. The warning's there, and FAWE has never shied from assuming a reasonable level of intelligence in its users (and I don't want to start making FAWE for dummies yet tbh). And besides, I'd rather a server OOM be someone's kick to implement reasonable security procedures than an actually dangerous exploit

worldedit-core/src/main/java/com/fastasyncworldedit/core/configuration/Settings.java
```diff
@@ -671,6 +671,14 @@ public static class WEB {
         })
         public int MAX_IMAGE_SIZE = 8294400;
 
+        @Comment({
+                "Whitelist of hostnames to allow images to be downloaded from",
+                " - Adding '*' to the list will allow any host, but this is NOT adviseable",
+                " - Crash exploits exist with malphormed images",
```
SirYwell, line 677:
```suggestion
                " - Crash exploits exist with malformed images",
```

SirYwell (commented):
Small typo, but looks good otherwise. This will also help people who do not trust imgur and rather trust their own services.

SirYwell (approved):

NotMyFault (approved):
Can still revert it if it causes major issues 🤷

PR #2248, opened 2023-05-25
refactor: switch to EnumSet for heightmaps to improve performance

dordsor21:


SirYwell:
Is there a reason for changing the return types/the type of the field?

dordsor21:
> Is there a reason for changing the return types/the type of the field?

No, I guess not. I'll fix that

SirYwell (approved):

NotMyFault (approved):

PR #2266, opened 2023-06-04
chore: improve queue documentation and submit history to better queue

dordsor21:


worldedit-core/src/main/java/com/fastasyncworldedit/core/history/changeset/AbstractChangeSet.java
```diff
@@ -384,12 +384,12 @@ public Future<?> addWriteTask(final Runnable writeTask, final boolean completeNo
         }
     }
 
-    private void triggerWorker() {
+    private synchronized void triggerWorker() {
```
SirYwell, line 387: What's the intention here?

dordsor21: Eh, tbf there's no real need for it. It's just rarely create an extra task that'll complete quickly I suppose

NotMyFault (approved):

Issue #2222, opened 2023-05-13
//copy - java.nio.channels.OverlappingFileLockExcption: null


skrevt:
### Server Implementation

Paper

### Server Version

1.19.4

### Describe the bug

If I copy something with //copy, then I am not on the server for a while and join again and try to copy something this error appears

### To Reproduce

1. Join Server
2. Copy Selection
3. Go offline for around 30-60min
4. Go Online
5. Copy again
6. Error
7. Restart Server
8. all wokrs

### Expected behaviour

Fix

### Screenshots / Videos

![Screenshot_4](https://github.com/IntellectualSites/FastAsyncWorldEdit/assets/88806692/bf56775f-e530-44ee-9f8f-4e60804c897c)
![Screenshot_5](https://github.com/IntellectualSites/FastAsyncWorldEdit/assets/88806692/46ff3e9c-a61f-4524-a429-81d5f9484e02)

### Fawe Debugpaste

https://athion.net/ISPaster/paste/view/93dc8edd01cc4565aef94f6f82ca60b9

### Fawe Version

FAWE-1.20

### Checklist

- [X] I have included a Fawe debugpaste.
- [X] I am using the newest build from https://ci.athion.net/job/FastAsyncWorldEdit/ and the issue still persists.

PR #2267, opened 2023-06-04
feat: implement a player-specific queue for clipboard IO tasks

dordsor21:
- Addresses #2222 (hopefully fixes)

### DIFF
--- a/worldedit-core/src/main/java/com/fastasyncworldedit/core/Fawe.java
+++ b/worldedit-core/src/main/java/com/fastasyncworldedit/core/Fawe.java
@@ -12,7 +12,9 @@ import com.fastasyncworldedit.core.util.RandomTextureUtil;
 import com.fastasyncworldedit.core.util.TaskManager;
 import com.fastasyncworldedit.core.util.TextureUtil;
 import com.fastasyncworldedit.core.util.WEManager;
+import com.fastasyncworldedit.core.util.task.KeyQueuedExecutorService;
 import com.github.luben.zstd.Zstd;
+import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import com.sk89q.worldedit.WorldEdit;
 import com.sk89q.worldedit.internal.util.LogManagerCompat;
 import net.jpountz.lz4.LZ4Factory;
@@ -33,6 +35,7 @@ import java.lang.management.MemoryPoolMXBean;
 import java.lang.management.MemoryUsage;
 import java.util.Date;
 import java.util.List;
+import java.util.UUID;
 import java.util.concurrent.TimeUnit;
 
 /**
@@ -86,6 +89,15 @@ public class Fawe {
      * The platform specific implementation.
      */
     private final IFawe implementation;
+    private final KeyQueuedExecutorService<UUID> clipboardExecutor = Settings.settings().CLIPBOARD.USE_DISK
+            ? new KeyQueuedExecutorService<>(
+            1,
+            Settings.settings().QUEUE.PARALLEL_THREADS,
+            0L,
+            TimeUnit.MILLISECONDS,
+            new ThreadFactoryBuilder().setNameFormat("fawe-clipboard-%d").build()
+    )
+            : null;
     private FaweVersion version;
     private TextureUtil textures;
     private QueueHandler queueHandler;
@@ -426,6 +438,17 @@ public class Fawe {
      */
     public Thread setMainThread() {
         return this.thread = Thread.currentThread();
+    }
+
+    /**
+     * Gets the executor used for clipboard IO if clipboard on disk is enabled or null
+     *
+     * @return Executor used for clipboard IO if clipboard on disk is enabled or null
+     * @since TODO
+     */
+    @Nullable
+    public KeyQueuedExecutorService<UUID> getClipboardExecutor() {
+        return this.clipboardExecutor;
     }
 
 }
--- a/worldedit-core/src/main/java/com/fastasyncworldedit/core/Fawe.java
+++ b/worldedit-core/src/main/java/com/fastasyncworldedit/core/Fawe.java
@@ -36,6 +36,8 @@ import java.lang.management.MemoryUsage;
 import java.util.Date;
 import java.util.List;
 import java.util.UUID;
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 
 /**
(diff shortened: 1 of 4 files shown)


NotMyFault:
Hey @skrevt,

do you mind testing with the PR build whether this resolves your issue or not?
Download FAWE from https://ci.athion.net/view/%20%20FastAsyncWorldEdit/job/FastAsyncWorldEdit-Pull-Requests/view/change-requests/job/PR-2267/lastSuccessfulBuild/artifact/artifacts/FastAsyncWorldEdit-Bukkit-2.6.2-SNAPSHOT.jar

dordsor21:
The clipboard issue will be intermittent and isn't particularly easy to replicate, so it's more an over-time test kinda thing

worldedit-core/src/main/java/com/fastasyncworldedit/core/util/task/KeyQueuedExecutorService.java
```diff
@@ -0,0 +1,239 @@
...
+    public Future<?> submit(@Nonnull K key, @Nonnull Runnable task) {
+        FutureTask<Void> ftask = newTaskFor(task, null);
+        execute(key, ftask);
+        return ftask;
+    }
+
+    public void execute(@Nonnull K key, @Nonnull FutureTask<?> command) {
+        KeyRunner runner;
+        synchronized (keyQueue) {
+            runner = keyQueue.merge(key, new KeyRunner(key), (existing, theNew) -> {
```
SirYwell, line 177: Looks like computeIfAbsent would be a better method here?

worldedit-core/src/main/java/com/sk89q/worldedit/entity/Player.java
```diff
@@ -437,14 +437,16 @@ default void unregister() {
...
-                    WorldEdit.getInstance().getExecutorService().submit(() -> {
+                    Fawe.instance().getClipboardExecutor().submit(getUniqueId(), () -> {
                         doc.close(); // Ensure closed before deletion
                         doc.getFile().delete();
                     });
                 }
             }
-        } else if (Settings.settings().CLIPBOARD.DELETE_ON_LOGOUT || Settings.settings().CLIPBOARD.USE_DISK) {
-            WorldEdit.getInstance().getExecutorService().submit(() -> session.setClipboard(null));
+        } else if (Settings.settings().CLIPBOARD.USE_DISK) {
```
SirYwell, line 446: I guess this is fine for now as it was like that before, but actually the clipboard can be set by plugins to arbitrary implementations, so relying on the settings isn't exactly correct.

worldedit-core/src/main/java/com/fastasyncworldedit/core/util/task/KeyQueuedExecutorService.java
```diff
@@ -0,0 +1,239 @@
...
+        }
+
+        void add(FutureTask<?> task) {
+            if (!tasks.add(task)) {
+                throw new RejectedExecutionException(rejection());
+            }
+        }
+
+        synchronized void triggerRun() {
+            Runnable task = tasks.remove();
```
SirYwell, line 201: `remove()` throws an exception if the queue is empty, what's the intention here?

worldedit-core/src/main/java/com/fastasyncworldedit/core/util/task/KeyQueuedExecutorService.java
```diff
@@ -0,0 +1,239 @@
...
+                    keyQueue.remove(key);
+                }
+                throw new RejectedExecutionException(rejection(), e);
+            }
+        }
+
+        private void run(Runnable task) {
+            parent.execute(() -> {
+                task.run();
+                Runnable next = tasks.remove();
```
SirYwell, line 218: Same as above, you probably want to use `poll()` instead

worldedit-core/src/main/java/com/fastasyncworldedit/core/util/task/KeyQueuedExecutorService.java
```diff
@@ -0,0 +1,239 @@
...
+     *                                  {@code keepAliveTime < 0}<br>
+     *                                  {@code maximumPoolSize <= 0}<br>
+     *                                  {@code maximumPoolSize < corePoolSize}
+     * @throws NullPointerException     if {@code workQueue}
+     *                                  or {@code threadFactory} is null
+     */
+    public KeyQueuedExecutorService(
+            int corePoolSize, int maximumPoolSize, long keepAliveTime, TimeUnit unit, ThreadFactory threadFactory
+    ) {
+        parent = new ThreadPoolExecutor(
```
SirYwell, line 99: would it make sense to just have a ctor that takes an ExecutorService directly?

worldedit-core/src/main/java/com/fastasyncworldedit/core/Fawe.java
```diff
@@ -86,6 +89,15 @@ public class Fawe {
      * The platform specific implementation.
      */
     private final IFawe implementation;
+    private final KeyQueuedExecutorService<UUID> clipboardExecutor = Settings.settings().CLIPBOARD.USE_DISK
```
SirYwell, line 92: Maybe initialize this in the ctor for consistency? But also same discussion as in other places, other plugins might use clipboard on disk features without the setting being enabled.

SirYwell (changes requested):

worldedit-core/src/main/java/com/fastasyncworldedit/core/util/task/KeyQueuedExecutorService.java
```diff
@@ -0,0 +1,164 @@
...
+            }
+        }
+
+        private void run(Runnable task) {
+            parent.execute(() -> {
+                task.run();
+                Runnable next = tasks.poll();
+                if (next == null) {
+                    synchronized (keyQueue) {
+                        next = tasks.remove();
```
SirYwell, line 146: I think this is still wrong. If the queue is still empty here, it throws an exception instead of removing the key.

worldedit-core/src/main/java/com/fastasyncworldedit/core/util/task/KeyQueuedExecutorService.java
```diff
@@ -0,0 +1,164 @@
...
+        execute(key, ftask);
+        return ftask;
+    }
+
+    public void execute(@Nonnull K key, @Nonnull FutureTask<?> command) {
+        KeyRunner runner;
+        synchronized (keyQueue) {
+            runner = keyQueue.computeIfAbsent(key, KeyRunner::new);
+            runner.add(command);
+            runner.triggerRun();
```
SirYwell, line 106: If I see this right, this might be called while the `command` task was already processed by a previous call to `triggerRun`, causing an exception. (and alternatively, run multiple commands for one player concurrently)

I think we need some logic to ensure only one worker is draining the queue here

SirYwell (changes requested):

worldedit-core/src/main/java/com/fastasyncworldedit/core/util/task/KeyQueuedExecutorService.java
dordsor21: Meant to replace all remove with poll, mb

NotMyFault (approved):
Let's see how it turns out!

PR #2270, opened 2023-06-05
fix: correctly processSet with blacklist when required

dordsor21:


worldedit-core/src/main/java/com/fastasyncworldedit/core/extent/MultiRegionExtent.java
```diff
@@ -172,7 +172,7 @@ public IChunkSet processSet(IChunk chunk, IChunkGet get, IChunkSet set) {
             set = intersection.processSet(chunk, get, set);
         }
         if (disallowedIntersection != null) {
-            intersection.processSet(chunk, get, set);
+            set = intersection.processSet(chunk, get, set, true);
```
SirYwell, line 175: shouldn't that be `disallowedIntersection` here?

SirYwell (changes requested):

worldedit-core/src/main/java/com/fastasyncworldedit/core/extent/MultiRegionExtent.java
dordsor21: Yes

SirYwell (approved):

NotMyFault (approved):

PR #2285, opened 2023-06-11
feat: prevent edits outside +/- 30,000,000 blocks

dordsor21:


NotMyFault (approved):

worldedit-core/src/main/java/com/fastasyncworldedit/core/queue/implementation/preloader/AsyncPreloader.java
```diff
@@ -3,6 +3,7 @@
 import com.fastasyncworldedit.core.Fawe;
 import com.fastasyncworldedit.core.configuration.Settings;
 import com.fastasyncworldedit.core.util.FaweTimer;
+import com.fastasyncworldedit.core.util.MathMan;
```
IronApollo, line 6: Stray import?

IronApollo (approved):
Looks good

worldedit-core/src/main/java/com/fastasyncworldedit/core/queue/implementation/preloader/AsyncPreloader.java
dordsor21: Ah yes

PR #2301, opened 2023-06-17
fix: towny max bounds are exclusive 

### LINKED ISSUE #2293: Restricted Regions - Towny FAWE not between plots, plot boundary crossed to wilderniss
### Server Implementation

Paper

### Server Version

1.19.4

### Describe the bug

FAWE is used together with Restricted Regions via Towny, the respective player has no Op and has received the respective rights, within a plot [townblocksize=16 /exatly 1 Chunk] all FAWE functions work as far as I can tell. However, if you want to edit between two plots, for example, only the plot in which you are standing at the time of the change will be edited. Accordingly, it is not possible to work in several plots at the same time. 

I also noticed another problem. Although it seems that you can only change something in the plot you are standing in at the time, there are always blocks sticking out of the plot to the south and east if the selection also covers this area. So the towny boundary seems to have a problem with south and east, which also allows editing in the wilderness [around the first row of blocks of a towny block in the given directions] and also does not allow cross-plot editing [between towny plots]. [see picture for both cases]

### To Reproduce

1. Select 3x3 Chunks [all part of your towny city]
2. Go to the chunk in the middle
3. use //set red_concrete
4. Only the chunk your are standing on will be edited [with overlapps in east & south] [townblocks are exactly one chunk]
[see picture]

Optional: [same result - regarding the overlapping]
1. Select 3x3 Chunks [only middle one city - rest wildness]
2. Go to your city chunk
3. use //set red_concrete
4. The middle chunk is processed but also with overlaps to the south and east which are not intended.
[see picture - Behaves in the same way as the upper, regarding the overlapping.][townblocks are exactly one chunk]

### Expected behaviour

1. it is possible to work across several city plots and not only in the plot where you are standing.
2. no overlapping blocks in southern or eastern direction outside of a plot that does not belong to you [in this case noticed by problem 1].

### Screenshots / Videos

Blue = Total Selection
Red = Blocks Placed after //set red_concrete
Top = North
Picture = All city plots

### Error log (if applicable)

_No response_

### Fawe Debugpaste

https://athion.net/ISPaster/paste/view/1de2cd4bcc454e6b82d87853ddbc929a

### Fawe Version

FastAsyncWorldEdit-Bukkit-2.6.4-SNAPSHOT-480

### Checklist

### Anything else?

If there are any questions I will answer as soon as possible.

dordsor21:
- Fixes #2293

### DIFF
--- a/worldedit-bukkit/src/main/java/com/fastasyncworldedit/bukkit/regions/TownyFeature.java
+++ b/worldedit-bukkit/src/main/java/com/fastasyncworldedit/bukkit/regions/TownyFeature.java
@@ -1,7 +1,6 @@
 package com.fastasyncworldedit.bukkit.regions;
 
 import com.fastasyncworldedit.core.regions.FaweMask;
-import com.palmergames.bukkit.towny.Towny;
 import com.palmergames.bukkit.towny.TownyAPI;
 import com.palmergames.bukkit.towny.object.Resident;
 import com.palmergames.bukkit.towny.object.Town;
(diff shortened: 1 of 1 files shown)


NotMyFault (approved):
Not a Towny user here, but looks good logic wise.

SirYwell (approved):

PR #2304, opened 2023-06-18
refactor: efficiency improvements to some editsession methods

dordsor21:
No major speed improvements, but sorta reduced memory footprint and quite large reductions in number of numeric calculations done


NotMyFault (approved):
Math wise, looks good to me, but performance wise, untested.

PR #2308, opened 2023-06-18
chore: Address kotlin String#capitalize() deprecation

### LINKED ISSUE #2283: Address `String#capitalize()` deprecation in `LibsConfig.kt`
`project.name.capitalize()` is deprecated in
https://github.com/IntellectualSites/FastAsyncWorldEdit/blob/083f8a4dd8b5ccb315da66603ee32fdbe108e02c/buildSrc/src/main/kotlin/LibsConfig.kt#L207
marked for removal. It should be updated according to a non-deprecated version.

dordsor21:
- Fixes #2283
 - Don't bother with the longer version that allows Locales, we don't need that

### DIFF
--- a/buildSrc/src/main/kotlin/LibsConfig.kt
+++ b/buildSrc/src/main/kotlin/LibsConfig.kt
@@ -23,6 +23,7 @@ import org.gradle.kotlin.dsl.provideDelegate
 import org.gradle.kotlin.dsl.register
 import org.gradle.kotlin.dsl.the
 import org.gradle.plugins.signing.SigningExtension
+import java.util.*
 import javax.inject.Inject
 
 fun Project.applyLibrariesConfiguration() {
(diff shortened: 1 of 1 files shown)


buildSrc/src/main/kotlin/LibsConfig.kt
```diff
@@ -23,6 +23,7 @@ import org.gradle.kotlin.dsl.provideDelegate
 import org.gradle.kotlin.dsl.register
 import org.gradle.kotlin.dsl.the
 import org.gradle.plugins.signing.SigningExtension
+import java.util.*
```
NotMyFault, line 26:
```suggestion
```
Unused? `Chart#` comes from Chart.kt.

buildSrc/src/main/kotlin/LibsConfig.kt
```diff
@@ -204,7 +205,7 @@ fun Project.applyLibrariesConfiguration() {
                 from(libsComponent)
 
                 group = "com.fastasyncworldedit"
-                artifactId = "FastAsyncWorldEdit-Libs-${project.name.capitalize()}"
+                artifactId ="FastAsyncWorldEdit-Libs-${project.name.replaceFirstChar(Char::titlecase)}"
```
NotMyFault, line 208:
```suggestion
                artifactId = "FastAsyncWorldEdit-Libs-${project.name.replaceFirstChar(Char::titlecase)}"
```

NotMyFault (approved):
Thanks!

Issue #2332, opened 2023-07-05
FAWE becomes unresponsive


SecretlyJealous:
### Server Implementation

Paper

### Server Version

1.20.1

### Describe the bug

This is the second day in a row I've had this issue so I thought I would report it since I was able to get a thread dump which should hopefully help. For some reason fawe stops working but I'm not sure what causes it, I'm unable to use /fawe threaddump or any worldedit related command. Using a worldedit command produces no feedback in chat. It's acting as if the plugin is stuck or something. In the log file I see an error but I'm not sure if it's related to this problem or not.

Log: https://pastebin.com/sEwHH4hD
Thread dump from visualvm: https://www.dropbox.com/s/xbccnbe8zoxv0gn/fawe%20commands%20not%20working.tdump?dl=1

### To Reproduce

1. Not sure

### Expected behaviour

WorldEdit commands should continue working

### Fawe Debugpaste

The fawe commands don't work when this happens so I'm unable to run this

### Fawe Version

https://ci.athion.net/job/FAWE-PS-v7/24

### Checklist

- [X] I have included a Fawe debugpaste.
- [X] I am using the newest build from https://ci.athion.net/job/FastAsyncWorldEdit/ and the issue still persists.

OneLiteFeather:
Can you do a /fawe debugpaste after a restart or send us the beginning of the log so we can look deeper into it? @SecretlyJealous

dordsor21:
Nah based on the threads there, it looks like There's a chunk taking a while to load (paper tings I guess, it seems to have forgotten about the chunk request), so if a brush is used a lot more afterwards, then the queue will completely block, preventing new submissions, and also preventing history from being written (assuming history db is enabled) causing the soft-"crash" of FAWE seen in this instance.

OneLiteFeather:
Well then I guess this can be approved as a bug. Thank you for the explaination 👍

SirYwell:
Did you by any chance unload the world where the FAWE action took place? Currently, this seems to be the only reason why the future wouldn't get completed.

PR #2339, opened 2023-07-07
feat: re-submit chunk load request after 10s after checking world is loaded

dordsor21:
- #2332 makes it seem like paper forgets to load a chunk sometimes
 - resubmit chunk load request after a second to attempt to counter this


NotMyFault (dismissed):

SirYwell:
This doesn't resubmit anything, no? It's basically just  *wait one second for the future to complete, if completed return, otherwise wait for the same future to complete without timeout*.

I also think if that's actually an issue with paper, this should be fixed in paper (or, if we use something wrong here, we should use it right).

dordsor21:
Unsure really, but the thread dump in the issue doesn't list any threads doing anything to do with chunk loading. It _does_ seem to be on Folia though, so I don't know if that might have anything to do with it?

SecretlyJealous:
I'm not using folia I'm using paper 1.20.1, it's weird that the thread dump mentions folia though?

SirYwell:
> I'm not using folia I'm using paper 1.20.1, it's weird that the thread dump mentions folia though?

It might come from a plugin which just uses suboptimal naming for threads. It might be worth to track that down and report it to the developer.

SirYwell (approved):

NotMyFault (approved):

PR #2346, opened 2023-07-09
feat: implement a new "type swap" pattern

dordsor21:
Open to better names for this

### DIFF
--- a/worldedit-core/src/main/java/com/fastasyncworldedit/core/function/pattern/TypeSwapPattern.java
+++ b/worldedit-core/src/main/java/com/fastasyncworldedit/core/function/pattern/TypeSwapPattern.java
@@ -55,67 +55,49 @@ public class TypeSwapPattern extends AbstractExtentPattern {
     @Override
     public boolean apply(Extent extent, BlockVector3 get, BlockVector3 set) throws WorldEditException {
         BlockState existing = get.getBlock(extent);
-        String oldId = existing.getBlockType().getId();
-        String newId = oldId;
-        if (inputPattern != null) {
-            newId = inputPattern.matcher(oldId).replaceAll(outputString);
-        } else if (inputs != null && inputs.length > 0) {
-            for (String input : inputs) {
-                newId = oldId.replace(input, outputString);
-            }
-        } else {
-            newId = oldId.replace(inputString, outputString);
-        }
-        if (newId.equals(oldId)) {
+        BlockState newBlock = getNewBlock(existing);
+        if (newBlock == null) {
             return false;
         }
-        BlockType newType = BlockTypes.get(newId);
-        if (newType == null) {
-            return false;
-        }
-        BlockState newBlock = newType.getDefaultState().withProperties(existing);
         return set.setBlock(extent, newBlock);
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


SirYwell:
Regex from user input opens up for DoS attacks. In this case, I think it should be behind a permission.

dordsor21:
There's a few other places we'd need to do that then...

SirYwell:
I'm not aware of other places where regexes are evaluated on a per-position basis, but if so, they should probably revisited too.

dordsor21:
I was specifically thinking of block masking https://github.com/IntellectualSites/FastAsyncWorldEdit/blob/main/worldedit-core/src/main/java/com/fastasyncworldedit/core/function/mask/BlockMaskBuilder.java#L75

SirYwell:
> I was specifically thinking of block masking [`main`/worldedit-core/src/main/java/com/fastasyncworldedit/core/function/mask/BlockMaskBuilder.java#L75](https://github.com/IntellectualSites/FastAsyncWorldEdit/blob/main/worldedit-core/src/main/java/com/fastasyncworldedit/core/function/mask/BlockMaskBuilder.java?rgh-link-date=2023-07-09T18%3A25%3A03Z#L75)

It looks like this is only called inside of an executor + timeout, so that should be fine (especially as it's only called on construction, not on block placement).

worldedit-core/src/main/java/com/fastasyncworldedit/core/function/pattern/TypeSwapPattern.java
```diff
@@ -30,24 +33,36 @@ public class TypeSwapPattern extends AbstractExtentPattern {
...
     public boolean apply(Extent extent, BlockVector3 get, BlockVector3 set) throws WorldEditException {
         BlockState existing = get.getBlock(extent);
         String oldId = existing.getBlockType().getId();
-        String newId;
+        String newId = oldId;
         if (inputPattern != null) {
             newId = inputPattern.matcher(oldId).replaceAll(outputString);
+        } else if (inputs != null && inputs.length > 0) {
+            for (String input : inputs) {
+                newId = oldId.replace(input, outputString);
```
SirYwell, line 64: I assume this should be 
```suggestion
                newId = newId.replace(input, outputString);
```
?

SirYwell: Also that handling is missing in the other methods, can this somehow be moved into one method to avoid inconsistencies?

dordsor21: The for loop? This is the only case where there's anything to loop over

SirYwell: But why? What's the difference between `apply` and the `applyBlock` methods?

dordsor21: Ah okay I see what you mean, had forgotten about the other one

NotMyFault (approved):

PR #2347, opened 2023-07-09
fix: fix command brush

### LINKED ISSUE #2318: The `/brush command` is broken.
### Server Implementation

Paper

### Server Version

1.20

### Describe the bug

According to the documentation, the expected usage of `/brush command` subcommand looks like this:
```
/br command 2 "set 1"
```
However, when run in game, it does nothing but selecting the region. No errors in console, no `/set 1`. `"/set 1"` and `"//set 1"` commands also do not work. According to @NotMyFault at #1952:

> According to the code, the supported syntax would be something like `/br command 5 "br sphere 5 5"`

I'm not sure about how is this the supposed syntax because it does not make much sense. The specialized brush command which just sets the brush is completely useless. Moreover, the [code I looked at](https://github.com/IntellectualSites/FastAsyncWorldEdit/blame/main/worldedit-core/src/main/java/com/fastasyncworldedit/core/command/tool/brush/CommandBrush.java) looks completely normal and I would expect it to act as a normal, proper brush. As further noted in that issue (incorrectly noted, but still acknowledged), combining commands through `;` also does not work:
 ```
 /br command 0 "br sphere stone 1; br sphere glass 2"
 ```
 only applies the first brush.
 
 Finally, the placeholders are apparently being interpolated only during the brush creation. 
 ```
 /br command 0 "br sphere {x} 1"
 ```
 Will put a blob of blocks based on where the command was executed, and not where the brush is pointing at.
 
This leads me to believe that the `brush command` executes its logic only during initialization, but it's still a mystery as to why (the source code looks pretty normal to me) and why it only works "sometimes" (why `/br cmd 0 "set 1"` does not run `/set 1` at startup, but `/br cmd 0 "br s 1 1"` does?).

### To Reproduce

Assuming on a paper server with a single plugin (FAWE) installed:

1. Take any tool, like a stone pickaxe
2. Run `/br command 2 "set 1"`
3. Use the brush and observe how it does not run `/set 1`

See other complaints above.

### Expected behaviour

The `/br command <r> <commands>` subcommand binds a tool to do the following every use:
1. Set the player selection to a cube of radius `r`, with a center at the target block
2. Replaces the placeholders in `<commands>` argument with the coordinates of the target block, the world name and the size of the selection
3. Runs the commands specified in `<commands>` one by one, using `;` as a separator
One might expect the selection to be reset to the one before the tool use, but I will argue against it for the sake of simplifying the implementation.

### Screenshots / Videos

_No response_

### Error log (if applicable)

_No response_

### Fawe Debugpaste

https://athion.net/ISPaster/paste/view/d87e3b252ec2423db2cceac3d2088ba2

### Fawe Version

2.6.4-SNAPSHOT-494;05d5ad1

### Checklist

### Anything else?

_No response_

dordsor21:
- fixes #2318

### DIFF
--- a/worldedit-core/src/main/java/com/fastasyncworldedit/core/command/tool/brush/CommandBrush.java
+++ b/worldedit-core/src/main/java/com/fastasyncworldedit/core/command/tool/brush/CommandBrush.java
@@ -22,9 +22,27 @@ import java.util.List;
 public class CommandBrush implements Brush {
 
     private final String command;
+    private final boolean print;
 
+    /**
+     * New instance
+     *
+     * @deprecated Use {@link CommandBrush#CommandBrush(String, boolean)}
+     */
+    @Deprecated(forRemoval = true)
     public CommandBrush(String command) {
-        this.command = command.charAt(0) == '/' ? "/" + command : command;
+        this(command, false);
+    }
+
+    /**
+     * New instance
+     *
+     * @param command command to run, or commands split by ';'
+     * @param print   if output should be printed to the actor for the run commands
+     */
+    public CommandBrush(String command, boolean print) {
+        this.command = command;
+        this.print = print;
     }
 
     @Override
@@ -36,7 +54,7 @@ public class CommandBrush implements Brush {
                 position.subtract(radius, radius, radius),
                 position.add(radius, radius, radius)
         );
-        String replaced = command.replace("{x}", position.getBlockX() + "")
+        String replaced = command.replace("{x}", Integer.toString(position.getBlockX()))
                 .replace("{y}", Integer.toString(position.getBlockY()))
                 .replace("{z}", Integer.toString(position.getBlockZ()))
                 .replace("{world}", editSession.getWorld().getName())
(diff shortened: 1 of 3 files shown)


worldedit-core/src/main/java/com/fastasyncworldedit/core/command/tool/brush/CommandBrush.java
```diff
@@ -46,21 +64,19 @@ public void build(EditSession editSession, BlockVector3 position, Pattern patter
...
-        ));
+        );
+        if (!print) {
+            wePlayer = new SilentPlayerWrapper(wePlayer);
+        }
         List<String> cmds = StringMan.split(replaced, ';');
         for (String cmd : cmds) {
-            CommandEvent event = new CommandEvent(wePlayer, cmd);
+            cmd = cmd.charAt(0) != '/' ? "/" + cmd : cmd;
+            cmd = cmd.charAt(1) == '/' ? cmd.substring(1) : cmd;
```
SirYwell, line 78: this can be out of bounds

SirYwell (approved):

NotMyFault (approved):

PR #2351, opened 2023-07-13
fix: fix adding/getting masks to/from MaskExtent when it is a processor

### LINKED ISSUE #2350: /mask is ignored by brush
### Server Implementation

Paper

### Server Version

1.20.1

### Describe the bug

When using a mask, with a brush, it seems like the mask is sometimes ignored. This does not seem to be limited to the >grass_block mask but effects at least a mask of 0 as well.

Using the brush sometimes places the grass correctly. So multiple tries might me necessary. In my testing it took a maximum of 10 attempts for it to fail. When using lapis_block it seems like it is more likely to fail.

Swapping the mask and the gmask yields a similar result as "/mask 0" is then ignored and existing blocks are replaced.

### To Reproduce

/br sphere tall_grass[half=lower] 5
/mask >grass_block
/gmask 0

OR

/br sphere lapis_block 5
/mask >grass_block
/gmask 0

Place blocks using the specified brush.

### Expected behaviour

Masks are respected when using a brush.

More specifically: Blocks are only placed on Grass Blocks and not above them.

### Screenshots / Videos

image
image

### Error log (if applicable)

_No response_

### Fawe Debugpaste

 https://athion.net/ISPaster/paste/view/0bc378b6912947cfb0673488c33fd9b5

### Fawe Version

2.6.5_SNAPSHOT-521:0a19f64

### Checklist

### Anything else?

_No response_

dordsor21:
- Fixes #2350

### DIFF
--- a/worldedit-core/src/main/java/com/fastasyncworldedit/core/util/ProcessorTraverser.java
+++ b/worldedit-core/src/main/java/com/fastasyncworldedit/core/util/ProcessorTraverser.java
@@ -3,6 +3,8 @@ package com.fastasyncworldedit.core.util;
 import com.fastasyncworldedit.core.extent.processor.BatchProcessorHolder;
 import com.fastasyncworldedit.core.extent.processor.MultiBatchProcessor;
 import com.fastasyncworldedit.core.queue.IBatchProcessor;
+import com.sk89q.worldedit.internal.util.LogManagerCompat;
+import org.apache.logging.log4j.Logger;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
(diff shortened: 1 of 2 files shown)


worldedit-core/src/main/java/com/fastasyncworldedit/core/util/ProcessorTraverser.java
```diff
@@ -0,0 +1,48 @@
...
+    private final T root;
+
+    public ProcessorTraverser(@Nonnull T root) {
+        this.root = root;
+    }
+
+    @SuppressWarnings("unchecked")
+    public @Nullable <U extends IBatchProcessor> U find(Class<U> clazz) {
+        try {
+            Queue<IBatchProcessor> processors = new ArrayDeque<>(5);
```
SirYwell, line 23: Why 5?

worldedit-core/src/main/java/com/fastasyncworldedit/core/util/ProcessorTraverser.java
```diff
@@ -0,0 +1,48 @@
...
+                processors.addAll(multiBatchProcessor.getBatchProcessors());
+            } else if (this.root instanceof BatchProcessorHolder holder) {
+                processors.add(holder.getProcessor());
+            } else {
+                return clazz.isAssignableFrom(this.root.getClass()) ? clazz.cast(this.root) : null;
+            }
+            IBatchProcessor processor;
+            while ((processor = processors.poll()) != null) {
+                if (clazz.isAssignableFrom(processor.getClass())) {
+                    return (U) processor;
```
SirYwell, line 34:
```suggestion
                    return clazz.cast(processor);
```
This way we can get rid of the SuppressWarnings too I think

worldedit-core/src/main/java/com/fastasyncworldedit/core/util/ProcessorTraverser.java
```diff
@@ -0,0 +1,48 @@
...
+            Queue<IBatchProcessor> processors = new ArrayDeque<>(5);
+            if (this.root instanceof MultiBatchProcessor multiBatchProcessor) {
+                processors.addAll(multiBatchProcessor.getBatchProcessors());
+            } else if (this.root instanceof BatchProcessorHolder holder) {
+                processors.add(holder.getProcessor());
+            } else {
+                return clazz.isAssignableFrom(this.root.getClass()) ? clazz.cast(this.root) : null;
+            }
+            IBatchProcessor processor;
+            while ((processor = processors.poll()) != null) {
```
SirYwell, line 32: I think this could be a do-while, reducing the duplicated code here?

worldedit-core/src/main/java/com/fastasyncworldedit/core/util/ProcessorTraverser.java
```diff
@@ -0,0 +1,48 @@
...
+                    return (U) processor;
+                } else if (processor instanceof MultiBatchProcessor multiProcessor) {
+                    processors.addAll(multiProcessor.getBatchProcessors());
+                } else if (processor instanceof BatchProcessorHolder holder) {
+                    processors.add(holder.getProcessor());
+                }
+            }
+            return null;
+        } catch (Throwable e) {
+            e.printStackTrace();
```
SirYwell, line 43: Should use a logger here instead, and probably add context

SirYwell (changes requested):

worldedit-core/src/main/java/com/fastasyncworldedit/core/util/ProcessorTraverser.java
```diff
@@ -0,0 +1,48 @@
...
+public class ProcessorTraverser<T extends IBatchProcessor> {
+
+    private final T root;
+
+    public ProcessorTraverser(@Nonnull T root) {
+        this.root = root;
+    }
+
+    @SuppressWarnings("unchecked")
+    public @Nullable <U extends IBatchProcessor> U find(Class<U> clazz) {
```
SirYwell, line 21: oh and the `@Nullable` should be at the return type I guess

worldedit-core/src/main/java/com/fastasyncworldedit/core/util/ProcessorTraverser.java
dordsor21: I did a quick count of the processors we almost always add, but tbh there'd realistically be no perf hit with any number

SirYwell (approved):

NotMyFault (approved):

PR #2353, opened 2023-07-13
fix: do not clash ChunkHolder recycling with processors that extend outside the chunk

dordsor21:
- Extend getChunkLock usage to capture more operations on the chunks
 - 5 ms grace period before recycling chunk
 - unlock ChunkHolder sooner when being called (with copying both the set and (immediately) the get this can be done now)
 - no longer a need for ReentrantWrappedStampedLock


worldedit-core/src/main/java/com/fastasyncworldedit/core/queue/implementation/SingleThreadQueueExtent.java
```diff
@@ -52,6 +54,7 @@ public class SingleThreadQueueExtent extends ExtentBatchProcessorHolder implemen
     // private static final ConcurrentLinkedQueue<IChunk> CHUNK_POOL = new ConcurrentLinkedQueue<>();
     // Chunks currently being queued / worked on
     private final Long2ObjectLinkedOpenHashMap<IQueueChunk> chunks = new Long2ObjectLinkedOpenHashMap<>();
+    private final Queue<Long> chunkOrder = new ArrayDeque<>(Settings.settings().QUEUE.TARGET_SIZE);
```
SirYwell, line 57: Why do we need this? Shouldn't the linked hashmap already ensure that we remove oldest first?

SirYwell (changes requested):

worldedit-core/src/main/java/com/fastasyncworldedit/core/queue/implementation/SingleThreadQueueExtent.java
dordsor21: Ah yeah, I had a moment

SirYwell (approved):
Nice cleanup

NotMyFault (approved):

PR #2361, opened 2023-07-15
fix: make IdMask thread-safe

dordsor21:


worldedit-core/src/main/java/com/fastasyncworldedit/core/function/mask/IdMask.java
```diff
@@ -7,7 +7,8 @@
 
 public class IdMask extends AbstractExtentMask implements ResettableMask {
 
-    private transient int id = -1;
+    private final Object lock = new Object();
+    private transient Integer id = -1;
```
SirYwell, line 11: Why `Integer`?

SirYwell (commented):
If I understand this right, wouldn't it be possible to simply use an AtomicInteger to ensure it is only set once? e.g.

```diff
Index: worldedit-core/src/main/java/com/fastasyncworldedit/core/function/mask/IdMask.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/worldedit-core/src/main/java/com/fastasyncworldedit/core/function/mask/IdMask.java b/worldedit-core/src/main/java/com/fastasyncworldedit/core/function/mask/IdMask.java
--- a/worldedit-core/src/main/java/com/fastasyncworldedit/core/function/mask/IdMask.java	(revision e1663cf5321d4648094658d92017cf11340610df)
+++ b/worldedit-core/src/main/java/com/fastasyncworldedit/core/function/mask/IdMask.java	(date 1689500235537)
@@ -5,10 +5,12 @@
 import com.sk89q.worldedit.function.mask.Mask;
 import com.sk89q.worldedit.math.BlockVector3;
 
+import java.util.concurrent.atomic.AtomicInteger;
+
 public class IdMask extends AbstractExtentMask implements ResettableMask {
 
-    private final Object lock = new Object();
-    private transient Integer id = -1;
+    private final static int SENTINEL = -1;
+    private final AtomicInteger id = new AtomicInteger(-1);
 
     public IdMask(Extent extent) {
         super(extent);
@@ -16,18 +18,9 @@
 
     @Override
     public boolean test(Extent extent, BlockVector3 vector) {
-        if (id != -1) {
-            return extent.getBlock(vector).getInternalBlockTypeId() == id;
-        } else {
-            synchronized (lock) {
-                if (id != -1) {
-                    return extent.getBlock(vector).getInternalBlockTypeId() == id;
-                } else {
-                    id = extent.getBlock(vector).getInternalBlockTypeId();
-                    return true;
-                }
-            }
-        }
+        int typeId = extent.getBlock(vector).getInternalBlockTypeId();
+        int result = id.compareAndExchange(SENTINEL, typeId);
+        return result == typeId || result == SENTINEL;
     }
 
     @Override
@@ -37,7 +30,7 @@
 
     @Override
     public void reset() {
-        this.id = -1;
+        this.id.set(SENTINEL);
     }
 
     @Override

```

worldedit-core/src/main/java/com/fastasyncworldedit/core/function/mask/IdMask.java
dordsor21: Unsure

dordsor21:
Ah yeah, I was forgetting there was more than just get, and set in AtomicInt

SirYwell:
> Ah yeah, I was forgetting there was more than just get, and set in AtomicInteger, although, having it as an int check rather than requiring synchronisation into the future is a lot faster if a little convoluted to start with

I thought about the overhead as well, but from my measurements, the AtomicInteger solution seems to be faster than synchronization.

dordsor21:
We only synchronise once though with the current method

SirYwell:
The difference of ~40% I measured might also come from the usage of `Integer`, but I doubt that the compareAndExchange is noticable slower in any case here

dordsor21:
Mmm yeah that's probably true

SirYwell (approved):

NotMyFault (approved):

PR #2364, opened 2023-07-17
feat: allow placement of minecraft ConfiguredFeatures

### LINKED ISSUE #2343: Generation of Minecraft features and structures

Hey FAWE Team,

I'm a massive fan of your work. The tools the team/community has made turned the Minecraft experience into an art studio. FAWE / WorldEdit has truly redefined Minecraft.

Having said all that, I'm here to toss you an idea: Amethyst Cluster commands and brushes. I've found myself often spending more time than I'd like placing Amethyst Clusters around my builds by hand or having to create seriously complex masks which feels redundant since we have commands like //forest & //brush tree. So, I thought "Why not add a way using FAWE?"

1. Amethyst Cluster Generation Command: Let's have a command that generates Amethyst Clusters in a designated area. It could work like the existing //generate or forest commands but is customized for Amethyst Clusters. For example, //amethyst_cluster <density> where <density> defines the spread of clusters in the selected region and may be an optional argument for setting the amethyst stages as blocks.

2. Amethyst Cluster Brush: A brush that paints Amethyst Clusters just like we paint with the other brushes like the //brush tree would be a godsend. It would be awesome to control the density as well as the stage of Budding Amethyst. I'm thinking something along the lines of /brush amethyst_cluster <radius> <density> or /brush budding_amethyst <radius> [density] [stage].

These new additions would streamline the process of creating vibrant, shiny Amethyst environments in our builds. No more block-by-block placements, let's make the Amethyst Clusters rain!

I would like to also request maybe the //ores commands could be updated to support generating these an an optional flag?

Thanks for considering my ideas. I believe it would these would fantastic additions to the FAWE toolkit. Cheers to you all for the great work, have a wonderful weekend!

Best,
StealthyX 

I've attempted doing a few complex masks but the process is complex using deformed hollow spheres and replacing blocks and then doing more mask transforms in order to have the Amethyst buds in the Amethyst build.

### Anything else?

Not really, the "What feature do you want to see added" part answered enough and was pretty clear!

dordsor21:
- closes #2343
 - scope to add further features


SirYwell:
This only works if the EditSession is backed by an actual World, I think this should be outlined in the Javadocs.

dordsor21:
noticed upstream has this functionality, will change to match them and cherry-pick where appropriate

worldedit-bukkit/adapters/adapter-1_17_1/src/main/java/com/sk89q/worldedit/bukkit/adapter/impl/fawe/v1_17_R1_2/PaperweightFaweAdapter.java
```diff
@@ -555,6 +571,164 @@ protected ServerLevel getServerLevel(final World world) {
         return ((CraftWorld) world).getHandle();
     }
 
+    @Override
```
SirYwell, line 574: There is some similarity to the generateTree code here. Can we move it to the FaweAdapter too? I also think the Map<BlockPos, CraftBlockState> can be simplified to a list similar to how it's done in generateTree now.

worldedit-bukkit/adapters/adapter-1_20_2/src/main/java/com/sk89q/worldedit/bukkit/adapter/ext/fawe/v1_20_R2/PaperweightServerLevelDelegateProxy.java
```diff
@@ -0,0 +1,121 @@
...
+import net.minecraft.world.level.WorldGenLevel;
+import net.minecraft.world.level.block.entity.BlockEntity;
+import net.minecraft.world.level.block.state.BlockState;
+import org.jetbrains.annotations.Nullable;
+
+import java.lang.reflect.InvocationHandler;
+import java.lang.reflect.Method;
+import java.lang.reflect.Proxy;
+
+public class PaperweightServerLevelDelegateProxy implements InvocationHandler {
```
SirYwell, line 38: This class seems to be unused in that version. Did you miss upstream changes in the PaperweightAdapter? Generally, do we use that code (the changes in this PR to files in `ext/`) ourselves?

SirYwell (changes requested):
As there's a mix of structures and features here, what exactly do we add what WE doesn't add? `StructureType` is in the WE source here, but the commands and the tools for it are in the FAWE source, that's a bit confusing to me right now.

worldedit-bukkit/adapters/adapter-1_17_1/src/main/java/com/sk89q/worldedit/bukkit/adapter/impl/fawe/v1_17_R1_2/PaperweightFaweAdapter.java
dordsor21: I was going to look at doing so yeah. It would be more complicated though as the method changes per version a couple of times

worldedit-bukkit/adapters/adapter-1_20_2/src/main/java/com/sk89q/worldedit/bukkit/adapter/ext/fawe/v1_20_R2/PaperweightServerLevelDelegateProxy.java
dordsor21: Yeah I think there's some cleanup required to marry the implementation across each of the adapter versions

dordsor21:
> As there's a mix of structures and features here, what exactly do we add what WE doesn't add? `StructureType` is in the WE source here, but the commands and the tools for it are in the FAWE source, that's a bit confusing to me right now.

The various Tool/Factory classes are in FAWE because they are added by FAWE. The actual command methods are in com.sk89q because all commands are there

worldedit-bukkit/adapters/adapter-1_17_1/src/main/java/com/sk89q/worldedit/bukkit/adapter/impl/fawe/v1_17_R1_2/PaperweightFaweAdapter.java
dordsor21: Yeah I don't think that's very viable. We're not making any meaningful amount of abstraction as almost every line has a NMS class

worldedit-bukkit/adapters/adapter-1_20_2/src/main/java/com/sk89q/worldedit/bukkit/adapter/ext/fawe/v1_20_R2/PaperweightServerLevelDelegateProxy.java
dordsor21: Also, no, we don't use the changes in the ext classes, however, I think it's useful to have the WE code there for reference

NotMyFault:
Please rebase for 1.20.3/4

worldedit-core/src/main/java/com/fastasyncworldedit/core/command/tool/FeaturePlacer.java
```diff
@@ -0,0 +1,77 @@
...
+            Location clicked,
+            @Nullable Direction face
+    ) {
+
+        try (EditSession editSession = session.createEditSession(player)) {
+            try {
+                boolean successful = false;
+
+                final BlockVector3 pos = clicked.toVector().add().toBlockPoint();
+                for (int i = 0; i < 10; i++) {
```
SirYwell, line 57: What's the repetition for?

worldedit-bukkit/adapters/adapter-1_18_2/src/main/java/com/sk89q/worldedit/bukkit/adapter/impl/fawe/v1_18_R2/PaperweightFaweAdapter.java
```diff
@@ -547,6 +562,163 @@ protected ServerLevel getServerLevel(final World world) {
         return ((CraftWorld) world).getHandle();
     }
 
+    @Override
+    public boolean generateFeature(ConfiguredFeatureType feature, World world, EditSession editSession, BlockVector3 pt) {
```
SirYwell, line 566: Could this be moved into FaweAdapter similar to generateTree?

SirYwell (changes requested):

worldedit-bukkit/adapters/adapter-1_18_2/src/main/java/com/sk89q/worldedit/bukkit/adapter/impl/fawe/v1_18_R2/PaperweightFaweAdapter.java
dordsor21: See https://github.com/IntellectualSites/FastAsyncWorldEdit/pull/2364#discussion_r1403098520

worldedit-core/src/main/java/com/fastasyncworldedit/core/command/tool/FeaturePlacer.java
dordsor21: I believe it exists/existed somewhere in Minecaaft or upstream? There is often a random aspect to a feature or structure (e.g. tree height) that means it might fail to be constructed occasionally. This is an arbitrary way to try a few times basically

worldedit-bukkit/adapters/adapter-1_18_2/src/main/java/com/sk89q/worldedit/bukkit/adapter/impl/fawe/v1_18_R2/PaperweightFaweAdapter.java
SirYwell: *Oh*.
I guess it's fine then, maybe I'll try to extract parts of it when working on Folia stuff, but it shouldn't be a blocker here.

dordsor21: It would just leave a method in FaweAdapter with about 5 lines in, and then the block placement separate really

SirYwell:
For some reason there is a folder with a `.` instead of nested folders for the 1.20.4 adapter
dot

SirYwell:
From a quick test, it looks like e.g. `//structure bastion_remnant` throws an exception. I don't know if we can *fix* it, but I think it should at least be handled in a way that doesn't spam the console making it look like FAWE is doing something wrong.

Similarly, features and structures often cannot be placed with messages "Failed to generated structure. Is it a valid spot for it?" and "This feature cannot go here. Ensure the area meets the requirements." In fact, I wasn't able to spawn any feature for now. Can we somehow lift those restrictions? Otherwise, this is very limiting in its use.

dordsor21:
> From a quick test, it looks like e.g. `//structure bastion_remnant` throws an exception. I don't know if we can _fix_ it, but I think it should at least be handled in a way that doesn't spam the console making it look like FAWE is doing something wrong.

Yes probably worth doing that

> Similarly, features and structures often cannot be placed with messages "Failed to generated structure. Is it a valid spot for it?" and "This feature cannot go here. Ensure the area meets the requirements." In fact, I wasn't able to spawn any feature for now. Can we somehow lift those restrictions? Otherwise, this is very limiting in its use.

I tried, but no. Minecraft generation code is complete spaghetti and it would require an incredible amount of reflection and likely need us to perform changes to the byte code. The command in upstream WE has the same limitations

SirYwell:
A few problems I noticed:
- e.g. `//featureplacer azalea_tree` only works when clicking on a block that can get replaced, otherwise `(FAWE) worldedit.tool.feature.failed` is printed in the chat
- `//structureplacer` defaults to none, but the tool still gets equipped and does nothing
- e.g. `//structureplacer igloo`: it seems like the structure is always placed in the same spot in every chunk, and if a chunk already has such a structure, there's no feedback.

dordsor21:
For a lot of these placement issues we would potentially need to make very specific (potentially sweeping) overrides with more Unsafe forced instantiations. I'm quite sure that upstream would have the same issues - the Minecraft generation code for structures is incredibly cursed and almost every structure type would need its own implementation of fixes

SirYwell:
For the features (e.g. trees) it might be enough to use the block above (or depending on which side the block was clicked) for non-destructible blocks. For structures, I think we should at least try to give the user some kind of feedback if no structure was placed. Not sure if that's possible.

dordsor21:
I have added the face to specific cases (most but not all, it would be even worse to add it to all that none imo). Fixed the feedback issue.

Also, for igloos etc. it's exactly the same as if the player did //set stone on the same area twice - something "happens" but we're not going to tell them nothing was changed in practice

SirYwell:
I still can't seem to place trees this way. Not sure if they just have weird requirements. For the face stuff, would it be easier to just try to use the face if a previous try failed?

dordsor21:
> I still can't seem to place trees this way. Not sure if they just have weird requirements. For the face stuff, would it be easier to just try to use the face if a previous try failed?

We don't necessarily know why it's failed, e.g. sizes of features/structures are usually chosen before checking whether they can spawn, and thus we could end up trying to generate something we don't want to add the face to, with the face added. This is realistically the only way to do it that won't break/cause issues

SirYwell (approved):

dordsor21:
2 years 😎

worldedit-bukkit/adapters/adapter-1_17_1/src/main/java/com/sk89q/worldedit/bukkit/adapter/impl/fawe/v1_17_R1_2/PaperweightFaweAdapter.java
SirYwell: After looking at this again, maybe we can at least use the `getServerLevel`, `preCaptureStates`, `getCapturedBlockStatesCopy`, `postCaptureBlockStates` methods here? And also go from the Map to the List, as the Map just duplicates information but doesn't add any benefit.

dordsor21: Made the changes

SirYwell (approved):

NotMyFault (approved):

dordsor21:
🎉🎉

PR #2365, opened 2023-07-18
fix: correct the delegated methods in AbstractDelegateExtent

dordsor21:
- Not all methods should be overridden: we need to keep delegating-to-setblock methods in parent classes
 - Supersedes #2196


NotMyFault (approved):

SirYwell (approved):
Not happy with the fact the delegate extent doesn't delegate, but I guess this can't be solved without larger breaking changes.

PR #2385, opened 2023-07-29
Do not tab complete if command not owned by FAWE

### LINKED ISSUE #1912: Support tab completions in FAWE commands owned/overwritten by other plugins

hey folks, i am on the denizen team. we've had a number of reports of the FAWE `/ex` command overwriting tab completions for denizen's `/ex` command for a long time now. the same thing happened with WE's `/ex` command back in 2019, and I reached out and a fix was promptly made on their end. they were sending out tab completions even when they did not own the command. it seems FAWE is doing the same.

the tab completions for denizen's `/ex` command are almost essential for the plugin, it is quite disruptive for those to be unavailable. can we get a patch for this made on FAWE's end as well?

what we are currently recommending is for people to use WE until this is patched if they really need tab completions

### Anything else?

a link to the WE patch: https://github.com/EngineHub/WorldEdit/commit/d81fd99340607f408324cd6baa1ce4daaff12127

a quick search over FAWE's codebase, this is likely the same culprit and a similar fix: https://github.com/IntellectualSites/FastAsyncWorldEdit/blob/main/worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/WorldEditPlugin.java#L654

Boothin:
## Overview
Stops FAWE from trying to fill in tab completes for commands it doesn't own

Fixes #1912

## Description
Adds a check if the command owner for the command is FAWE

```[tasklist]
### Submitter Checklist
```

### DIFF
--- a/worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/WorldEditPlugin.java
+++ b/worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/WorldEditPlugin.java
@@ -672,6 +672,13 @@ public class WorldEditPlugin extends JavaPlugin {
             String label = buffer.substring(0, firstSpace);
             // Strip leading slash, if present.
             label = label.startsWith("/") ? label.substring(1) : label;
+
+            // If command not owned by FAWE, do not tab complete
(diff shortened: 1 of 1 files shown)


Boothin:
Oops made a mistake with the PR

PR #2400, opened 2023-08-15
Replace instanceof AliasedParser check for .getMatchedAliases() Try-Catch

Zeranny:
## Overview
The instanceof AliasedParser check frustrates attempts at Masks/Patterns which are compatible with both FAWE and WE.

I do hope I have missed an alternate (and maybe obvious?) way to implement Parsers without using the FAWE RichParser/Aliased Parser, but from what I can tell this singular check is a blocker.

The current code causes the parseFromInput() part of custom parsers to not run, which is an issue that has been flagged a couple times in the Discord support channel.

## Description
This change will still only match parsers which contain a getMatchedAliases() method so everything that currently passes the check will continue to do so.

By swapping the instanceof check for this, we can allow WE compatible parsers with their own Rich/argument parsing implementation to function when FAWE is present on a server.

This doesn't change functionality of other existing FAWE parsers.

```[tasklist]
### Submitter Checklist
```

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/internal/registry/AbstractFactory.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/internal/registry/AbstractFactory.java
@@ -121,10 +121,11 @@ public abstract class AbstractFactory<E> {
      */
     public boolean containsAlias(String alias) {
         return parsers.stream().anyMatch(p -> {
-            if (!(p instanceof AliasedParser)) {
+            try {
+                return ((AliasedParser) p).getMatchedAliases().contains(alias);
+            } catch (Exception e) {
                 return false;
             }
-            return ((AliasedParser) p).getMatchedAliases().contains(alias);
         });
     }
 


SirYwell:
I don't understand how this solves the issue. Could you elaborate your use case and how this change impacts the semantics there?

Zeranny:
> I don't understand how this solves the issue. Could you elaborate your use case and how this change impacts the semantics there?

Hi SirYwell, what I'm going for is a Pattern or Mask parser which will work regardless of FAWE or WE being present on the server.
To achieve this for WE I have a copy of RichParser within my tool, which works fine and as expected. The issue is when I instead load FAWE, for my parser to be used it must pass the "instanceof AliasedParser" check.

I can work around this by having a FAWE class extend my parser & implement FAWE's AliasedParser and register that instead when FAWE is loaded. 
However this change removes that requirement by changing the check from an instanceof condition to a try-catch around the existing .contains(alias) check.

As my parsers have the getMatchedAliases() method, they will pass this check without having to implement FAWE's AliasedParser and will work regardless of FAWE or WE being loaded.


I hope the use case is a bit clearer but, as I noted in my original PR note, I accept I may be missing some better alternative here.
In any case I appreciate your time looking at my PR 😃

SirYwell:
Calling your method will only work if you implement the AliasedParser interface, otherwise the code will throw a ClassCastException. Can't you just implement the interface regardless of whether WE or FAWE is used? Afaik this should not break as long as nothing depends on functionality from that interface.

Zeranny:
> Calling your method will only work if you implement the AliasedParser interface, otherwise the code will throw a ClassCastException. Can't you just implement the interface regardless of whether WE or FAWE is used? Afaik this should not break as long as nothing depends on functionality from that interface.

If I implement FAWE's AliasedParser I get a NoClassDefFoundError upon loading due to the import. 
I may be mistunderstanding what you mean though.

SirYwell:
You are right, I was misremembering things there. I think there isn't a proper solution that doesn't require special casing for FAWE then.

PR #2420, opened 2023-09-06
Unify the limit permission

RedstoneFuture:
## Overview

I have modified the limit permission so that the permission check is more accurate and the wording is more correct. This also makes it easier to assign permissions and disable the limit bypass for admins (especially the block blacklist of WE and FAWE).

## Description

- We down't need the `fawe.limit.*` permission as default permission if we have already checks for `fawe.limit.unlimited`.
- It's enough if only the granted permission is checked.

```[tasklist]
### Submitter Checklist
```

### DIFF
--- a/worldedit-bukkit/src/main/resources/plugin.yml
+++ b/worldedit-bukkit/src/main/resources/plugin.yml
@@ -21,7 +21,7 @@ permissions:
     default: op
     children:
       fawe.bypass.regions: true
-      fawe.limit.*: true
+      fawe.limit.unlimited: true
   fawe.tips:
     default: op
   fawe.admin:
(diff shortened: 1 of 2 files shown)


NotMyFault:
That feels somewhat of breaking, but we have no metrics about the usage either.

Maybe we can deprecate it for removal in fawe 3 and point the .* permission in the plugin.yml to .unlimited

Thoughts?

RedstoneFuture:
_"and point the .* permission in the plugin.yml to .unlimited"_

Do you mean:
```
  fawe.limit.*:
    default: false
    children:
      fawe.limit.unlimited: true
```
Is that allowed? `fawe.limit.*` actually means that only all manually created limits of the FAWE config are merged. "unlimit" goes beyond that.

dordsor21:
This shouldn't _really_ be breaking? It's effectively a fix if a user has disabled `*` wildcard permission nodes, as `fawe.bypass` is supposed to not have a limit assigned

dordsor21 (approved):

PR #2465, opened 2023-10-14
Fix "cleared" ellipsoid/cylinder regions still passing isDefined check

Zeranny:
## Overview
When you clear (`//sel`) a region it should effectively remove your selection. For all selection types except for Ellipsoid/Sphere & Cylinder this is true, but for these two types it simply sets their radii to 0 and center to 0,0,0 whilst leaving them in a state to pass their `isDefined()` checks.

For the Ellipsoid regions, you can still operate on the cleared region, so `//set 1` will set a 1x1x1 cube of stone at 0,0,0

## Description
This fix sets the boolean used to check if the region selection is defined to `false` in the `clear()` method so the user will correctly receieve the "Make a region selection first" error should they try to operate on the region after clearing it.

```[tasklist]
### Submitter Checklist
```

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/regions/selector/CylinderRegionSelector.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/regions/selector/CylinderRegionSelector.java
@@ -254,6 +254,8 @@ public class CylinderRegionSelector implements RegionSelector, CUIRegion {
     @Override
     public void clear() {
         region = new CylinderRegion(region.getWorld());
+        selectedCenter = false;
+        selectedRadius = false;
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


dordsor21:
This looks like it would also be an issue in upstream (WorldEdit itself), and so the PR would likely be better raised there

Zeranny:
> This looks like it would also be an issue in upstream (WorldEdit itself), and so the PR would likely be better raised there

Good shout, will see if I can get it accepted upstream

Zeranny:
https://github.com/EngineHub/WorldEdit/pull/2412

dordsor21 (approved):
I'll go ahead and merge as this was merged in upstream

dordsor21 (approved):
I'll go ahead and merge as this was merged in upstream

PR #2475, opened 2023-10-31
Support -n (select only) in //place

SirYwell:
## Description

I noticed that `//paste -n` was not working as expected. The reason is the fast path that delegates to the  `//place` implementation, which didn't support the `-n` flag before. By adding support for it there, we can simply fix that bug (and make the supported flags more consistent).

```[tasklist]
### Submitter Checklist
```


NotMyFault (approved):
Possibly needed in upstream too?

SirYwell:
Upstream doesn't have the `//place` command

PR #2478, opened 2023-11-01
Use FastSchematicReader by default

SirYwell:
## Description

`HashMultiMap` has no guarantees about the order of the values for a key. In our case, this leads to SPONGE_SCHEMATIC being first when iterating in `ClipboardFormats.findByExtension(...)`. We want FAST to come before, therefore we should use a `MultiMap` where values are in the order they are registered. This matches the `registeredFormats` list used by `findByFile`.

The change to the command is needed to allow actually enforcing a different schematic format.

```[tasklist]
### Submitter Checklist
```

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/command/SchematicCommands.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/command/SchematicCommands.java
@@ -313,9 +313,9 @@ public class SchematicCommands {
             Actor actor, LocalSession session,
             @Arg(desc = "File name.")
                     String filename,
-            @Arg(desc = "Format name.", def = "fast")
-                    String formatName,
             //FAWE start - random rotation
+            @Arg(desc = "Format name.", def = "")
+                    String formatName,
             @Switch(name = 'r', desc = "Apply random rotation to the clipboard")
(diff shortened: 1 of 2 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/command/SchematicCommands.java
```diff
@@ -313,9 +313,9 @@ public void load(
             Actor actor, LocalSession session,
             @Arg(desc = "File name.")
                     String filename,
-            @Arg(desc = "Format name.", def = "fast")
-                    String formatName,
             //FAWE start - random rotation
+            @Arg(desc = "Format name.", def = "")
+                    String formatName,
```
dordsor21, line 318: Why this change?

SirYwell: Otherwise we don't know whether the user wrote `//schem load <name>` or `//schem load <name> fast`, but those cases need to be handled separately, otherwise it's impossible to to use e.g. `//schem load <name> slow`.

NotMyFault (approved):

dordsor21 (approved):

PR #2482, opened 2023-11-02
Reuse generateTree code across versions

SirYwell:
## Description

The `generateTree` code is pretty stable, and we currently just copy it from one version to another. We can introduce a common superclass to deal with the version independent parts of the code and only implement the relevant parts in the versioned adapters.

It would be possible to replace more code using MethodHandles, but I don't think this makes things clearer.

There are also more things in those classes that could potentially deduplicated in a similar manner, but I think this is already hard enough to review properly, and the benefit in other methods seems smaller.

```[tasklist]
### Submitter Checklist
```


dordsor21:
Presumably this is the only method that can be extracted this way?

SirYwell:
I didn't look at every method in detail, but many are pretty compact already. It might make sense to look into that in future, but I didn't want to make reviewing more difficult than that.

SirYwell:
Moving back to draft, I also want to make future Folia integrations easier.

dordsor21 (approved):

NotMyFault (approved):

PR #2510, opened 2023-12-03
fix: set biome biome on clear

### LINKED ISSUE #2509: Biome does not get reset with `/p delete`
### Server Implementation

Paper

### Server Version

1.19.4

### Describe the bug

When the user is deleting their plot most things are getting deleted and cleared, but the biome still persists.

This is quite annoying for people that are just claiming a new plot and then this plot has a biome like the `nether_wastes` or the `snowy_peaks`.

### To Reproduce

1. Claim a plot
2. Change the biome with `/p setbiome <biome>` and if needed rejoin
3. Delete the plot

### Expected behaviour

The biome should revert to the default biome in the world so that people that claim the plot again have the default variant of the plot and not some changes that were made previously.

### Screenshots / Videos

**The default config of the plot world**
> 
**A claimed plot with the changed biome**
> 
**The deleted plot (not claimed anymore)**
> 

### Error log (if applicable)

_No response_

### Plot Debugpaste

https://athion.net/ISPaster/paste/view/61e975b53acf4559a613aca207373a6f

### PlotSquared Version

PlotSquared Version 7.0.0-SNAPSHOT

### Checklist

- [X] I have included a Plot debugpaste.
- [X] I am using the newest build from https://www.spigotmc.org/resources/77506/ and the issue still persists.

### Anything else?

_No response_

dordsor21:
- fixes #2509

### DIFF
--- a/worldedit-bukkit/src/main/java/com/fastasyncworldedit/bukkit/regions/plotsquared/FaweDelegateRegionManager.java
+++ b/worldedit-bukkit/src/main/java/com/fastasyncworldedit/bukkit/regions/plotsquared/FaweDelegateRegionManager.java
@@ -160,6 +160,8 @@ public class FaweDelegateRegionManager {
                         );
                         editSession.setBlocks(onTop, air);
                     }
+
+                    new CuboidRegion(pos1, pos2).forEach(bv3 -> editSession.setBiome(bv3, biome));
                 }
 
                 if (hybridPlotWorld.PLOT_SCHEMATIC) {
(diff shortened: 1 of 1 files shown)


worldedit-bukkit/src/main/java/com/fastasyncworldedit/bukkit/regions/plotsquared/FaweDelegateRegionManager.java
```diff
@@ -160,6 +160,8 @@ public boolean handleClear(
                         );
                         editSession.setBlocks(onTop, air);
                     }
+
+                    new CuboidRegion(pos1, pos2).forEach(bv3 -> editSession.setBiome(bv3, biome));
```
TheMeinerLP, line 164: Is there no better way to set biomes of the plot ?

TheMeinerLP (approved):

PR #2525, opened 2023-12-17
feat: add switch to allow removal of entities on paste

### LINKED ISSUE #2395: Overwrite entities when copy-pasting with entities enabled.

Brief: When using copy-paste with entities enabled (-e, or using the API), also remove the existing entities in the areas you are pasting to.

Using the API to copy-paste and allowing entities works fine, however when I try to paste that selection to an area that already includes entities, they do not get removed first, as you may expect, blocks on the other hand do get replaced. I'd like to see this also work for entities.

I tried to apply a workaround solution but the issue lies with entities being loaded differently from chunks, at least if I've understood this correctly. 

If this is something that is unrealistic to add, or unwanted then feel free to deny this request.

See the example code below of what I've been using to copy-paste.

```java
        //Get the worlds in worldEdit format
        com.sk89q.worldedit.world.World copyWorld = new BukkitWorld(copy);
        com.sk89q.worldedit.world.World pasteWorld = new BukkitWorld(paste);

        Polygonal2DRegion copyRegion = new Polygonal2DRegion(copyWorld, copyVector, MIN_Y, MAX_Y - 1);
        Polygonal2DRegion pasteRegion = new Polygonal2DRegion(pasteWorld, pasteVector, MIN_Y, MAX_Y - 1);
        BlockArrayClipboard clipboard = new BlockArrayClipboard(copyRegion);

        try (EditSession editSession = WorldEdit.getInstance().newEditSessionBuilder()
                .world(copyWorld).fastMode(true).checkMemory(false).limitUnlimited().changeSetNull().build()) {
            ForwardExtentCopy forwardExtentCopy = new ForwardExtentCopy(
                    editSession, copyRegion, clipboard, copyRegion.getMinimumPoint()
            );
            forwardExtentCopy.setCopyingBiomes(true);
            forwardExtentCopy.setCopyingEntities(true);
            // configure here
            Operations.complete(forwardExtentCopy);
        } catch (WorldEditException e) {
            e.printStackTrace();
            return false;
        }

        try (EditSession editSession = WorldEdit.getInstance().newEditSessionBuilder()
                .world(pasteWorld).fastMode(true).checkMemory(false).limitUnlimited().changeSetNull().build()) {

            Operation operation = new ClipboardHolder(clipboard)
                    .createPaste(editSession)
                    .to(pasteRegion.getMinimumPoint())
                    .copyBiomes(true)
                    .copyEntities(true)
                    // configure here
                    .build();
            Operations.complete(operation);
            editSession.flushQueue();
        } catch (WorldEditException e) {
            e.printStackTrace();
            return false;
        }
```

For my personal use I have applied a more dramatic workaround of deleting all entities to prevent people from being able to leave entities in the area of a region after they've deleted them. Else it could be abused by the playerbase.

This is obviously not a workaround for most situations, but to make sure our servers are secure from abuse, this had to be done.

Example:
```java
    /**
     * Deletes all entities in the given world
     *
     * @param world the world
     */
    public static void deleteEntities(World world) {

        @NotNull List<Entity> entities = world.getEntities();
        final int[] count = {0};
        entities.forEach(entity -> {
            if (!entity.getType().equals(EntityType.PLAYER)) {
                count[0]++;
                entity.remove();
            }
        });
        LOGGER.info(String.format("Removed %d entities from world %s", count[0], world.getName()));
    }
```

### Anything else?

I have only tested this using the API, not with the ingame commands, but I'd assume it functions the same. If not then this suggestion can be regarded as an API suggestion only.

If there is already a way to do this, then I'd be happy to hear, but I was unable to find it.

dordsor21:
- closes #2395

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/command/ClipboardCommands.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/command/ClipboardCommands.java
@@ -47,6 +47,7 @@ import com.sk89q.worldedit.command.util.CommandPermissionsConditionGenerator;
 import com.sk89q.worldedit.command.util.Logging;
 import com.sk89q.worldedit.command.util.annotation.Confirm;
 import com.sk89q.worldedit.command.util.annotation.Preload;
+import com.sk89q.worldedit.entity.Entity;
 import com.sk89q.worldedit.extension.platform.Actor;
 import com.sk89q.worldedit.extent.clipboard.BlockArrayClipboard;
 import com.sk89q.worldedit.extent.clipboard.Clipboard;
@@ -70,6 +71,7 @@ import com.sk89q.worldedit.math.BlockVector3;
 import com.sk89q.worldedit.math.Vector3;
 import com.sk89q.worldedit.math.transform.AffineTransform;
 import com.sk89q.worldedit.math.transform.Transform;
+import com.sk89q.worldedit.regions.CuboidRegion;
 import com.sk89q.worldedit.regions.NullRegion;
 import com.sk89q.worldedit.regions.Region;
 import com.sk89q.worldedit.regions.RegionIntersection;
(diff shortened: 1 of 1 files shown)


SirYwell (approved):

worldedit-core/src/main/java/com/sk89q/worldedit/command/ClipboardCommands.java
```diff
@@ -513,14 +522,19 @@ public void paste(
                     boolean pasteBiomes,
             @ArgFlag(name = 'm', desc = "Only paste blocks matching this mask")
             @ClipboardMask
-                    Mask sourceMask
+                    Mask sourceMask,
+            //FAWE start - entity removal
+            @Switch(name = 'x', desc = "Remove existing entities in the affect region")
```
PierreSchwang, line 527: Same as above

PierreSchwang (approved):

PR #2540, opened 2024-01-06
Drop support for 1.17.1 and 1.18.2

NotMyFault:
As proven evidently, 1.17.1 is used by ~1.9% and 1.18.2 by 4.7% of our user base, declaring a stagnating or no interest in these versions, it's only logical to drop support for them.
Given the fast-pacing release economy of Minecraft, it's not viable to upkeep versions released more than ~3 years ago only a small fraction of our user base sticks to.

Screenshot 2024-01-06 at 11 04 13


TheMeinerLP (approved):

PierreSchwang (approved):
🚮

SirYwell:
I think 4.7% is still a fair amount. Is there any technical reason to drop it?

NotMyFault:
> I think 4.7% is still a fair amount.

Providing compatibility for a stagnating minority has never been a goal of this project. 1.8.8 is used by 8.3%, which is an even more "fair amount" following your logic.

SirYwell:
> Providing compatibility for a stagnating minority has never been a goal of this project. 1.8.8 is used by 8.3%, which is an even more "fair amount" following your logic.

Not supporting 1.8.8 has a very strong technical reason as we'd basically need to maintain two plugins. That's why I'm asking. Not supporting a version that has basically zero overhead to maintain (besides probably copy-pasting some code) does not seem reasonable to me.

dordsor21:
Yeah I

> > Providing compatibility for a stagnating minority has never been a goal of this project. 1.8.8 is used by 8.3%, which is an even more "fair amount" following your logic.
> 
> Not supporting 1.8.8 has a very strong technical reason as we'd basically need to maintain two plugins. That's why I'm asking. Not supporting a version that has basically zero overhead to maintain (besides probably copy-pasting some code) does not seem reasonable to me.

I agree with this tbh, there's no real need to remove support yet

NotMyFault:
In this case, I'll withdraw my attention from anything older than 1.19.4.

PR #2560, opened 2024-01-31
Change -h Tool Command Flag to Always Impact Secondary Brush

Zeranny:
## Overview
I was asked recently what the `-h` flag did in various tool util commands as the existing description of "Whether the offhand should be considered or not" is unclear to others and myself.
Upon using it, and from the feedback of others trying to use it too, it felt inconsistent in when it would change the primary or secondary brush.

From reading the code it appears the intent is for tool util commands to change the last used brush when a primary & secondary are set, with the flag inverting which one it impacts.

I think this isn't very intuitive and believe it would make more sense to just have the flag set to change the secondary brush, otherwise the primary.

## Description
This changes the `-h` flag to always change the value for the secondary brush, if one is set. If no flag is used, or no secondary brush is set, then it will just change the primary brush.

I have changed the help page info to make this clear.

Additionally I have added the same flag functionality to the `/size` / `/br size` command.

Lastly, some minor tidying of the unused editSession import, and the inconsistent tool/brushTool/bt naming.

Hopefully this change makes sense, I'd be curious to hear if myself and those I spoke to are in the minority when it comes to thinking the existing method is unclear and hard to use.

```[tasklist]
### Submitter Checklist
```

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/command/ToolUtilCommands.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/command/ToolUtilCommands.java
@@ -27,7 +27,6 @@ import com.fastasyncworldedit.core.extent.ResettableExtent;
 import com.fastasyncworldedit.core.util.MathMan;
 import com.fastasyncworldedit.core.util.StringMan;
 import com.google.common.collect.Iterables;
-import com.sk89q.worldedit.EditSession;
 import com.sk89q.worldedit.LocalSession;
 import com.sk89q.worldedit.WorldEdit;
 import com.sk89q.worldedit.WorldEditException;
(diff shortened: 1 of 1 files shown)


SirYwell:
> This changes the `-h` flag to always change the value for the secondary brush, if one is set. If no flag is used, or no secondary brush is set, then it will just change the primary brush.

I think this should rather be a warning, otherwise the user might modify something they didn't want to modify. Would that make sense?

Zeranny:
> > This changes the `-h` flag to always change the value for the secondary brush, if one is set. If no flag is used, or no secondary brush is set, then it will just change the primary brush.
> 
> I think this should rather be a warning, otherwise the user might modify something they didn't want to modify. Would that make sense?

I think adding a message that states wether it was the primary or secondary that the mask, size, or whatever has been applied to could help, but this change instead requires the user to explicitely declare that they want to modify the secondary which removes any ambiguity.

Unless I misunderstood your concern?

SirYwell:
> I think adding a message that states wether it was the primary or secondary that the mask, size, or whatever has been applied to could help, but this change instead requires the user to explicitely declare that they want to modify the secondary which removes any ambiguity.
> 
> Unless I misunderstood your concern?

My concern is that explicitly wanting to change a secondary brush changes the primary brush. But I also see how only preventing that could be confusing. I guess a message clarifying which brush was changed would be enough then.

PR #2613, opened 2024-03-10
Allow factory suggestions to have parser context

Zeranny:
## Overview
Closely mirrors upstream changes in https://github.com/EngineHub/WorldEdit/pull/2236

"This is a refactor to allow access to a ParserContext within the suggestion methods. This is required for any suggestions within the factory parser that are player-specific"

## Description
This PR adds context as an option to the getSuggestions() method of parsers.
While no changes to existing Masks or Patterns are present, this change would allow for that to be done in future.

Originally I had gone through and added context to every parser in FAWE, allowing for the new `getSuggestions(String argumentInput, int index, ParserContext context)` method in RichParser to be abstract, replacing the existing method without context. 
However, that would be a breaking change for any addon plugin using rich parsing such as Arceon.

While it does work & prevents breaking existing parsers, I don't think my approach of 2x methods looks the best so I would appreciate any thoughts on that.
I have tested against existing FAWE, Arceon, and my own custom masks/patterns to double check this is backwards compatible.

It's worth mentioning that my reason for pursuing this change is a custom mask I implemented which would benefit from player-specific tab completion.

Additionally as a basic example, this is a mask that wants a Y value, which can suggest the user's Y level:

image

```[tasklist]
### Submitter Checklist
```

### DIFF
--- a/worldedit-core/src/main/java/com/fastasyncworldedit/core/extension/factory/parser/RichParser.java
+++ b/worldedit-core/src/main/java/com/fastasyncworldedit/core/extension/factory/parser/RichParser.java
@@ -127,6 +127,14 @@ public abstract class RichParser<E> extends InputParser<E> implements AliasedPar
         return Stream.empty();
     }
 
+    /**
+     * Returns a stream of suggestions for the argument at the given index.
+     *
+     * @param argumentInput the already provided input for the argument at the given index.
+     * @param index         the index of the argument to get suggestions for.
+     * @param context       the context which may optionally be provided by a parser.
+     * @return a stream of suggestions matching the given input for the argument at the given index.
+     */
     protected Stream<String> getSuggestions(String argumentInput, int index, ParserContext context) {
         return getSuggestions(argumentInput, index);
     }
--- a/worldedit-core/src/main/java/com/fastasyncworldedit/core/extension/factory/parser/RichParser.java
+++ b/worldedit-core/src/main/java/com/fastasyncworldedit/core/extension/factory/parser/RichParser.java
@@ -124,7 +124,7 @@ public abstract class RichParser<E> extends InputParser<E> implements AliasedPar
      * @return a stream of suggestions matching the given input for the argument at the given index.
      */
     protected Stream<String> getSuggestions(String argumentInput, int index) {
-        return Stream.empty();
+        return getSuggestions(argumentInput, index, new ParserContext());
     }
 
     /**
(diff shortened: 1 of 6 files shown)


worldedit-core/src/main/java/com/fastasyncworldedit/core/extension/factory/parser/RichParser.java
```diff
@@ -123,7 +123,21 @@ public E parseFromInput(String input, ParserContext context) throws InputParseEx
      * @param index         the index of the argument to get suggestions for.
      * @return a stream of suggestions matching the given input for the argument at the given index.
      */
-    protected abstract Stream<String> getSuggestions(String argumentInput, int index);
+    protected Stream<String> getSuggestions(String argumentInput, int index) {
+        return Stream.empty();
```
dordsor21, line 127: This should default to the new method with an empty parser context

Zeranny: > This should default to the new method with an empty parser context

I've swapped them round https://github.com/IntellectualSites/FastAsyncWorldEdit/pull/2613/commits/5c257f60ea9ec002e407b2a4a8a7c6ede6bee5e7

dordsor21: I meant that they should both default to each other so parsers can implement one of either method and it will still work. Similar is already done in extents https://github.com/IntellectualSites/FastAsyncWorldEdit/blob/main/worldedit-core/src/main/java/com/sk89q/worldedit/extent/InputExtent.java#L52-L59

Zeranny: Maybe a dumb question, but would there be any risk if someone did not overwrite one or both of them when using the API? Would the getSuggestions not just loop?

dordsor21: Yeah I suppose we're not necessarily going to always do suggestions? Return it to how it was but add a deprecation to it I think, as that method should never be called; overriding is fine but calling is not

Zeranny: Added the Deprecated tag & a note: https://github.com/IntellectualSites/FastAsyncWorldEdit/pull/2613/commits/2873359ec458a8b7d3fe8566fc4d01a826c66a55

dordsor21 (approved):

NotMyFault (approved):

NotMyFault:
Much appreciated!

PR #2614, opened 2024-03-11
fix: do not forcefully direct to x,y,z methods in AbstractDelegateExtent

dordsor21:
Probably shouldn't be doing this


SirYwell:
How does this affect extending extents? E.g. SlowExtent only overrides the method that takes ints, and I assume that might also apply for other extents.

dordsor21:
Input/Output extent have default methods that just redirect to the other method. This change just means we're not creating new BV3 objects if the delegate extent only overrides the BV3 method(s) - it shouldn't have any other impacts.

NotMyFault (approved):

PR #2615, opened 2024-03-11
refactor: apply final/abstract to various fawe extent classes

dordsor21:


dordsor21:
Thoughts?

dordsor21:
I'll fix this lol

worldedit-core/src/main/java/com/fastasyncworldedit/core/extent/TransformExtent.java
```diff
@@ -12,6 +12,12 @@
 import com.sk89q.worldedit.world.block.BlockState;
 import com.sk89q.worldedit.world.block.BlockStateHolder;
 
+/**
+ * Unused internal, will be removed
+ *
+ * @deprecated Will be removed - unused internally
+ */
+@Deprecated(forRemoval = true, since = "TODO")
```
SirYwell, line 20: Would it make sense to remove it in v3 and deprecated it in v2 already?

SirYwell:
You removed the comment from `BlockTransformExtent` but you didn't make it final, I assume that was unintentional?

dordsor21:
> You removed the comment from `BlockTransformExtent` but you didn't make it final, I assume that was unintentional?

Yeah I realised it was a WE class

SirYwell (approved):

NotMyFault (approved):
`final`ly

PR #2617, opened 2024-03-13
fix: set and clone Expression Environment after Expression cloning

### LINKED ISSUE #2616: Mask queries don't work anymore
### Server Implementation

Paper

### Server Version

1.20.4

### Describe the bug

`queryRel` Masks don't seem to work anymore. Whenever I try to use it, it doesn't work and spams the same NullPointerException `Cannot invoke "com.sk89q.worldedit.internal.expression.ExpressionEnvironment.getBlockTypeRel(double, double, double)" because "this.environment" is null`  in the console over and over again. 

- I tried it with different servers even with ones that only have FAWE installed
- I tried it with newest server versions and older server versions (1.19.4)
- I tried it with newest and older FAWE versions till half a year ago
=> No matter what I do I can always reproduce this error with any queryRel and any mask or gmask that should produce a result but fails.

I also looked at the code and compiled it with some debug messages. Whenever I type the queryRel command, new Function classes get created and the Environment variable in the Function class is set. But once I execute the command another Function class is created but this time no Environment is set which then obviously leads to a NullPointerException.

What should I do now? I really need this to work and I'm willing to help out to fix this problem. 

I talk about this line in the code but it will probably also fail for the other methods like `query` or `queryAbs`: https://github.com/IntellectualSites/FastAsyncWorldEdit/blob/main/worldedit-core/src/main/java/com/sk89q/worldedit/internal/expression/Functions.java#L425

### To Reproduce

1. Create a brush (e.g. `//brush sphere stone`)
2. Create a mask with queryRel (e.g. `//mask =queryRel(0,1,0,0,0)` - Mask all blocks that have air above them)
3. Click with the brush and see the errors in the console. Nothing gets replaced even though it should work.

### Expected behaviour

Normally the queryRel should work. In my Steps to Reproduce example it should replace all surface blocks with stone. 
It says that it worked but no blocks get placed and an error gets printed in the console.

### Screenshots / Videos

image

Debug Messages:
- FUNCTIONS CREATED message in Functions.java constructor
- ENVIRONMENT SET message in Functions.setEnvironment() function
- Environment check on command execution
image

### Error log (if applicable)

https://paste.gg/p/anonymous/a1cdb839af32405a98c6bcbd68ac0357

### Fawe Debugpaste

https://athion.net/ISPaster/paste/view/89e8261d1a60492f859535bd46e51942

### Fawe Version

2.9.1-SNAPSHOT

### Checklist

### Anything else?

_No response_

MineFact:
## Overview

Fixes #2616

## Description
This Pull Requests sets the WorldEditExpressionEnvironment after an Expression gets cloned to make sure that the Functions class remains its environment after cloning  to not run into a NullPointerException.
When the Expression is cloned, the WorldEditExpressionEnvironment gets cloned as well while making sure that the offset of the environment is not changing during the cloning process.

```[tasklist]
### Submitter Checklist
```

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/internal/expression/ExpressionEnvironment.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/internal/expression/ExpressionEnvironment.java
@@ -22,7 +22,7 @@ package com.sk89q.worldedit.internal.expression;
 /**
  * Represents a way to access blocks in a world. Has to accept non-rounded coordinates.
  */
-public interface ExpressionEnvironment {
+public interface ExpressionEnvironment extends Cloneable {
 
     int getBlockType(double x, double y, double z);
 
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/internal/expression/ExpressionEnvironment.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/internal/expression/ExpressionEnvironment.java
@@ -36,4 +36,5 @@ public interface ExpressionEnvironment {
 
     int getBlockDataRel(double x, double y, double z);
 
+    ExpressionEnvironment clone();
 }
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/internal/expression/ExpressionEnvironment.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/internal/expression/ExpressionEnvironment.java
@@ -36,5 +36,7 @@ public interface ExpressionEnvironment {
 
     int getBlockDataRel(double x, double y, double z);
 
+    // FAWE start
     ExpressionEnvironment clone();
+    // FAWE end
 }
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/regions/shape/WorldEditExpressionEnvironment.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/regions/shape/WorldEditExpressionEnvironment.java
@@ -94,10 +94,13 @@ public class WorldEditExpressionEnvironment implements ExpressionEnvironment {
     public Vector3 toWorldRel(double x, double y, double z) {
         return current.add(x, y, z);
     }
+
+    public WorldEditExpressionEnvironment clone() {
+        return new WorldEditExpressionEnvironment(extent, unit, zero2);
+    }
     //FAWe end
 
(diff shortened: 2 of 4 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/regions/shape/WorldEditExpressionEnvironment.java
```diff
@@ -28,6 +28,8 @@
 
 public class WorldEditExpressionEnvironment implements ExpressionEnvironment {
 
+    private final Vector3 UN_OFFSET = Vector3.at(0.5, 0.5, 0.5);
```
SirYwell, line 31: Maybe `BLOCK_CENTER_OFFSET` might be a better name?

worldedit-core/src/main/java/com/sk89q/worldedit/internal/expression/ExpressionEnvironment.java
```diff
@@ -36,4 +36,5 @@ public interface ExpressionEnvironment {
 
     int getBlockDataRel(double x, double y, double z);
 
+    ExpressionEnvironment clone();
```
SirYwell, line 39: Could you put `// FAWE` comments around this method, similar to how it is done in the implementation?

SirYwell (commented):
A few small suggestions

worldedit-core/src/main/java/com/sk89q/worldedit/regions/shape/WorldEditExpressionEnvironment.java
```diff
@@ -28,6 +28,8 @@
 
 public class WorldEditExpressionEnvironment implements ExpressionEnvironment {
 
+    private final Vector3 BLOCK_CENTER_OFFSET = Vector3.at(0.5, 0.5, 0.5);
```
PierreSchwang, line 31: should be made static, it's already uppercase, and we don't have to allocate a Vector for each instance of the environment (especially due to potential clones)

PierreSchwang:
I'd say ExpressionEnvironment should implement Clonable, if there is already the clone method

SirYwell (approved):
Looks good, thanks!

NotMyFault (approved):

TheMeinerLP (approved):

dordsor21 (approved):

PR #2619, opened 2024-03-14
Integrate WE Schematic Share system

SirYwell:
## Description

A WE 7.3.0 feature from 2021, it's time to pull some upstream changes.

```[tasklist]
### Submitter Checklist
```

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/command/SchematicCommands.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/command/SchematicCommands.java
@@ -554,6 +554,9 @@ public class SchematicCommands {
                       String schematicName,
                       @Arg(desc = "Format name.", def = "sponge")
                       String formatName) throws WorldEditException {
+        if (true) {
+            throw new UnsupportedOperationException("This feature is currently not implemented");
+        }
         if (worldEdit.getPlatformManager().queryCapability(Capability.GAME_HOOKS).getDataVersion() == -1) {
             actor.printError(TranslatableComponent.of("worldedit.schematic.unsupported-minecraft-version"));
             return;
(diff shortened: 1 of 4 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/command/SchematicCommands.java
```diff
@@ -984,7 +1057,32 @@ public Void call() throws Exception {
...
+        @Override
+        public URL call() throws Exception {
+            ByteArrayOutputStream baos = new ByteArrayOutputStream();
+            try {
+                writeToOutputStream(baos);
+            } catch (Exception e) {
+                throw new CommandException(TextComponent.of(e.getMessage()), e, ImmutableList.of());
+            }
+
+            EngineHubPaste pasteService = new EngineHubPaste();
```
dordsor21, line 1079: Is this something we're comfortable using in FAWE? We already have the schematic download portal ourselves

PierreSchwang: If I remember me's blog entry correctly, other plugins should be able to hook into the paste system and allow custom schematic paste / load services. If that's still the case, our portal should be accessed via that api - the EngineHub way could still be enabled (iirc, that was even a command argument - tho I'm not seeing anything like that in this PR), if the EH peeps don't have any issue with that. That's totally up to them, and we shouldn't utilize their service if they're not fine with that.

SirYwell: I don't have a strong opinion on that stuff, the main reason for this PR is that otherwise we have major merge conflicts for other stuff in 7.3 and in future.

dordsor21: Probably worth just double checking first really

SirYwell: I disabled the feature for now, so we can integrate the code and take care of details later

dordsor21 (approved):

PierreSchwang (approved):

PR #2621, opened 2024-03-15
feat: save region selector to session store

### LINKED ISSUE #2397: Make //sel extend persist through sessions and restarts

I almost exclusively use //sel extend for ease of use and peace of mind. It is quite frustrating having to type //sel extend everytime I relog or server restarts.

Please allow that selection method to persist through sessions and reboots!

A workaround is making a script that executes //sel extend for me everytime I log in, but a native feature would be much cleaner.

### Anything else?

_No response_

dordsor21:
- closes #2397

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/util/gson/RegionSelectorAdapter.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/util/gson/RegionSelectorAdapter.java
@@ -1,0 +1,33 @@
+package com.sk89q.worldedit.util.gson;
+
+import com.google.gson.JsonDeserializationContext;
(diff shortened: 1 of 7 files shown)


NotMyFault (approved):

worldedit-core/src/main/java/com/sk89q/worldedit/util/gson/RegionSelectorAdapter.java
```diff
@@ -0,0 +1,33 @@
+package com.sk89q.worldedit.util.gson;
```
SirYwell, line 1: Would it make sense to move this to a FAWE package?

SirYwell (approved):

PR #2627, opened 2024-03-16
Drop support for 1.17.1

NotMyFault:
With more issues occurring related to unavailable dependency methods on older versions we support, e.g. guava's `Builder.buildOrThrow()`, FAWE does no longer work on 1.17.1.

With [1.5%](https://bstats.org/plugin/bukkit/FastAsyncWorldEdit/1403#minecraftVersion) share, I propose to drop 1.17.1 without a warning period, given we're past this state already.
Screenshot 2024-03-16 at 20 37 20


NotMyFault:
De minimis, as it's de facto no longer functional.

PR #2626, opened 2024-03-16
Fix Linear pattern index incrementing above array length

Zeranny:
## Overview
When running async the `#linear` pattern can increment index above the length of the `patternsArray`.
This tends not to happen on small operations.

```
[14:49:33 ERROR]: [com.fastasyncworldedit.core.queue.implementation.ParallelQueueExtent] Catching
java.lang.ArrayIndexOutOfBoundsException: Index 9 out of bounds for length 3
        at com.fastasyncworldedit.core.function.pattern.LinearBlockPattern.apply(LinearBlockPattern.java:37) ~[FastAsyncWorldEdit-Bukkit-2.9.1-SNAPSHOT.jar:?]
        at com.sk89q.worldedit.function.pattern.Pattern.applyBlock(Pattern.java:59) ~[FastAsyncWorldEdit-Bukkit-2.9.1-SNAPSHOT.jar:?]
        at com.fastasyncworldedit.core.extent.filter.LinkedFilter.applyBlock(LinkedFilter.java:28) ~[FastAsyncWorldEdit-Bukkit-2.9.1-SNAPSHOT.jar:?]
        at com.fastasyncworldedit.core.extent.filter.block.CharFilterBlock.filter(CharFilterBlock.java:130) ~[FastAsyncWorldEdit-Bukkit-2.9.1-SNAPSHOT.jar:?]
        at com.sk89q.worldedit.regions.Region.filter(Region.java:367) ~[FastAsyncWorldEdit-Bukkit-2.9.1-SNAPSHOT.jar:?]
        at com.sk89q.worldedit.regions.Region.filter(Region.java:301) ~[FastAsyncWorldEdit-Bukkit-2.9.1-SNAPSHOT.jar:?]
        at com.sk89q.worldedit.regions.CuboidRegion.filter(CuboidRegion.java:762) ~[FastAsyncWorldEdit-Bukkit-2.9.1-SNAPSHOT.jar:?]
        at com.fastasyncworldedit.core.extent.filter.block.ChunkFilterBlock.filter(ChunkFilterBlock.java:88) ~[FastAsyncWorldEdit-Bukkit-2.9.1-SNAPSHOT.jar:?]
        at com.fastasyncworldedit.core.queue.implementation.chunk.ChunkHolder.filterBlocks(ChunkHolder.java:903) ~[FastAsyncWorldEdit-Bukkit-2.9.1-SNAPSHOT.jar:?]
[...]

PR #2633, opened 2024-03-17
Make usage of stateful patterns thread-safe

SirYwell:
## Description

Some patterns make use of `MutableBlockVector` or might delegate to patterns that might do so, therefore we need to fork it to prevent race conditions.

The example command `//set #!z[#linear2d[white_wool,2%light_gray_wool,4%gray_wool,8%black_wool]]` and how it looks:

Before:
baseline

With this change:
patched

```[tasklist]
### Submitter Checklist
```


dordsor21:
This wouldn't work if the patterns are nested within another pattern. We ought to create an AbstractDelegatePattern and override fork in there and ensure all patterns with nested patterns utilise this

SirYwell:
Good point, I'll look into it.

SirYwell:
I went for an interface `StatefulPattern`, there are more scenarios where patterns need forking and the ones delegating to other patterns are mostly not a good fit for an `AbstractDelegatePattern` (e.g. `RandomPattern`) and might also already extend other classes (e.g. `AbstractExtentPattern`, do we need to fork them?).

Let me know what you think.

SirYwell:
This might also be relevant for #2626, I'm just not sure what to expect from that pattern in a multi-threaded setting

dordsor21:
Are there definitely no further instances where pattern might not necessarily get forked here?

SirYwell:
> Are there definitely no further instances where pattern might not necessarily get forked here?

I'll have to go through the WE patterns again, it seems there are a few more

SirYwell:
Went through the impls again, should be all now hopefully.

dordsor21:
Is there much of a point in having the new interface? If a new pattern that would use it is created then either we remember the interface or we don't (same chances as overriding the fork method tbh)

SirYwell:
I thought about forking some of the patterns only if the inner patterns need to be forked, in which case it was somewhat useful as a marker interface, but I decided against the additional checking overhead, so I guess it can be removed again.

worldedit-core/src/main/java/com/fastasyncworldedit/core/function/pattern/RelativePattern.java
```diff
@@ -63,4 +63,11 @@ public void reset() {
         origin = null;
     }
 
+    @Override
+    public Pattern fork() {
+        RelativePattern forked = new RelativePattern(this.pattern.fork(), this.minY, this.maxY);
+        forked.origin = this.origin; // maintain origin for forks
```
dordsor21, line 69: Should probably look at having the origin be `toImmutable()`ised when it's set

SirYwell: That would otherwise already cause bugs I think? But yeah, that might make sense.

dordsor21: Possibly, but can't be too careful aha

dordsor21 (approved):

NotMyFault (approved):

PR #2645, opened 2024-03-22
Fix queryRel returning data value for block at absolute coordinates

### LINKED ISSUE #2643: queryRel masks with data values don't work anymore
### Server Implementation

Paper

### Server Version

1.20.4

### Describe the bug

`queryRel` Masks don't accept data values anymore. They perform any operation without any error, but as soon as the data value in a queryRel is not 0 or -1 it doesn't do anything anymore. This is quite annoying because this eliminates half of the blocks that you can work with. As soon as one queryRel has a data value the entire mask doesn't do anything anymore.

- I tried it with different servers even with ones that only have FAWE installed
- I tried it with newest server versions and older server versions (1.19.4)
- I tried it with newest and older FAWE versions till half a year ago
- I tried it with normal WorldEdit and there it works. Only FAWE has this problem.
=> No matter what I do I can always reproduce this error with any queryRel and any mask or gmask that should produce a result but fails.

### To Reproduce

1. Select a region that contains a block with a data value (e.g. yellow wool)
2. Create a mask with queryRel (e.g. //gmask =queryRel(0,0,0,35,5) - Mask all blocks that are yellow wool)
3. Replace them (e.g. //set stone)

### Expected behaviour

The queryBlock identifies blocks with data values. 

### Screenshots / Videos

https://github.com/IntellectualSites/FastAsyncWorldEdit/assets/66020920/bc7ce129-4614-4354-aa09-0f9178fa6757

### Error log (if applicable)

no errors

### Fawe Debugpaste

https://athion.net/ISPaster/paste/view/4861214b41a1481fafbb3a6b55360eab

### Fawe Version

2.9.1-SNAPSHOT

### Checklist

### Anything else?

_No response_

Zeranny:
## Overview
Fixes the incorrect method for relative data checks in expressions (queryRel)

Fixes #2643

## Description
Current version erroneously uses toWorld instead of toWorldRel, so `queryRel(0,0,0,t,d)` would always return the data of the block at `0,0,0` for `d`, where it should return the data value of whatever block is `0,0,0` offset.

```[tasklist]
### Submitter Checklist
```

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/regions/shape/WorldEditExpressionEnvironment.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/regions/shape/WorldEditExpressionEnvironment.java
@@ -83,6 +83,7 @@ public class WorldEditExpressionEnvironment implements ExpressionEnvironment {
     }
 
     @SuppressWarnings("deprecation")
+    @Override
     public int getBlockDataRel(double x, double y, double z) {
         return extent.getBlock(toWorldRel(x, y, z).toBlockPoint()).getBlockType().getLegacyCombinedId() & 0xF;
     }
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/regions/shape/WorldEditExpressionEnvironment.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/regions/shape/WorldEditExpressionEnvironment.java
@@ -83,9 +83,8 @@ public class WorldEditExpressionEnvironment implements ExpressionEnvironment {
     }
 
     @SuppressWarnings("deprecation")
-    @Override
     public int getBlockDataRel(double x, double y, double z) {
-        return extent.getBlock(toWorld(x, y, z)).getBlockType().getLegacyCombinedId() & 0xF;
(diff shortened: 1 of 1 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/regions/shape/WorldEditExpressionEnvironment.java
```diff
@@ -83,9 +83,8 @@ public int getBlockTypeRel(double x, double y, double z) {
     }
 
     @SuppressWarnings("deprecation")
-    @Override
```
SirYwell, line 86: The annotation should be kept

SirYwell (commented):
As a side note, we should probably also pull <https://github.com/EngineHub/WorldEdit/commit/b5b9c6ab8694b7d3c4dcb435c0342d6e83c0d014> as a follow-up?

worldedit-core/src/main/java/com/sk89q/worldedit/regions/shape/WorldEditExpressionEnvironment.java
Zeranny: Good spot, didn't mean to remove that

SirYwell (approved):

NotMyFault (approved):

PR #2356, opened 2023-07-15
feat: improved entity operations

dordsor21:
- Add new extent that does an action on chunk GET load allowing for "unloaded" entities to be edited
 - closes #1826

PR #2649, opened 2024-03-24
Use correct find class in ExtentTraverser

### LINKED ISSUE #2485: Offset mask does not work reliable with `//br s`
### Server Implementation

Paper

### Server Version

1.20.2

### Describe the bug

Using the offset mask with a sphere brush only seems to work around 50% of the time

### To Reproduce

In this example I'm using red tulips, but it also happens with other blocks
1. `//br s red_tulip 3`
2. `/mask >water`
3. `/gmask 0`

Using the brush (sometimes) leads to semi-spheres of roses:

### Expected behaviour

All roses should be flat

### Screenshots / Videos

_No response_

### Error log (if applicable)

None

### Fawe Debugpaste

https://athion.net/ISPaster/paste/view/f05eb4e1d79448d39852ce8b03c740cb

### Fawe Version

FastAsyncWorldEdit 2.8.2-SNAPSHOT-602;8e2691c

### Checklist

### Anything else?

_No response_

### LINKED ISSUE #2546: Gmasks and Masks not working together
### Server Implementation

Paper

### Server Version

1.20.4

### Describe the bug

When using "gmask 0" and "mask >gold_block" with a sphere brush, the brush doesn't follow the masks. 

https://discord.com/channels/268444645527126017/344128526435221505/1197635519312560218

### To Reproduce

1. "/br s diamond_block"
2. "/gmask 0"
3. "/mask >gold_block"
4. Brush ontop of a gold block pile. 

### Expected behaviour

It should create a top layer of diamond blocks, without cutting into the gold blocks. 

### Screenshots / Videos

weird
The left side is with the method explained, the right side is placed by hand and is the expected outcome. 

### Error log (if applicable)

_No response_

### Fawe Debugpaste

https://athion.net/ISPaster/paste/view/7aabd1467b9e43f9affede1aea657444

### Fawe Version

FAWE 12.8.5-SNAPSHOT (1.20.4)

### Checklist

### Anything else?

_No response_

SirYwell:
## Overview

Fixes #2485
Fixes #2546

## Description

We already discovered this in #2356, but it took me some debugging to figure out that the linked issues are also caused by this bug. There are two `find` methods, and previously, the bound type of the parameter wasn't strict enough to use the method that takes a `Class<U extends Extent>`, but instead the `find(Object)` method was used. This can be fixed by using a sharper bound.

This *can* theoretically cause compile time issues with existing code, but that will mean that the method is used wrongly most likely.

As a bit of background, this caused two `MaskingExtent`s to exist. But the code in https://github.com/IntellectualSites/FastAsyncWorldEdit/blob/2a08ad28a43a3e179c2b78178b27918ffd3c96a3/worldedit-core/src/main/java/com/fastasyncworldedit/core/extent/processor/MultiBatchProcessor.java#L118-L124 prevents from having multiple different `Filter` processors of the same type. As the order in which the processors are called depends on a HashSet, the result would be 50/50 either of the masks.

```[tasklist]
### Submitter Checklist
```

### DIFF
--- a/worldedit-core/src/main/java/com/fastasyncworldedit/core/util/ExtentTraverser.java
+++ b/worldedit-core/src/main/java/com/fastasyncworldedit/core/util/ExtentTraverser.java
@@ -51,11 +51,10 @@ public class ExtentTraverser<T extends Extent> {
         return last;
     }
 
-    @SuppressWarnings("unchecked")
     @Nullable
-    public <U> U findAndGet(Class<U> clazz) {
-        ExtentTraverser<Extent> traverser = find(clazz);
-        return (traverser != null) ? (U) traverser.get() : null;
+    public <U extends Extent> U findAndGet(Class<U> clazz) {
+        ExtentTraverser<U> traverser = find(clazz);
+        return (traverser != null) ? traverser.get() : null;
     }
 
     @SuppressWarnings("unchecked")


dordsor21 (approved):

dordsor21 (approved):

dordsor21:
Only a small change with good big results so just merging

PR #2651, opened 2024-03-25
Simplify processor ordering

SirYwell:
## Description

This is a little refactoring for how processors are ordered before being called.

```[tasklist]
### Submitter Checklist
```

### DIFF
--- a/worldedit-core/src/main/java/com/fastasyncworldedit/core/extent/processor/MultiBatchProcessor.java
+++ b/worldedit-core/src/main/java/com/fastasyncworldedit/core/extent/processor/MultiBatchProcessor.java
@@ -19,12 +19,9 @@ import org.apache.logging.log4j.Logger;
 import javax.annotation.Nullable;
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Collections;
 import java.util.HashMap;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
-import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.Future;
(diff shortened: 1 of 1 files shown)


worldedit-core/src/main/java/com/fastasyncworldedit/core/extent/processor/MultiBatchProcessor.java
```diff
@@ -80,28 +77,18 @@ public void removeBatchProcessor(IBatchProcessor processor) {
 
     @Override
     public IChunkSet processSet(IChunk chunk, IChunkGet get, IChunkSet set) {
-        Map<Integer, Set<IBatchProcessor>> ordered = new HashMap<>();
+        Map<Integer, List<IBatchProcessor>> ordered = new HashMap<>();
```
dordsor21, line 80: Could it be worth caching this per processor and marking dirty when changed?

SirYwell: I tried to profile that code, but I got like 3 traces in that code vs thousands in the `processSet` calls below. So it's not really even measurable yet noticeable.

The change from HashSet to ArrayList should simplify insertion and iteration logic, but given the measurement difficulty, that's rather theoretical.

dordsor21: Hmm fair enough. Should have been cached in the first place really but I suppose it doesn't have any noticeable performance impact

dordsor21 (approved):

NotMyFault (approved):

PR #2682, opened 2024-04-11
Implement Folia Support

SirYwell:
# Rules (read carefully)

This PR will be around for quite some time most likely, so we need a few rules:

- **General**
  - Neither Folia nor this PR are production ready. Expect your server to crash and to hang. Expect to loose (world) data.
  - Only use this software when you're able to debug problems. There are enough people who can run commands until something breaks. There are not enough people who can analyze and resolve such problems.
  - If you use modifications of Folia or FAWE, expect even more things to break. We won't look into your problems in such 
- **Reporting Bugs**
  - Make sure that you don't report bugs that were reported already.
  - Make sure you can reproduce the bug on the latest Folia build with no other plugins installed.
  - There is no benefit in reporting bugs we cannot reproduce at this stage.
  - This PR targets the next major FAWE version. Before reporting bugs, make sure those bugs are not present on the main branch or the v3 branch.
- **Contributing**
  - If you want to help, make sure you understand what's going on.
  - Keep your changes small. I won't look through thousand lines of code.
  - We do not try to re-enable features that just aren't supported by Folia at the moment.
  - Don't try to implement something just to have it implemented. Only implement things you can reason about.
  - Open a PR targeting this branch. Expect your PR to be reviewed carefully. Expect your PR to be rejected.
  - Discussing changes with us before will increase the likelihood of changes useful for us.

  
# Known Bugs and Limitations

- `//regen` does not work as it requires world loading
- Copying entities does not work
- Copying blocks with block entities does not work well

```[tasklist]
### Submitter Checklist
- [ ] Ensure that the pull request title represents the desired changelog entry.
```

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,6 +1,7 @@
 import org.ajoberstar.grgit.Grgit
 import org.gradle.api.tasks.testing.logging.TestExceptionFormat.FULL
 import org.gradle.api.tasks.testing.logging.TestLogEvent.FAILED
+import xyz.jpenilla.runpaper.task.RunServer
 import java.net.URI
 import java.time.format.DateTimeFormatter
 import xyz.jpenilla.runpaper.task.RunServer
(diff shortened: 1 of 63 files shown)


Bufferless:
Issues that I have found when using this version with the latest Folia 1.20.4 without any plugins besides FAWE:

- //up 1 will send an error and not work, it will send "Editing on unsupported versions is disabled"
- Making big changes with FAWE or loading a big schematic will cause FAWE to no longer respond to any worldedit / fawe commands.
- After setting the pos2 with the //wand, it will no longer respond if you try right-clicking a different area making you have to use the command //pos2.
- You will have to relog in order to see changes made with huge edits.

Q2297045667:
Will it be updated? Can you rebase it into the main branch?

Issue #2740, opened 2024-05-20
Track upstream API incompatibilities


PierreSchwang:
FAWE has collected quite a few upstream diffs over the years, which need to be addressed to ensure compatibility (especially with 3rd-party plugins interacting with the WE api).
This issue contains known problems:
- https://github.com/EngineHub/WorldEdit/commit/6e2b0a1df8a6077c3cf8193e38dc9817038bcbe9 (`Clipboard#transform(Transform)` does not exist in FAWE)
- Record-style getters are missing (affects WorldGuard support): https://github.com/EngineHub/WorldEdit/commit/0df2b6af4c1ce18b77eedd5c62eeb45011512103

AndyReckt:
bump

PR #2743, opened 2024-05-23
Fix compatibility with WorldGuard

### LINKED ISSUE #2738: Newer WorldGuard Builds now break since methods are missing in FastAsyncWorldEdit
### Server Implementation

Paper

### Server Version

1.20.5/6

### Describe the bug

Within recent WorldGuard builds, there is an issue to where now it causes WorldGuard to fail because ``com.sk89q.worldedit.math.BlockVector2.x()`` does not exist in current FastAsyncWorldEdit builds, which is now required in newer builds of WorldGuard.

Note that I have tested this on a test server and not in production.

### To Reproduce

1. Download latest experimental build of WorldGuard from https://builds.enginehub.org/job/worldguard/24198 (as of this bug report)
2. Download latest build of FastAsyncWorldEdit from https://ci.athion.net/job/FastAsyncWorldEdit/
3. Grab the latest Paper Build
4. Start the server
5. Error should generate
6. Region data will fail and WorldGuard will disable, thus breaks any other plugin that relies on WorldGuard

### Expected behaviour

Should load region data, if there is any and continue to load the server without issues.

### Screenshots / Videos

_No response_

### Error log (if applicable)

https://paste.gg/p/anonymous/d05b9fbce43242b787832fb640b1c56e

### Fawe Debugpaste

https://athion.net/ISPaster/paste/view/7dec0193f5ad4a839dfb2b484d4ef505

### Fawe Version

FastAsyncWorldEdit v.2.10.0;d8eb03f

### Checklist

### Anything else?

Running Paper version 1.20.6-82-master@0ad09de as of writing this bug report if that helps out

PierreSchwang:
## Overview
Basically, integrates the record updates from upstream in a kind of fitting matter (due to use of abstract classes only "simulated" records).
Started updating references to the now deprecated methods, but there are a few hundreds left - which can be addressed on the fly when working on other things or if someone want's to do that. Works right now, but will extend the amount of deprecation warnings quite a bit (as if that matters lol)

WorldGuard seems to run, and basic WE commands as well(?)

Fixes #2738
Partially addresses #2740

```[tasklist]
### Submitter Checklist
```

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/math/BlockVector3.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/math/BlockVector3.java
@@ -87,9 +87,9 @@ public abstract class BlockVector3 {
     }
 
     public static boolean isLongPackable(BlockVector3 location) {
-        return isHorizontallyInBounds(location.getX())
-                && isHorizontallyInBounds(location.getZ())
-                && WORLD_Y_MIN <= location.getY() && location.getY() <= WORLD_Y_MAX;
+        return isHorizontallyInBounds(location.x())
+                && isHorizontallyInBounds(location.z())
+                && WORLD_Y_MIN <= location.y() && location.y() <= WORLD_Y_MAX;
     }
 
     public static void checkLongPackable(BlockVector3 location) {
@@ -109,9 +109,9 @@ public abstract class BlockVector3 {
     private static final class YzxOrderComparator {
 
         private static final Comparator<BlockVector3> YZX_ORDER =
-                Comparator.comparingInt(BlockVector3::getY)
-                        .thenComparingInt(BlockVector3::getZ)
-                        .thenComparingInt(BlockVector3::getX);
+                Comparator.comparingInt(BlockVector3::y)
+                        .thenComparingInt(BlockVector3::z)
+                        .thenComparingInt(BlockVector3::x);
 
     }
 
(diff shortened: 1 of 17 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/math/BlockVector3.java
```diff
@@ -135,57 +137,69 @@ public MutableBlockVector3 setComponents(int x, int y, int z) {
...
 
     /**
      * Get the X coordinate.
      *
      * @return the x coordinate
+     * @deprecated use {@link #x()} instead
      */
-    //FAWE start - getter
+    @Deprecated(forRemoval = true)
+    public int getX() {
```
dordsor21, line 189: `@since` tabs

worldedit-core/src/main/java/com/sk89q/worldedit/math/Vector2.java
```diff
@@ -83,11 +80,22 @@ public Vector2 withX(double x) {
         return Vector2.at(x, z);
     }
 
+    /**
+     * Get the Z coordinate, aligned to the block grid.
+     *
+     * @return the block-aligned z coordinate
+     */
+    public int blockZ() {
```
dordsor21, line 88: Since tags here too

dordsor21 (commented):
Just the since tags, otherwise, I wonder if we want/need to keep hashcodes the same as upstream? Should also put the FAWE start/end comments around those

SirYwell (approved):

NotMyFault (approved):

PR #2750, opened 2024-05-27
chore: address deprecations in vector classes

dordsor21:
:)

### DIFF
--- a/worldedit-core/src/main/java/com/fastasyncworldedit/core/function/pattern/BufferedPattern2D.java
+++ b/worldedit-core/src/main/java/com/fastasyncworldedit/core/function/pattern/BufferedPattern2D.java
@@ -33,7 +33,7 @@ public class BufferedPattern2D extends BufferedPattern {
 
     @Override
     public boolean set(BlockVector3 pos) {
-        return set.add(pos.x(), 0, pos.y());
+        return set.add(pos.x(), 0, pos.z());
     }
 
 }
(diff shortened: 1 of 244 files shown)


PierreSchwang (approved):
uhm, yes - totally 100% reviewed.

worldedit-core/src/main/java/com/fastasyncworldedit/core/function/pattern/BufferedPattern2D.java
```diff
@@ -33,7 +33,7 @@ public BufferedPattern2D(Actor actor, Pattern parent, @Nullable Region region) {
 
     @Override
     public boolean set(BlockVector3 pos) {
-        return set.add(pos.getBlockX(), 0, pos.getBlockY());
+        return set.add(pos.x(), 0, pos.y());
```
dordsor21, line 36: Should this be x and z? Been Y since 2018

worldedit-core/src/main/java/com/sk89q/worldedit/function/mask/NoiseFilter.java
```diff
@@ -97,7 +97,7 @@ public void setDensity(double density) {
     @Override
     public boolean test(BlockVector3 vector) {
         //FAWE start - mutable
-        return noiseGenerator.noise(mutable.setComponents(vector.getX(), vector.getZ(), vector.getZ())) <= density;
+        return noiseGenerator.noise(mutable.setComponents(vector.x(), vector.z(), vector.z())) <= density;
```
dordsor21, line 100: Should be x,y,z not x,z,z?

dordsor21:
This should probably be rebased and merged rather than squashed so that the coord fixes will be a separate commit on main

NotMyFault (approved):

PR #2755, opened 2024-06-02
feat: allow NBT on wand items

### LINKED ISSUE #2383: NBT on items in configuration file

I want you to add CustomModelData support to the plugin. For example in the file worldedit-config.yml 
wand-item: 'PAPER {CustomModelData: 10162}'

I do not know...

### Anything else?

_No response_

dordsor21:
- closes #2383

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/command/SelectionCommands.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/command/SelectionCommands.java
@@ -342,7 +342,6 @@ public class SelectionCommands {
                 return;
             }
         }
-        System.out.println("a "+ wand);
         player.giveItem(new BaseItemStack(wand.getType(), wand.getNbtReference(), 1));
         //FAWE end
         //FAWE start - instance-iate session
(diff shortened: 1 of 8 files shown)


SirYwell (approved):

worldedit-core/src/main/java/com/sk89q/worldedit/command/SelectionCommands.java
```diff
@@ -325,22 +328,29 @@ public void wand(
...
+            parserContext.setSession(session);
+            try {
+                wand = WorldEdit.getInstance().getItemFactory().parseFromInput(wandId, parserContext);
+            } catch (InputParseException e) {
+                player.print(Caption.of("worldedit.wand.invalid"));
+                return;
+            }
         }
-        player.giveItem(new BaseItemStack(itemType, 1));
+        System.out.println("a "+ wand);
```
PierreSchwang, line 345: ^

dordsor21: Stop finding my debug

PierreSchwang (approved):
Untested, but looks good so far

dordsor21:
> Untested, but looks good so far

I tested setting damage to an axe and it worked so I called it okay. No idea how the specific use in the issue would work but Minecraft has forever allowed for custom NBT on basically everything so there's no reason for NBT that is consequently picked up by something else to be set (and incidentally I tested this because of the toLowerCase on the config strings as damage is "Damage" in NBT :) )

NotMyFault (approved):

PR #2757, opened 2024-06-02
fix: improve biome setting to avoid writing directly to chunk

dordsor21:
- Removes possibility of writing to the LevelChunkSection biomes PalettedContainer whilst it is being read from the main thread

### DIFF
--- a/worldedit-bukkit/adapters/adapter-1_19_4/src/main/java/com/sk89q/worldedit/bukkit/adapter/impl/fawe/v1_19_R3/PaperweightPlatformAdapter.java
+++ b/worldedit-bukkit/adapters/adapter-1_19_4/src/main/java/com/sk89q/worldedit/bukkit/adapter/impl/fawe/v1_19_R3/PaperweightPlatformAdapter.java
@@ -97,7 +97,7 @@ public final class PaperweightPlatformAdapter extends NMSAdapter {
 
     private static final Field fieldTickingFluidCount;
     private static final Field fieldTickingBlockCount;
-    private static final Field fieldNonEmptyBlockCount;
+    private static final Field fieldBiomes;
 
     private static final MethodHandle methodGetVisibleChunk;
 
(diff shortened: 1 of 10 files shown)


SirYwell:
From my understanding, the linked issue is caused by the fact that we manipulate the `LevelChunk` object. The `ClientboundLevelChunkPacketData` writes a chunk in two steps:
1. Calculate a byte array, with a size calculated by `calculateChunkSize(LevelChunk)`
   - calls `getSections()` and calls `LevelChunkSection#getSerializedSize()` on each of the sections
2. With this byte array, it now calls `extractChunkData`, and passes the chunk again
    - calls `getSections()` *again* and calls `LevelChunkSection#write(...)` on each of the sections

So if we change anything about the `LevelChunk` object in between those two steps (i.e. modifying the `sections` array, or replacing the `biomes` palette of a `LevelChunkSection` (solution proposed here), or writing to a `biomes` palette (current solution)) we might get a mismatch between the calculated size and the actual size it tries to write.

Am I missing something?

dordsor21:
Ah yeah that does make more sense. The changes here are ones that should probably go in anyway so I'll edit the PR desc

SirYwell (approved):

worldedit-bukkit/adapters/adapter-1_19_4/src/main/java/com/sk89q/worldedit/bukkit/adapter/impl/fawe/v1_19_R3/PaperweightPlatformAdapter.java
```diff
@@ -138,8 +138,15 @@ public final class PaperweightPlatformAdapter extends NMSAdapter {
...
-            fieldNonEmptyBlockCount = LevelChunkSection.class.getDeclaredField(Refraction.pickName("nonEmptyBlockCount", "f"));
-            fieldNonEmptyBlockCount.setAccessible(true);
+            Field tmpFieldBiomes;
+            try {
+                // It seems to actually be biomes, but is apparently obfuscated to "j"
+                tmpFieldBiomes = LevelChunkSection.class.getDeclaredField("biomes");
+            } catch (NoSuchFieldException ignored) {
+                tmpFieldBiomes = LevelChunkSection.class.getDeclaredField("j");
+            }
+            fieldBiomes = tmpFieldBiomes;
```
SirYwell, line 148: Whoops one more question, why isn't this using Refraction like the other fields?

dordsor21: > // It seems to actually be biomes, but is apparently obfuscated to "j"

MiniMappingViewer said `j` but when actually using it, it was `biomes` for whatever reason

SirYwell: If you tested that on paper, you probably encountered the remapped version - but paper should also successfully remap `j` to `biomes` then

SirYwell: It works for me on 1.20.6

dordsor21: I was on 1.20.4, I also didn't bother checking every version, just assumed it could be an issue there as well

SirYwell: It also works on 1.20.4 for me, Refraction properly detects that it's running in a mojmapped environment

dordsor21: I'll stick it back to normal and assume it was something off just once

NotMyFault (approved):

PR #2768, opened 2024-06-08
chore: address deprecations from new records

dordsor21:


worldedit-core/src/main/java/com/sk89q/worldedit/world/World.java
```diff
@@ -419,7 +419,7 @@ default boolean isWorld() {
     }
 
     @Override
-    default String getId() {
+    default String id() {
```
dordsor21, line 422: Thoughts on altering the overriden method?

worldedit-core/src/main/java/com/sk89q/worldedit/world/NullWorld.java
```diff
@@ -78,7 +78,7 @@ public String getNameUnsafe() {
     //FAWE end
 
     @Override
-    public String getId() {
+    public String id() {
```
dordsor21, line 81: Thoughts on altering the overriden method?

worldedit-core/src/main/java/com/sk89q/worldedit/registry/Category.java
```diff
@@ -36,7 +36,7 @@ protected Category(final String id) {
         this.id = id;
     }
 
-    public final String getId() {
+    public final String id() {
```
SirYwell, line 39: This is not an overridden method, so that's breaking I guess?

worldedit-core/src/main/java/com/sk89q/worldedit/world/NullWorld.java
SirYwell: Should be fine as long as `getId()` calls `id()`

SirYwell (changes requested):

worldedit-core/src/main/java/com/sk89q/worldedit/registry/Category.java
dordsor21: Upstream had it implementing Keyed, fixed

SirYwell (approved):

PierreSchwang (approved):

PR #2773, opened 2024-06-09
feat: improve fawe limits

### LINKED ISSUE #1874: Bad calculation of the number of blocks changed?
### Server Implementation

Paper

### Server Version

1.18.2

### Describe the bug

```yaml
limits:
[some-line]
  builder-plus:
    max-actions: 2
    max-changes: 8000000
    max-checks: 80000000
    max-fails: 4000000
    max-iterations: 500
[some-line]
```

image

### To Reproduce

1. Define limitation

### Expected behaviour

### Screenshots / Videos

image

### Error log (if applicable)

_No response_

### Fawe Debugpaste

### Fawe Version

FastAsyncWorldEdit-2.4.1

### Checklist

### Anything else?

_No response_

### LINKED ISSUE #2679: max-blockstates config limit is not enforced on //set
### Server Implementation

Paper

### Server Version

1.20.4

### Describe the bug

I have `max-blockstates: 5` however this restriction doesn't seem to be enforced when using `//set` or `//set #clipboard`.

Here is an example of where I was able to set 59,000 hoppers when it should have stopped me:
Example: //set hopper

### To Reproduce

1. Set max-blockstates: to a low number
2. Set a block entity beyond the limit you set

### Expected behaviour

WorldEdit action should be prevented

### Screenshots / Videos

_No response_

### Error log (if applicable)

_No response_

### Fawe Debugpaste

https://athion.net/ISPaster/paste/view/a64ea766b59b41a796b4fb6fe87c1008

### Fawe Version

2.9.3-SNAPSHOT-724;ca2c18e

### Checklist

### Anything else?

_No response_

dordsor21:
- add FaweLimit implementations for increasing concurrency levels
 - allow FaweLimit to perform processing (and forcefully disable as required) to capture [tile] entities
 - Use `BlockVector3#set(Extent orDefault)` where appropriate to reduce block checks
 - fixes #2679
 - fixes #1874


worldedit-core/src/main/java/com/sk89q/worldedit/command/BrushCommands.java
```diff
@@ -992,15 +992,15 @@ public void command(
                     Expression radius,
             @Arg(desc = "Command to run")
                     List<String> input,
-            @Switch(name = 'p', desc = "Show any printed output")
-                    boolean print
+            @Switch(name = 'h', desc = "Hide any printed output")
+                    boolean hide
```
dordsor21, line 996: Not having output from this brush is awkward, especially since it only runs the command once

PierreSchwang (approved):
Untested, but looks good code-wise

NotMyFault (approved):

PR #2776, opened 2024-06-11
Support Sponge Schematic v3

PierreSchwang:
## Overview
WorldEdit's new default schematic is v3 of sponge - the rise of named schematics requires FAWE to follow up.

## Description
- Adds upstream classes and schematic reader/writer.
- Adds FAWE-specific optimized writers / (readers?)
  - Writer is faster, especially on larger regions
    - For 300mil blocks WorldEdit's implementation takes about 41s whereas the fast implementation takes about 15s

```[tasklist]
### Submitter Checklist
```


worldedit-core/src/main/java/com/fastasyncworldedit/core/extent/clipboard/io/FastSchematicReaderV3.java
```diff
@@ -0,0 +1,326 @@
...
+        if (!dataInputStream.markSupported()) {
+            throw new IOException("InputStream does not support mark");
+        }
+    }
+
+    @Override
+    public Clipboard read(final UUID uuid, final Function<BlockVector3, Clipboard> createOutput) throws IOException {
+        dataInputStream.skipNBytes(1 + 2); // 1 Byte = TAG_Compound, 2 Bytes = Short (Length of tag name = "")
+        dataInputStream.skipNBytes(1 + 2 + 9); // as above + 9 bytes = "Schematic"
+        this.dataInputStream.mark(Integer.MAX_VALUE); // allow resets to basically the start of stream (file) - just skip header
```
dordsor21, line 76: We should make certain this is a resettableinputstream

PierreSchwang: When used by the BuiltinClipboardFormats enums, the stream is wrapped into a ResettableFileInputStream. Or do you mean, that the variable should show more, that the stream is optimized for resets without flooding the memory?

dordsor21: This is public API that we should expect to be used basically

PierreSchwang:
The reset itself seems to work quite nicely with the FileInputStream - but `//schem load` supports url (using SchematicWeb iirc?) which passes the Response as the InputStream. That would require the BufferedInputStream to hold all bytes for the reset instead of utilizing FileChannels.

Either these schematics are stored temporary on disks (meh) or the FastReader should just delegate to the "normal" SpongeSchematicReaderV3?

dordsor21:
I figure it's probably okay to handle web schematics like we do V2 with a compressed cache for what we can't read yet. It'll be no more memory intense than WE, but allows us to make us of correctly ordered schematics still if it's present

PierreSchwang:
(Tile-)Entities are working (now?) for me and schematics seem to be cross-compatible between fast + WE implementation (which guarantees conformity with the spec I'd say lol)

Fast Implementations are a tad faster than it's WE counterparts - especially for larger schematics - but don't consume near as much memory while saving / reading. The garbage collector isn't doing much either while the fast implementation does it's job.

PierreSchwang:
Worth noting: Passenger entities are not working - Well, they are, but metadata (e.g. sheep wool color) is not applied.
The data is available in the schematic but is not applied on paste. Though same applies for the WE implementation so it's rather an issue with one of the FAWE extents.

worldedit-core/src/main/java/com/fastasyncworldedit/core/extent/clipboard/io/FastSchematicReaderV3.java
```diff
@@ -0,0 +1,824 @@
...
+     * <p>
+     * Requires {@link #read()}, {@link #read(UUID)} or {@link #read(UUID, Function)} to be called before.
+     */
+    @Override
+    public OptionalInt getDataVersion() {
+        return this.dataVersion > -1 ? OptionalInt.of(this.dataVersion) : OptionalInt.empty();
+    }
+
+    private void readBlocks(Clipboard target) throws IOException {
+        this.blockPalette = new BlockState[BlockTypesCache.states.length];
```
dordsor21, line 298: Ensures that `() -> this.blockPalette.length == 0` is always false as it is always read imemdiately in the `readPalette` call..?

Do we also ever used this "default" value? It seems like an unnecessary memory expense

worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/io/BuiltInClipboardFormat.java
```diff
@@ -19,39 +19,105 @@
...
+                    byte type = stream.readByte();
+                    String name = stream.readUTF();
+                    if (type == NBTConstants.TYPE_END) {
+                        return false;
+                    }
+                    if (type == NBTConstants.TYPE_INT && name.equals("Version")) {
+                        return stream.readInt() == FastSchematicWriterV3.CURRENT_VERSION;
+                    }
+                    nbt.readTagPayloadLazy(type, 0);
+                } while (true);
```
dordsor21, line 109: I really hate this about the sponge format, but I guess letting people put whatever file extension they want is important for some reason...? Really stupid aspect of the format and I wouldn't be opposed to checking file extension == .schem or .sponge and maybe having a magic something that FAWE writes into the schematic we can check? Given the header of the schematic will always be the same (if written by FAWE) perhaps we can just check magic N bytes matches the start of the file, whether that basically includes the `"": { "Schematic": { "Version": 3` or whatever

worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/io/share/BuiltInClipboardShareDestinations.java
```diff
@@ -75,7 +75,7 @@ public Consumer<Actor> share(ClipboardShareMetadata metadata, ShareOutputProvide
 
         @Override
         public ClipboardFormat getDefaultFormat() {
-            return BuiltInClipboardFormat.SPONGE_SCHEMATIC;
+            return BuiltInClipboardFormat.SPONGE_V2_SCHEMATIC;
```
dordsor21, line 78: I guess given upstream defaults to this then we ought to

worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/io/sponge/ReaderUtil.java
```diff
@@ -0,0 +1,283 @@
...
+import java.util.List;
+import java.util.Map;
+import javax.annotation.Nullable;
+
+import static com.google.common.base.Preconditions.checkState;
+
+/**
+ * Common code shared between schematic readers.
+ */
+public class ReaderUtil { //FAWE - make public
```
dordsor21, line 64: Does it need to be? Doesn't look like it's used

worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/io/BuiltInClipboardFormat.java
PierreSchwang: I guess for these few bytes, yea - simple BufferedReader#mark + reset should work

dordsor21: We're still having to read the whole file though, I'm suggesting having two checks that will return much quicker - limiting the file extensions we support (as is done by literally everything - wrong extensions, assume wrong file) and checking for some "magic" that FAWE writes, as we have control over that, otherwise, use this method

PierreSchwang: Exactly, the Stream is wrapped into a Buffered stream either way (iirc). Even when checking for the magic bytes (the schematic part you posted) it'd have to reset potentially. When the sequence does not match, the stream might be somewhere in between a compound or tag name.

dordsor21: Yeah that's true, I would still add a file extension check though

worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/io/share/BuiltInClipboardShareDestinations.java
PierreSchwang: ah, might be a good idea to use the fast impl there to - no need to use the upstream implementation when the schematic is conform either way.

worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/io/BuiltInClipboardFormat.java
PierreSchwang: Yeah, might be worth doing:
> The highly recommended file extension for files using this specification is .schem

So I guess we can enforce that

worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/io/share/BuiltInClipboardShareDestinations.java
PierreSchwang: I now see that this is for the EngineHub Paster - which I would not want to touch. I'll update the supported + default format for Ark though,

worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/io/BuiltInClipboardFormat.java
```diff
@@ -265,6 +400,80 @@ public String getPrimaryFileExtension() {
...
+            0x00, 0x00, // len("") - unsigned short (no following bytes for string because there is none)
+            0x10, // TAG_Compound
+            0x00, 0x09, // len("Schematic") - unsigned short
+            0x53, 0x63, 0x68, 0x65, 0x6D, 0x61, 0x74, 0x69, 0x63, // "Schematic" (modified UTF-8)
+            0x03, // TAG_INT
+            0x00, 0x07, // len("Version") - unsigned short
+            0x56, 0x65, 0x72, 0x73, 0x69, 0x6F, 0x6E  // "Version" (modified UTF-8)
+    };
+
+    private static boolean detectOldSpongeSchematic(InputStream inputStream, int version) {
```
dordsor21, line 423: Do sponge v1 and v2 definitely have the same magic bytes?

worldedit-core/src/main/java/com/sk89q/worldedit/extent/clipboard/io/BuiltInClipboardFormat.java
dordsor21: Or maybe we don't so we don't have to deal with the occasional person that tries to load a v2 or a v3 schematic with a .schematic extension because some schematic download sites are stupid

worldedit-core/src/main/java/com/fastasyncworldedit/core/extent/clipboard/io/FastSchematicReaderV3.java
PierreSchwang: oic, I fucked something up there

PierreSchwang: The check is now correct - it's used so the code can skip the palette initializing (aka assigning the block states to the ordinals from the schematic) if the method is called again (see `#readRemainingDataReset`)

dordsor21: But it will still always return true (now)?

PierreSchwang: in which case?

dordsor21: Yeah nevermind missed that the calling method was called from two places

PierreSchwang: *Theoretically* the spec does not enforce the palette to start with index `0`:
> It is recommended for maximum data compression that your indices start at zero and skip no values

But highly recommends it. In worst case, if some schematic generators don't do it that way the palette will just be read twice - which is okay for me

dordsor21 (approved):

dordsor21 (approved):

dordsor21 (approved):

SirYwell (changes requested):

NotMyFault (approved):

itsme-to:
Hey it's make error in loading a lot of schematic, It's don't support well schematic made in v2.

https://paste.itsme.to/toxitefuga.sql

PR #2778, opened 2024-06-14
ref: switch from adventure NBT to LinBus

dordsor21:
- Hard replace all adventure NBT with linbus
 - Keep where we continue to use "deprecated" jnbt


NotMyFault (approved):

gradle/libs.versions.toml
```diff
@@ -40,6 +40,7 @@ paperlib = "1.0.8"
 paster = "1.1.6"
 vault = "1.7.1"
 serverlib = "2.3.6"
+linbus = "0.1.0-SNAPSHOT"
```
PierreSchwang, line 43: 0.1.0 was released as a non-snapshot version (yesterday?) - might be worth directly switching (https://github.com/EngineHub/WorldEdit/commit/203236d22a98de1251d0f1583da4b9a7aea37dde)

worldedit-bukkit/adapters/adapter-1_20_5/src/main/java/com/sk89q/worldedit/bukkit/adapter/ext.fawe/v1_20_R4/PaperweightDataConverters.java
```diff
@@ -81,16 +80,16 @@
 @SuppressWarnings({ "rawtypes", "unchecked" })
 public class PaperweightDataConverters extends DataFixerBuilder implements com.sk89q.worldedit.world.DataFixer {
 
-    //FAWE start - BinaryTag
+    //FAWE start - LinTag
```
PierreSchwang, line 83: FAWE does not diff from Upstream here (anymore)
https://github.com/EngineHub/WorldEdit/blob/version/7.3.x/worldedit-bukkit/adapters/adapter-1.20.6/src/main/java/com/sk89q/worldedit/bukkit/adapter/impl/v1_20_R4/PaperweightDataConverters.java#L88

(At least for `fixUp(FixType, T, int)`, the actual `fix...` methods use fully qualified imports)



Applies to the remaining DataConverters as well

worldedit-core/src/main/java/com/fastasyncworldedit/core/extent/clipboard/io/FastSchematicReader.java
```diff
@@ -109,10 +109,10 @@ private CompoundTag fixBlockEntity(CompoundTag tag) {
         if (fixer == null || dataVersion == -1) {
             return tag;
         }
-        //FAWE start - BinaryTag
-        return (CompoundTag) AdventureNBTConverter.fromAdventure(fixer.fixUp(
+        //FAWE start - LinTag
```
PierreSchwang, line 112: Why did this class had fawe annotations in the first place if it's not an upstream class?

PierreSchwang (changes requested):
Would be great if the annotations could be removed, other than that looks good imo

worldedit-core/src/main/java/com/fastasyncworldedit/core/extent/clipboard/io/FastSchematicReader.java
PierreSchwang: You wanna remove from the FastReader / (Writer?) as well?

PierreSchwang: different PR but I take it lol

dordsor21: Ah I checked out the wrong branch. Oh well

PierreSchwang (approved):
LGTM (though untested - but as long as it compiles?)

PR #2803, opened 2024-06-27
fix: override min/max pos methods from SimpleClipboard

### LINKED ISSUE #2800: Creating a schematic with biomes using `//lazycopy -b` results in wrong biomes
### Server Implementation

Paper

### Server Version

1.20.5/6

### Describe the bug

When using `//lazycopy -b` for creating a schematic with biome data, the biome will be wrong. 

The underlying `WorldCopyClipboard`s `getBiome(BlockVector3)` method is called with relative coordinates, whereas `getBlock` is called using absolute coordinates. This results in the the relative coordinate being used as absolute coordinates for world access. 

That is no issue with the actual clipboard (backed by e.g. DiskOptimizedClipboard) as those use a linear pattern and need to be called with relative coordinates. 

### To Reproduce

1. select region
2. `//lazycopy -b`
3. `//schem create <filename>` / `//schem create <filename> safe`
4. `//schem load <filename>`
5. `//paste -b`

### Expected behaviour

The biome should be retrieved from the region, not based of world 0;0

### Checklist

dordsor21:
- fixes #2800

### DIFF
--- a/worldedit-core/src/main/java/com/fastasyncworldedit/core/extent/clipboard/ReadOnlyClipboard.java
+++ b/worldedit-core/src/main/java/com/fastasyncworldedit/core/extent/clipboard/ReadOnlyClipboard.java
@@ -2,13 +2,10 @@ package com.fastasyncworldedit.core.extent.clipboard;
 
 import com.fastasyncworldedit.core.Fawe;
 import com.sk89q.jnbt.CompoundTag;
-import com.sk89q.jnbt.NBTUtils;
-import com.sk89q.jnbt.Tag;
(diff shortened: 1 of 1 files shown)


PierreSchwang (approved):
Works for me

TechnicallyCoded:
I can still reproduce the issue. Details:
https://github.com/IntellectualSites/FastAsyncWorldEdit/issues/2800#issuecomment-2195921775

dordsor21:
FWIW if the biome is always forest, that suggests a different issue. This PR addresses the specific issue fine

PR #2827, opened 2024-07-06
Add worldedit.tool.none to fawe.permpack.basic

RadND:
If node "fawe.permpack.basic" exist in other platform than bukkit,this minor bug would exist here too,but i lack the knowledge of finding it out.

A really minor fix , so I just ignore all the contributing requirements :)


dordsor21 (approved):

NotMyFault (approved):
Thanks!

PR #2836, opened 2024-07-18
Compile with target java 21, remove unsupported MC versions

SirYwell:
## Description

LinBus is compiled with Java 21, but older Spigot releases can't load those classes because the bundled ASM version is too old. This means we can't support these versions anymore, which also means we can target Java 21 with FAWE too.

1.20.1 still seems to be used somewhat widely according to bStats, but I don't see any way for us to still provide support for it.

```[tasklist]
### Submitter Checklist
```


NotMyFault:
Oh I can't approve it by hey, lgtm

NotMyFault (approved):

dordsor21 (approved):

dordsor21:
Should probably adjust the issue template

MattBDev:
Sounds similar to how we had to split with Legacy FAWE due to the source code being a mess and other issues. I'm all for dropping support for older versions if there is no way around it. Could we bundle a more up to date version of ASM or would that cause some other compatibility issues?

SirYwell:
> Could we bundle a more up to date version of ASM or would that cause some other compatibility issues?

The ASM version is provided and loaded by the server. We can't change that in any way.

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -27,9 +27,9 @@ body:
       description: Which server version version you using? If your server version is not listed, it is not supported. Update to a supported version first.
       multiple: false
       options:
+        - '1.21'
         - '1.20.5/6'
         - '1.20'
```
dordsor21, line 32: 1.20.2 I guess?

SirYwell: I changed it to mention only the versions we actually support now (unless I missed something)

PR #2838, opened 2024-07-20
fix: improve schematic format selection

dordsor21:
- no longer allow selecting a format specifically because of the generic extension `.schem`


PierreSchwang:
I guess I misunderstand something - but how would the explicit extension work? The extensions are always schem and not schem3 / schem2? Or is it just in case the schematic is named like that and then falls back to the version inside the schematic?

dordsor21:
The problem was that it didn't work now that we have V2 and V3 both with .schem. explicit extensions allow for defining the format in the extension to maintain the possibility, but I do not expect it to be used often. The main goal of the PR is to fix (by removing) format choice by extension for indeterminate extension (i.e. .schem)

PierreSchwang (approved):
LGTM

PR #2844, opened 2024-07-21
fix: improve FAWE stream history

### LINKED ISSUE #2583: Can't undo moving blocks far out of plot region
### Server Implementation

Paper

### Server Version

1.20.2

### Describe the bug

When a player moves their plot far outside of it's region, the action cannot be undone

### To Reproduce

1. Select region in plot
2. Move that region far away with `//move 300`
3. Try to undo the action with `//undo`

### Expected behaviour

The region should reappear when the action is undone

### Screenshots / Videos

https://youtu.be/hC_Sb-7Pkag

### Error log (if applicable)

None

### Plot Debugpaste

https://athion.net/ISPaster/paste/view/6b358805637f4cbfaa333a355c3247e8

### PlotSquared Version

PlotSquared 7.1.0-Premium

### Checklist

- [X] I have included a Plot debugpaste.
- [X] I am using the newest build from https://www.spigotmc.org/resources/77506/ and the issue still persists.

### Anything else?

_No response_

dordsor21:
- reset "origin"/relative X/Z when larger than short min/max value so we do not write incorrect positions
 - history size can be larger than int max value
 - fixes #2583

### DIFF
--- a/worldedit-core/src/main/java/com/fastasyncworldedit/core/history/changeset/FaweStreamChangeSet.java
+++ b/worldedit-core/src/main/java/com/fastasyncworldedit/core/history/changeset/FaweStreamChangeSet.java
@@ -37,8 +37,8 @@ public abstract class FaweStreamChangeSet extends AbstractChangeSet {
 
     public static final int HEADER_SIZE = 9;
     private static final int VERSION = 2;
-    // equivalent to Short#MIN_VALUE three times
-    private static final byte[] MAGIC_NEW_RELATIVE = new byte[]{(byte) 128, 0, (byte) 128, 0, (byte) 128, 0};
+    // equivalent to Short#MIN_VALUE three times stored with [(x) & 0xff, ((rx) >> 8) & 0xff]
+    private static final byte[] MAGIC_NEW_RELATIVE = new byte[]{0, (byte) 128, 0, (byte) 128, 0, (byte) 128};
     private int mode;
     private final int compression;
     private final int minY;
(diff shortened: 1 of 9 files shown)


worldedit-core/src/main/java/com/fastasyncworldedit/core/history/changeset/FaweStreamChangeSet.java
```diff
@@ -266,12 +294,19 @@ public boolean isEmpty() {
...
         // Flush so we can accurately get the size
         flush();
         return blockSize;
     }
 
+    @Override
+    public int size() {
+        // Flush so we can accurately get the size
+        flush();
+        return (int) blockSize;
```
SirYwell, line 307: this could simply call `longSize()` instead

worldedit-core/src/main/java/com/fastasyncworldedit/core/history/changeset/FaweStreamChangeSet.java
```diff
@@ -35,11 +36,18 @@
 public abstract class FaweStreamChangeSet extends AbstractChangeSet {
 
     public static final int HEADER_SIZE = 9;
-    private static final int version = 1;
+    private static final int VERSION = 2;
+    // equivalent to Short#MIN_VALUE three times
+    private static final byte[] MAGIC_NEW_RELATIVE = new byte[]{(byte) 128, 0, (byte) 128, 0, (byte) 128, 0};
```
SirYwell, line 41: might be worth to not that this is big endian

dordsor21: The initial implementation was in fact wrong, so adjusted to match how it would actually be stored and added a descriptor for it

NotMyFault (approved):

SirYwell (approved):

PR #2850, opened 2024-07-25
feat: add litematica error when failing to load schematic

dordsor21:


worldedit-core/src/main/java/com/sk89q/worldedit/command/SchematicCommands.java
```diff
@@ -421,6 +421,9 @@ public void load(
         } catch (URISyntaxException | IOException e) {
             actor.print(Caption.of("worldedit.schematic.file-not-exist", TextComponent.of(Objects.toString(e.getMessage()))));
             LOGGER.warn("Failed to load a saved clipboard", e);
+        } catch (Exception e) {
+            actor.print(Caption.of("fawe.worldedit.schematic.schematic.load-failed", TextComponent.of(e.getMessage())));
```
PierreSchwang, line 425: Wouldn't that put the exception message after `Version: ` in the message?

dordsor21: That would get caught be the IOException above (current behaviour)

PierreSchwang: Ah I'm blind - wrong translation key....

PierreSchwang (approved):

SirYwell (approved):

PR #2852, opened 2024-07-26
chore: deprecate FaweApi#load for clipboards as it does not allow closing

dordsor21:


NotMyFault (approved):

PierreSchwang:
Shouldn't the method be fixed in ClipboardFormat then - or rather deprecated as well? Upstream does not contain those methods - so it's a pure design choice on our end then.

dordsor21:
> Shouldn't the method be fixed in ClipboardFormat then - or rather deprecated as well? Upstream does not contain those methods - so it's a pure design choice on our end then.

I'd rather we just remove methods that introduce actual API diff from upstream for no reason tbh

PierreSchwang (approved):

SirYwell (approved):

Issue #2729, opened 2024-05-18
Error generated in console when sending chunk packets to users after pasting a set of schematics


WyndevCodes:
### Server Implementation

Paper

### Server Version

1.20

### Describe the bug

I run a dungeon server that generates dungeons and then teleports players into those dungeons. The players are teleported from the spawn world into the dungeon world after the schematics are loaded.

The issue is that I get an error almost every time a dungeon is generated (see the stack trace below). It doesn't seem to do anything normally, but it may result rarely in parts of the dungeon not being pasted properly. 

Here is the code I use to paste the schematics (this runs about 16 times in 1-5 ticks on an async thread):
```java
public boolean pasteSchematic(String schematicName, Location loc, Main main) {

        World weWorld = FaweAPI.getWorld(loc.getWorld().getName());
        File file = new File(main.getDataFolder() + File.separator + "schematics" + File.separator + schematicName + ".schem");
        if (!file.exists()) {
            Bukkit.getLogger().warning("The dungeon room file " + schematicName + ".schem does not exist in the DungeonDodge schematics folder!");
            return false;
        }
        //Bukkit.getLogger().info("[Dungeon Generation] Dungeon room instance has schematic " + schematicName + ".schem");

        BlockVector3 vec = BlockVector3.at(loc.getBlockX(), loc.getBlockY(), loc.getBlockZ());
        try {
            ClipboardFormat clipboardFormat = ClipboardFormats.findByFile(file);
            if (clipboardFormat != null) {
                Clipboard clipboard = clipboardFormat.load(file);
                EditSession editSession = clipboard.paste(weWorld, vec);
                clipboard.close();
                editSession.close();
            }
        } catch (IOException e) {
            Bukkit.getLogger().severe("Schematic " + schematicName + ".schem was not pasted properly!");
            e.printStackTrace();
        }
        return true;
    }
```

### To Reproduce

1. Create a piece of code that pastes schematics in a sequential form inside of a minecraft world different from the one a player resides in (e.g. paste a hollow square 16 times, adding the square's length to the z-axis each time). This code should be run async and via a command.
2. Teleport the player inside of the first square pasted.
3. Check console for any errors.

### Expected behaviour

There should be no error when sending pasted chunk packets to players.

### Error log (if applicable)

https://paste.gg/p/anonymous/9edac91f450e4000b82844d086674fa4

### Fawe Debugpaste

[...]

PR #2853, opened 2024-07-26
fix: some improvements to GET chunk writing

dordsor21:
- ensure levelChunk is loaded before giving to copy GET - this is not necessarily guaranteed to be nonnull if two edits overlap. Whilst not advised, such an easy failure should not occur when two edits collide
 - aims to help #2729

### DIFF
--- a/worldedit-bukkit/src/main/java/com/fastasyncworldedit/bukkit/adapter/NMSAdapter.java
+++ b/worldedit-bukkit/src/main/java/com/fastasyncworldedit/bukkit/adapter/NMSAdapter.java
@@ -1,5 +1,6 @@
 package com.fastasyncworldedit.bukkit.adapter;
 
+import com.fastasyncworldedit.bukkit.FaweBukkitWorld;
 import com.fastasyncworldedit.core.FAWEPlatformAdapterImpl;
 import com.fastasyncworldedit.core.math.IntPair;
 import com.fastasyncworldedit.core.queue.IChunkGet;
@@ -144,8 +145,6 @@ public class NMSAdapter implements FAWEPlatformAdapterImpl {
         ((BukkitGetBlocks) chunk).send();
     }
 
-    private static final ConcurrentHashMap<String, ConcurrentHashMap<IntPair, ChunkLock>> SENDING_CHUNKS = new ConcurrentHashMap<>();
-
     /**
      * Atomically set the given chunk section to the chunk section array stored in the chunk, given the expected existing chunk
      * section instance at the given layer position.
@@ -170,13 +169,10 @@ public class NMSAdapter implements FAWEPlatformAdapterImpl {
             return false;
         }
         StampLockHolder holder = new StampLockHolder();
-        ConcurrentHashMap<IntPair, ChunkLock> chunks = SENDING_CHUNKS.computeIfAbsent(
-                worldName,
-                world -> new ConcurrentHashMap<>()
-        );
+        ConcurrentHashMap<IntPair, ChunkSendLock> chunks = FaweBukkitWorld.getWorldSendingChunksMap(worldName);
         chunks.compute(pair, (k, lock) -> {
             if (lock == null) {
-                lock = new ChunkLock();
+                lock = new ChunkSendLock();
             } else if (lock.writeWaiting) {
                 throw new IllegalStateException("Attempting to write chunk section when write is already ongoing?!");
             }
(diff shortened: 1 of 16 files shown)


NotMyFault (approved):

dordsor21:
fwiw I threw this at someone who experienced fewer issues after installing

worldedit-bukkit/src/main/java/com/fastasyncworldedit/bukkit/adapter/NMSAdapter.java
```diff
@@ -140,4 +144,122 @@ public void sendChunk(IChunkGet chunk, int mask, boolean lighting) {
...
+            lock.writeWaiting = true;
+            return lock;
+        });
+        try {
+            if (stamp[0] == 0) {
+                stamp[0] = chunkLock[0].lock.writeLock();
+            }
+            return ReflectionUtils.compareAndSet(sections, expected, value, layer);
+        } finally {
+            chunks = SENDING_CHUNKS.get(worldName);
```
SirYwell, line 195: is there any scenario where this is actually needed? The old reference should still be the correct one as we never remove it. Alternatively, did you consider using a `ConcurrentHashMap<*WorldIntPair*, ChunkLock>` instead? This way we won't have unused CHM instances laying around (which could probably be a problem with many temporary worlds).

worldedit-bukkit/src/main/java/com/fastasyncworldedit/bukkit/adapter/NMSAdapter.java
```diff
@@ -140,4 +144,122 @@ public void sendChunk(IChunkGet chunk, int mask, boolean lighting) {
...
+            LevelChunkSection[] sections,
+            LevelChunkSection expected,
+            LevelChunkSection value,
+            int layer
+    ) {
+        if (layer < 0 || layer >= sections.length) {
+            return false;
+        }
+        long[] stamp = new long[1];
+        ChunkLock[] chunkLock = new ChunkLock[1];
```
SirYwell, line 173: Could these two arrays be replaced by a class? Iirc arrays have 16 byte object headers alone in current Hotspot versions, so reducing to a class would significantly cut down memory per call here (assuming it doesn't get scalar replaced, which is rather unlikely due to the complexity of the involved code).

worldedit-bukkit/src/main/java/com/fastasyncworldedit/bukkit/adapter/NMSAdapter.java
```diff
@@ -140,4 +144,122 @@ public void sendChunk(IChunkGet chunk, int mask, boolean lighting) {
...
+    /**
+     * Releases the read lock acquired when sending a chunk packet for a chunk
+     *
+     * @since TODO
+     */
+    protected static void endChunkPacketSend(String worldName, IntPair pair, long stamp, StampedLock stampedLock) {
+        ConcurrentHashMap<IntPair, ChunkLock> chunks = SENDING_CHUNKS.get(worldName);
+        chunks.computeIfPresent(pair, (k, lock) -> {
+            if (lock.lock != stampedLock) {
+                throw new IllegalStateException("SENDING_CHUNKS stored lock is not ");
```
SirYwell, line 250: seems like the message got messed up

SirYwell (changes requested):

worldedit-bukkit/src/main/java/com/fastasyncworldedit/bukkit/adapter/NMSAdapter.java
dordsor21: The idea was to reduce lock contention between worlds when locking/unlocking. I suppose the operations themselves shouldn't take too long, though perhaps it's worth having the CHM in a BukkitWorld instance so it's GCed as needed if still attempting to do this

SirYwell: Moving it into BukkitWorld sounds good.

dordsor21: FAWE does not actually cache/limit the number of BukkitWorld instances, and given upstream does not either, it feels quite wrong to do this. I don't know if it's worth wrapping a WeakHashMap into a synchronised map and using the NMS world (or a normal bukkit world) as a key?

SirYwell (approved):

CyberFlameGO (approved):

PR #2867, opened 2024-08-01
feat: alter clipboard thread to move all clipboard loading from main thread

dordsor21:


dordsor21:
Currently untested

worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/WorldEditPlugin.java
```diff
@@ -565,14 +567,15 @@ public BukkitPlayer wrapPlayer(Player player) {
...
+            Supplier<BukkitPlayer> task = () -> {
+                BukkitPlayer bukkitPlayer = getCachedPlayer(player);
+                if (bukkitPlayer == null) {
+                    bukkitPlayer = new BukkitPlayer(this, player);
+                    player.setMetadata("WE", new FixedMetadataValue(this, bukkitPlayer));
                 }
-            }
+                return bukkitPlayer;
+            };
+            TaskManager.taskManager().sync(task);
```
SirYwell, line 578: This can cause deadlocks when the server thread is waiting for the current thread. What's the intention behind it?

worldedit-core/src/main/java/com/fastasyncworldedit/core/Fawe.java
```diff
@@ -140,14 +147,27 @@ private Fawe(final IFawe implementation) {
...
                 TimeUnit.MILLISECONDS,
                 new LinkedBlockingQueue<>(),
-                new ThreadFactoryBuilder().setNameFormat("FAWE Clipboard - %d").build()
+                new ThreadFactoryBuilder().setNameFormat("FAWE UUID-key-queued - %d").setThreadFactory(new ThreadFactory() {
+                    private final ThreadFactory defaultFactory = Executors.defaultThreadFactory();
+
+                    @Override
+                    public Thread newThread(@Nonnull Runnable r) {
+                        return defaultFactory.newThread(() -> {
+                            IS_UUID_KEYED_EXECUTOR_THREAD.set(true);
```
SirYwell, line 162: What's the idea here? I'd assume every thread in this thread pool is always a uuid keyed executor thread? Would it be enough to subtype `Thread` and do an instanceof check where needed?

worldedit-core/src/main/java/com/sk89q/worldedit/extension/platform/AbstractPlayerActor.java
```diff
@@ -520,22 +531,71 @@ public Region getLargestRegion() {
...
+            default -> new CuboidRegionSelector(null, region.getMinimumPoint(), region.getMaximumPoint());
+        };
         selector.setWorld(region.getWorld());
 
         getSession().setRegionSelector(getWorld(), selector);
     }
+
+    @Override
+    public void loadClipboardFromDisk() {
+        if (clipboardLoading.isWriteLocked()) {
```
SirYwell, line 547: It looks like we only use the write lock. Can't we use a different lock then (or a `Semaphore` if you want to unlock from a different thread)? Besides that, I think there might be a race condition where the write lock is not locked at this call but below the if. If that happens on the main thread, we stall the main thread.

SirYwell (changes requested):

worldedit-core/src/main/java/com/fastasyncworldedit/core/Fawe.java
dordsor21: It needs to be accessible from runners within the thread itself. I didn't want to offload to an async thread from main thread needlessly to then run on this thread again. I suppose checking if the current thread is an instance of subtype would be cleaner

worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/WorldEditPlugin.java
dordsor21: The task manager method here will run the given task immediately if it's on the main thread

SirYwell: Yes but it doesn't know if the main thread is blocked by the current thread

worldedit-core/src/main/java/com/fastasyncworldedit/core/util/task/UUIDKeyQueuedThreadFactory.java
```diff
@@ -0,0 +1,38 @@
+package com.fastasyncworldedit.core.util.task;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.atomic.AtomicInteger;
+
+public class UUIDKeyQueuedThreadFactory implements ThreadFactory {
```
SirYwell, line 8: Might make sense to mark this as internal?

SirYwell (approved):
Seems to work

worldedit-core/src/main/java/com/sk89q/worldedit/extension/platform/AbstractPlayerActor.java
```diff
@@ -520,22 +531,72 @@ public Region getLargestRegion() {
...
+            }
+            return;
+        }
+
+        File file = MainUtil.getFile(
+                Fawe.platform().getDirectory(),
+                Settings.settings().PATHS.CLIPBOARD + File.separator + getUniqueId() + ".bd"
+        );
+        try {
+            Future<?> fut = Fawe.instance().submitUUIDKeyQueuedTask(getUniqueId(), () -> {
```
PierreSchwang, line 564: Is it really wanted to enqueue twice? `LocalSession#loadCLipboardFromDisk(File)` submits a task in the KeyQueuedExecutorService - or is the threading purely for acquiring / synchronizing on the lock?

worldedit-core/src/main/java/com/sk89q/worldedit/extension/platform/AbstractPlayerActor.java
```diff
@@ -520,22 +531,72 @@ public Region getLargestRegion() {
...
+                return;
+            }
+            fut.get();
+        } catch (Exception e) {
+            print(Caption.of("fawe.error.clipboard.invalid"));
+            LOGGER.error("Error loading clipboard form disk", e);
+            print(Caption.of("fawe.error.stacktrace"));
+            print(Caption.of("fawe.error.no-failure"));
+            print(Caption.of("fawe.error.clipboard.invalid.info", file.getName(), file.length()));
+            print(Caption.of("fawe.error.stacktrace"));
```
PierreSchwang, line 597: Those errors are not really fitting I'd say - the only exception(s) can be thrown by either enqueuing the task or waiting for the completion (e.g. interrupted). I guess a more-generic error would be more fitting, as nothing is wrong with the clipboard itself.

worldedit-core/src/main/java/com/sk89q/worldedit/entity/Player.java
```diff
@@ -423,7 +423,7 @@ default void unregister() {
         if (Settings.settings().CLIPBOARD.USE_DISK && Settings.settings().CLIPBOARD.DELETE_ON_LOGOUT) {
             session.deleteClipboardOnDisk();
         } else if (Settings.settings().CLIPBOARD.USE_DISK) {
-            Fawe.instance().getClipboardExecutor().submit(getUniqueId(), () -> session.setClipboard(null));
+            Fawe.instance().submitUUIDKeyQueuedTask(getUniqueId(), () -> session.setClipboard(null));
```
PierreSchwang, line 426: Kinda the same as my first comment, `#setClipboard` has basically no work except for synchronizing on the lock - is enqueuing that worth it? Might be required due to the lock holder thread, but that would require `getClipboard` calls to be enqueued as well, which they aren't.

PierreSchwang (approved):
Seems to work quite nicely

worldedit-core/src/main/java/com/sk89q/worldedit/extension/platform/AbstractPlayerActor.java
dordsor21: That's done on purpose as we want to ensure that we offload from the main thread in AbstractPlayerActor, but without creating a new thread, or using up a thread in an unrelated thread pool. Since we will end up resubmitting to the queue pool again (which results in immediately running the task) we don't lose anything, but only gain safety in offloading in two places

worldedit-core/src/main/java/com/sk89q/worldedit/entity/Player.java
dordsor21: Yeah so basically "free" synchronization from the main thread without having to wait for any potentially ongoing clipboard loading (players joining and leaving a server instantaneously could otherwise lag the server)

PR #2883, opened 2024-08-18
Avoid deprecated CompoundTag in API

SirYwell:
## Description

`CompoundTag` is deprecated and its implementation is now backed by `LinCompoundTag`. This causes significant overhead when access its data due to the needed transformation/copy.

This PR aims to address spots where `CompoundTag` is used in FAWE api, as well as some uses that can benefit from the reduced overhead. There is more to do to get `CompoundTag` fully out of FAWE, but more changes can be addressed later.

```[tasklist]
### Submitter Checklist
```


NotMyFault (approved):

dordsor21:
Was moving to jetbrains annotations from javax done on purpose? I believe we're relatively consistent for the latter

NotMyFault:
> Was moving to jetbrains annotations from javax done on purpose? I believe we're relatively consistent for the latter

JavaX is superseded by Jakarta. A migration away from JavaX is a good choice, whether it's its successor or a replacement.

dordsor21:
> > Was moving to jetbrains annotations from javax done on purpose? I believe we're relatively consistent for the latter
> 
> JavaX is superseded by Jakarta. A migration away from JavaX is a good choice, whether it's its successor or a replacement.

Should probably make a decision and move all of them at once imo

SirYwell:
> Was moving to jetbrains annotations from javax done on purpose? I believe we're relatively consistent for the latter

No, it's mostly because IntelliJ likes to use its annotations by default. I switched to javax for now, but we might want to consider switching to jspecify or checkerframework in future.

dordsor21 (approved):

PR #2898, opened 2024-08-26
Faster undo operations

SirYwell:
## Description

`//undo` is inherently single-threaded due to the fact that changes are read from input streams. However, we can split the reading work (producer) from the writing work (consumer) so that it only takes the time the slower of those two operations requires. To reduce allocations, FAWE uses mutable changes, which further complicates things as we need to make sure to not write to a change that is yet to be read. Having an exchanger dealing with two arrays makes that simple.

From my testing, this means the slower action is reading from the input streams, actual improvements might heavily depend on hardware (cpu, memory, disk). 

```[tasklist]
### Submitter Checklist
```

### DIFF
--- a/worldedit-core/src/main/java/com/fastasyncworldedit/core/configuration/Settings.java
+++ b/worldedit-core/src/main/java/com/fastasyncworldedit/core/configuration/Settings.java
@@ -623,6 +623,13 @@ public class Settings extends Config {
     public static class EXPERIMENTAL {
 
         @Comment({
+                "Undo operation batch size",
+                " - The size defines the number of changes read at once.",
+                " - Larger numbers might reduce overhead but increase latency for edits with only few changes.",
+                " - 0 means undo operations are not batched."})
+        public int UNDO_BATCH_SIZE = 128;
+
+        @Comment({
                 "[UNSAFE] Directly modify the region files. (OBSOLETE - USE ANVIL COMMANDS)",
                 " - IMPROPER USE CAN CAUSE WORLD CORRUPTION!",
         })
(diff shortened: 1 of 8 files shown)


NotMyFault (approved):

worldedit-core/src/main/java/com/fastasyncworldedit/core/configuration/Settings.java
```diff
@@ -622,6 +622,13 @@ public static class PROGRESS {
     })
     public static class EXPERIMENTAL {
 
+        @Comment({
+                "Undo operation batch size",
+                " - The size defines the number of changes read at once.",
+                " - Larger numbers might reduce overhead but increase latency for edits with only few changes.",
+                " - 0 means undo operations are not batched."})
+        public int UNDO_BATCH_SIZE = 128;
```
dordsor21, line 630: We should know history size before undoing, so we could dynamically set this?

SirYwell: I'm not sure if dynamically setting it has much benefit, it's basically just a tradeoff between cross-thread-communication and memory overhead. Not sure if there is any good way to find an appropriate value for it.

dordsor21: I suppose if people care that much they can find the best value for themselves, otherwise, I would assume very large edits where this would matter more is going to be more affected by chunk loading times anyway

dordsor21 (approved):

PR #2899, opened 2024-08-27
Fix: Copy-Paste-Brush with falsy mask

PierreSchwang:
## Overview
Fixes a bug shared on discord. 

## Description
Use copypaste-brush (`//br copypaste`) with a falsy mask (e.g. `//gmask 0`) and select the area via the first click / interact. Output should state, that `0 blocks were copied.`. Attempting to paste with that brush, results in no future actions being possible (potentially depending on the queue type). 

That's due to the `ResizableClipboardBuilder`. As the mask did not match any blocks, no blocks were added to the builder. Therefor, on build the default dimensions are used, which results in the region spanning from XYZ Integer.MAX_VALUE to XYZ Integer.MIN_VALUE. The RegionVisitor - called by the ForwardExtentCopy - attempts to load all chunks in that range via the SingleThreadQueueExtent. 

The fix is simply using an `EmptyClipboard` if nothing was being written into the `ResizableClipboardBuilder`.

```[tasklist]
### Submitter Checklist
```


SirYwell:
I think I'd feel more comfortable with the check being in the `ResizableClipboardBuilder#build()` instead. It's currently not used anywhere else in FAWE, but that's also nothing we ever want to stumble upon again

SirYwell (approved):

NotMyFault (approved):

PR #2901, opened 2024-08-29
fix: allow webinterface schematic format to be detected

### LINKED ISSUE #2892: Error on loading schematics v2 from URL
### Server Implementation

Paper

### Server Version

1.20.4

### Describe the bug

Loading schematics v2 fails since v3 support was introduced in https://github.com/IntellectualSites/FastAsyncWorldEdit/pull/2776.

### To Reproduce

1. Upload the [schematic v2](https://pixeldrain.com/u/MUQZE1uv) to the [website](https://schem.intellectualsites.com/fawe/index.php).
2. Try to load it with `//schem load url:<code>`.

### Expected behaviour

Schematic should load successfully.

### Screenshots / Videos

_No response_

### Error log (if applicable)

https://paste.gg/p/anonymous/63215063001843d697981a832ed6c50e

### Fawe Debugpaste

https://athion.net/ISPaster/paste/view/a49a651595e4489dbb9b5c7935821283

### Fawe Version

2.11.1

### Checklist

### Anything else?

It seems loading the same schematic from file `//schem load <schem>` doesn't cause any errors.

PierreSchwang:
## Overview
Fixes #2892

## Description
Not completely happy with how it works - but that's just a limitation of the interface itself.
When not specifying the format, it'll be auto detected. Sadly, the interface outputs the command with the `schem` format specified (which now defaults to v3). But I've updated the exception handling to state that, if the schematic fails to load (the EOFException triggered, when loading a v2 schematic with the default `schem` was handled by the IOException block):
image

As I'd have to download the file either way when checking for the format, it's simply downloaded even if the format is specified. That way it can utilize the reset functionality of the underlying stream for the reader (v3 in particular). The Closer handles the deletion either way - so it shouldn't clog up anything nor be noticeable in any way.

```[tasklist]
### Submitter Checklist
```

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/command/SchematicCommands.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/command/SchematicCommands.java
@@ -76,12 +76,9 @@ import java.io.FileInputStream;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.OutputStream;
 import java.net.URI;
 import java.net.URL;
-import java.nio.channels.Channels;
-import java.nio.channels.FileChannel;
-import java.nio.channels.ReadableByteChannel;
-import java.io.OutputStream;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
@@ -94,8 +91,8 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.UUID;
 import java.util.concurrent.Callable;
-import java.util.function.Consumer;
 import java.util.concurrent.ThreadLocalRandom;
+import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.regex.Pattern;
 
@@ -328,7 +325,6 @@ public class SchematicCommands {
         LocalConfiguration config = worldEdit.getConfiguration();
 
         //FAWE start
-        final Closer closer = Closer.create();
         ClipboardFormat format;
         InputStream in;
         // if format is set explicitly, do not look up by extension!
@@ -336,7 +332,7 @@ public class SchematicCommands {
         if (noExplicitFormat) {
             formatName = "fast";
         }
-        try {
+        try(final Closer closer = Closer.create()) {
             URI uri;
             if (formatName.startsWith("url:")) {
                 String t = filename;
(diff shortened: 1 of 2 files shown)


PierreSchwang:
I thought about adding a check if the format `schem` is specified (as the schematic spits the command out with that format explicitly set) and check the format either way - but that felt quite wrong

Feniksovich:
> I thought about adding a check if the format `schem` is specified (as the schematic interface spits the command out with that format explicitly set) and check the format either way - but that felt quite wrong

I agree with that, but I'm also afraid most users don't pay so much attention to commands syntax, what the each argument means or what schematics schemas actually exist (I'm not sure if the syntax of `/schem` command is documented anywhere).

For now, on my end I monitor `/schem` command execution and display the message with exact instructions and commands that can be executed to workaround current loading behavior (e.g. suggesting try `/schem load fast.2 url:<code>`).

PierreSchwang:
> > I thought about adding a check if the format `schem` is specified (as the schematic interface spits the command out with that format explicitly set) and check the format either way - but that felt quite wrong
> 
> I agree with that, but I'm also afraid most users don't pay so much attention to commands syntax, what the each argument means or what schematics schemas actually exist (I'm not sure if the syntax of `/schem` command is documented anywhere).
> 
> For now, on my end I monitor `/schem` command execution and display the message with exact instructions and commands that can be executed to workaround current loading behavior (e.g. suggesting try `/schem load fast.2 url:<code>`).

I guess it's already enough, that the more precise error is printed, when specifying the wrong format. That should help a bit - maybe (given the error is read)

worldedit-core/src/main/java/com/sk89q/worldedit/command/SchematicCommands.java
```diff
@@ -325,8 +328,9 @@ public void load(
         LocalConfiguration config = worldEdit.getConfiguration();
 
         //FAWE start
+        final Closer closer = Closer.create();
```
SirYwell, line 331: Can this be moved into  the outer try as a resource? Then the closing doesn't need to be handled explicitly

worldedit-core/src/main/java/com/sk89q/worldedit/command/SchematicCommands.java
```diff
@@ -346,11 +350,32 @@ public void load(
...
-                uri = url.toURI();
+                final Path temp = Files.createTempFile("faweremoteschem", null);
+                final File tempFile = temp.toFile();
+                // delete temporary file when we're done
+                closer.register((Closeable) () -> Files.deleteIfExists(temp));
+                // write schematic into temporary file
+                try (final ReadableByteChannel byteChannel = Channels.newChannel(url.openStream());
+                     final FileOutputStream out = new FileOutputStream(tempFile);
+                     final FileChannel outChannel = out.getChannel()) {
+                    outChannel.transferFrom(byteChannel, 0, Long.MAX_VALUE);
```
SirYwell, line 370: `InputStream` also has a `transferTo` method, I think we could skip channels (and probably `File`?) here

SirYwell (changes requested):

worldedit-core/src/main/java/com/sk89q/worldedit/command/SchematicCommands.java
PierreSchwang: The file is still needed - once for actually having an InputStream, and for `ClipboardFormats.findByFile` which requires the file

dordsor21 (approved):

PR #2902, opened 2024-08-30
fix: don't process out of bound section while trimming Y sections

### LINKED ISSUE #2894: Region Restrictions incorrectly protect regions
### Server Implementation

Paper

### Server Version

1.21

### Describe the bug

When creating a region restriction, if the y axis of your starting coordinate is 64, 32, 16, or 0, it makes your region bigger than intended. For example, if your region starts at y=64, users will be able to place blocks up to y=48. while i haven't completely tested this, it seemingly only happens when the y axis of the chunk coordinate is 0.

When using `/wer` to select the region you have access to, it doesn't select outside the intended region. But if you manually select outside of your region, it will let you place outside the intended area.

### To Reproduce

1. Create a FaweMask. Make sure it starts at the yAxis of 64. In my case, I'm using this simple code.
```kt
class MaskManager(plugin: Sandbox) : FaweMaskManager(plugin.name) {

    val yAxis = 64

    override fun getMask(player: Player?, type: MaskType?, isWhitelist: Boolean): FaweMask {
        val region = CuboidRegion(
            BlockVector3.at(20, yAxis, 20),
            BlockVector3.at(30, 100, 30)
        )
        return FaweMask(region)
    }

}
```
```kt
class Sandbox : JavaPlugin() {

    override fun onEnable() {
        FaweAPI.addMaskManager(MaskManager(this))
    }

}
```
2. give yourself the permissions `fawe.permpack.basic` and `fawe.[plugin name]`
3. run /wer to select your region.
4. do `//set stone` and watch worldedit placing a cube with coordinates (20, 64, 20) (30, 100, 30)
5. now run `//expand vert` and `//set diamond_block` and observe worledit placing a cube with coordinates (20, 48, 20) (30, 100, 30)
6. you have now placed blocks outside of the intended region
7. (optional) try the same thing while setting the yAxis variable to 65 or 63. the behavior i described will not happen and the region will be correct

### Expected behaviour

I would expect my region to be 1:1 to what i coded!

### Screenshots / Videos

I've sent a video in the IntellectualSites discord.
 https://discord.com/channels/268444645527126017/344128526435221505/1277358741431390322

### Error log (if applicable)

_No response_

### Fawe Debugpaste

https://athion.net/ISPaster/paste/view/32d2c89b00ed4e24af46a83e80177658

### Fawe Version

2.11.2-SNAPSHOT-886;3b4e849

### Checklist

### Anything else?

_No response_

PierreSchwang:
## Overview
Fixes #2894

## Description
trimY now doesn't overflow into unrelated sections anymore, allowing out of bounds editing. Seems to work for me so far - tests are passing. I couldn't find any edge cases or how that might break other existing behavior. 

I'd highly appreciate additional tests by others.

```[tasklist]
### Submitter Checklist
```

### DIFF
--- a/worldedit-core/src/main/java/com/fastasyncworldedit/core/queue/IBatchProcessor.java
+++ b/worldedit-core/src/main/java/com/fastasyncworldedit/core/queue/IBatchProcessor.java
@@ -65,45 +65,38 @@ public interface IBatchProcessor {
         int minLayer = minY >> 4;
         int maxLayer = maxY >> 4;
         if (keepInsideRange) {
-            for (int layer = set.getMinSectionPosition(); layer <= minLayer; layer++) {
-                if (set.hasSection(layer)) {
-                    if (layer == minLayer) {
-                        char[] arr = set.loadIfPresent(layer);
-                        if (arr != null) {
-                            int index = (minY & 15) << 8;
-                            for (int i = 0; i < index; i++) {
-                                arr[i] = BlockTypesCache.ReservedIDs.__RESERVED__;
-                            }
-                            set.setBlocks(layer, arr);
-                        }
-                    } else {
-                        set.setBlocks(layer, null);
+            for (int layer = set.getMinSectionPosition(); layer <= set.getMaxSectionPosition(); layer++) {
+                if (!set.hasSection(layer)) {
+                    continue;
+                }
+                // wipe all data from chunk layers above or below the max / min layer
+                if (layer < minLayer || layer > maxLayer) {
+                    set.setBlocks(layer, null);
+                    continue;
+                }
+                // if chunk layer / section is fully enclosed by minY to maxY, keep as is
+                if (layer > minLayer && layer < maxLayer) {
+                    continue;
+                }
+                char[] blocks = set.loadIfPresent(layer);
+                if (blocks == null) {
+                    continue;
+                }
+                // When on the minimum layer (as defined by minY), remove blocks up to minY (exclusive)
+                if (layer == minLayer) {
+                    int index = (minY & 15) << 8;
+                    for (int i = 0; i < index; i++) {
+                        blocks[i] = BlockTypesCache.ReservedIDs.__RESERVED__;
                     }
                 }
-            }
-            for (int layer = maxLayer; layer <= set.getMaxSectionPosition(); layer++) {
-                if (set.hasSection(layer)) {
-                    if (layer == maxLayer) {
-                        char[] arr = set.loadIfPresent(layer);
-                        if (arr != null) {
-                            /*
-                              If maxY is the last coordinate in a chunk section (`2^n-1`), it bleeds into
-                              the next higher section and represents the 0th block on the y-axis on that section. This results
-                              in the index being 0 (so the whole section is wiped) - even though only the last slice should be
-                              overwritten.
-                              If maxY is any other coordinate (0 - 14) `((maxY + 1) & 15) << 8` is evaluated, resulting in
-                              getting the index for the next (upper) slice of this section (as this one should be overwritten)
-                             */
-                            int index = (maxY + 1) % 16 == 0 ? arr.length : ((maxY + 1) & 15) << 8;
-                            for (int i = index; i < arr.length; i++) {
-                                arr[i] = BlockTypesCache.ReservedIDs.__RESERVED__;
-                            }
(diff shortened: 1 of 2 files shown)


worldedit-core/src/main/java/com/fastasyncworldedit/core/queue/IBatchProcessor.java
```diff
@@ -62,8 +62,8 @@ default boolean processGet(int chunkX, int chunkZ) {
      * @return false if chunk is empty of blocks
      */
     default boolean trimY(IChunkSet set, int minY, int maxY, final boolean keepInsideRange) {
-        int minLayer = (minY - 1) >> 4;
-        int maxLayer = (maxY + 1) >> 4;
+        int minLayer = minY >> 4;
+        int maxLayer = maxY >> 4;
```
PierreSchwang, line 66: It would theoretically work with `maxY + 1` when reverting the change in L97 (the ternary) - but that does not really feel wanted.
Currently, when crossing the chunk section border on the upper end (e.g. Y 255), the maxLayer is the section from 256-271. That still is fine and works - the layer of the section representing `maxY` is set to `__RESERVED__` as it should - but the chunk section from 256-271 is set to 4096 chars only containing __RESERVED__ instead of just being set to null.
With that change, that now is more consistent. In the previous example, the section is filled with the correct blocks from Y 240-254 and 255 is set to __RESERVED__. The section 256-271 is set to null.

SirYwell: I think it would make sense to clarify the meaning of `minLayer` and `maxLayer`. From my understanding, `minLayer` is the highest layer that should be partially trimmed below, and `maxLayer` is the lowest layer that should be partially trimmed above.

The edge cases here are, when not adding/subtracting before the shift, when `minY % 16 == 0` (the full section should *not* be trimmed) or `maxY % 16 == 15` (the full section should *not* be trimmed,).

In the minLayer case, that's easy because `0 << 8 == 0`, so the loop won't run.
In the maxLayer case, `15 << 8 == 3840`, means we would clear one layer too early. But `(15 + 1) % 16 == 0`, so adding the layer this way results in the whole section being wiped. But I think moving the addition should work: `15 % 16 == 15`,`(15 + 1) << 8 == 4096`, so `((maxX & 15) + 1) << 8` should simplify the maxLayer case.

That's at least my understanding, maybe I missed something.

worldedit-core/src/main/java/com/fastasyncworldedit/core/queue/IBatchProcessor.java
```diff
@@ -86,7 +86,15 @@ default boolean trimY(IChunkSet set, int minY, int maxY, final boolean keepInsid
...
-                            int index = ((maxY + 1) & 15) << 8;
+                            /*
+                              If maxY is the last coordinate in a chunk section (`2^n-1`), it bleeds into
+                              the next higher section and represents the 0th block on the y-axis on that section. This results
+                              in the index being 0 (so the whole section is wiped) - even though only the last slice should be
+                              overwritten.
+                              If maxY is any other coordinate (0 - 14) `((maxY + 1) & 15) << 8` is evaluated, resulting in
+                              getting the index for the next (upper) slice of this section (as this one should be overwritten)
+                             */
+                            int index = (maxY + 1) % 16 == 0 ? arr.length : ((maxY + 1) & 15) << 8;
```
dordsor21, line 97: Is there a reason for using modulus rather than a bit mask here?

PierreSchwang: just my brain being fried

worldedit-core/src/main/java/com/fastasyncworldedit/core/queue/IBatchProcessor.java
PierreSchwang: I've simplified the logic itself a bit which should make the whole thing easier to understand I'd say? (At least it does for me). And yeah, the simplification should work (and does based on my testing).

If the new loop logic is alright, I could (and should) apply that to the other part of the trimY method (where `keepInsideRange` is false)

SirYwell (approved):
Looks good now.

dordsor21 (approved):

PR #2906, opened 2024-09-02
Avoid map lookup in `isMovementBlocker` calls

SirYwell:
## Description

`isMovementBlocker` is called in the `HeightmapProcessor` for every edited column at least once, so it's kind of a hot method. On newer Spigot versions, the call `Material#isSolid()` however requires a map lookup as it goes through the registry, slowing down the method call unnecessarily. As we already have the NMS block state available, we can just take the shortcut and call the method that is used by Spigot (and by the NMS heightmap calculation) directly.

```[tasklist]
### Submitter Checklist
```


dordsor21 (approved):

PierreSchwang (approved):
feels reasonable

PR #2913, opened 2024-09-15
fix: adjust linked filter to be left-right and do not link to new forked instances

### LINKED ISSUE #1874: Bad calculation of the number of blocks changed?

dordsor21:
- Assume that child filters know about their own forks
 - fixes #1874

### DIFF
--- a/worldedit-core/src/main/java/com/fastasyncworldedit/core/extent/filter/LinkedFilter.java
+++ b/worldedit-core/src/main/java/com/fastasyncworldedit/core/extent/filter/LinkedFilter.java
@@ -1,69 +1,102 @@
 package com.fastasyncworldedit.core.extent.filter;
 
-import com.fastasyncworldedit.core.extent.filter.block.DelegateFilter;
 import com.fastasyncworldedit.core.extent.filter.block.FilterBlock;
 import com.fastasyncworldedit.core.internal.simd.VectorizedFilter;
 import com.fastasyncworldedit.core.queue.Filter;
+import com.fastasyncworldedit.core.queue.IChunk;
+import com.sk89q.worldedit.regions.Region;
 import jdk.incubator.vector.ShortVector;
+import org.jetbrains.annotations.Nullable;
 
 /**
- * Filter which links two Filters together for single-filter-input operations.
+ * Filter which links two Filters together for single-filter-input operations. Left filter is operated first.
  *
- * @param <T> Parent which extends Filter
- * @param <S> Child which extends Filter
+ * @param <L> Left filter
+ * @param <R> Right filter
  */
-public sealed class LinkedFilter<T extends Filter, S extends Filter> extends DelegateFilter<T> {
+public sealed class LinkedFilter<L extends Filter, R extends Filter> implements Filter {
 
-    private final S child;
+    private final L left;
+    private final R right;
+
+    public LinkedFilter(L left, R right) {
+        this.left = left;
+        this.right = right;
+    }
 
     @SuppressWarnings({"unchecked", "rawtypes"}) // we defeated the type system
-    public static <T extends Filter, S extends Filter> LinkedFilter<? extends T, ? extends S> of(T parent, S child) {
-        if (parent instanceof VectorizedFilter p && child instanceof VectorizedFilter c) {
-            return new VectorizedLinkedFilter(p, c);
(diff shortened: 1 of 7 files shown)


worldedit-core/src/main/java/com/fastasyncworldedit/core/extent/filter/LinkedFilter.java
```diff
@@ -1,69 +1,102 @@
...
+    public <T extends IChunk> T applyChunk(T chunk, @Nullable Region region) {
+        chunk = getLeft().applyChunk(chunk, region);
+        return getRight().applyChunk(chunk, region);
     }
 
-    public S getChild() {
-        return this.child;
+    @Override
+    public boolean appliesLayer(IChunk chunk, int layer) {
+        return getLeft().appliesLayer(chunk, layer) && getRight().appliesLayer(chunk, layer);
```
SirYwell, line 56: This should be `||`, shouldn't it?

worldedit-core/src/main/java/com/fastasyncworldedit/core/extent/filter/LinkedFilter.java
```diff
@@ -1,69 +1,102 @@
...
+    public boolean appliesChunk(int chunkX, int chunkZ) {
+        return getLeft().appliesChunk(chunkX, chunkZ) && getRight().appliesChunk(chunkX, chunkZ);
     }
 
-    public LinkedFilter(T parent, S child) {
-        super(parent);
-        this.child = child;
+    @Override
+    public <T extends IChunk> T applyChunk(T chunk, @Nullable Region region) {
+        chunk = getLeft().applyChunk(chunk, region);
```
SirYwell, line 50: `chunk` could be null after the call according to the docs, would it make sense to pass the original value to the right in that case?

worldedit-core/src/main/java/com/fastasyncworldedit/core/extent/filter/LinkedFilter.java
```diff
@@ -1,69 +1,102 @@
...
+    public L getLeft() {
+        return this.left;
+    }
+
+    public R getRight() {
+        return this.right;
+    }
+
+    @Override
+    public boolean appliesChunk(int chunkX, int chunkZ) {
```
SirYwell, line 44: Same as for `appliesLayer`

SirYwell (changes requested):

worldedit-core/src/main/java/com/fastasyncworldedit/core/extent/filter/LinkedFilter.java
dordsor21: I wasn't really sure what was best. Do we want to still filter in one if the other doesn't want to? That implies that edit shouldn't be happening in the layer

SirYwell: My understanding was that the methods basically mean return whether a filter *may* do something with the chunk/layer. Now if we say "right only needs to do something if left did something", then AND should be fine, but otherwise not. There's also a theoretical scenario where right can only decide whether it needs to do something *after* left already did its work. We currently don't have any custom implementations for those methods (the default returns true, other impls just delegate), so maybe it would make sense to just remove them? And if we ever think that something like that might be useful, we can reintroduce it and think about it again, having an actual example at hand.

dordsor21: Mm that might be the best solution for now, setting the return for applyChunk to nonnull I suppose

SirYwell (approved):

NotMyFault (approved):

PR #2927, opened 2024-09-23
Implement vectorization of #existing and inverse masks

SirYwell:
## Description

`#existing` is e.g. implicitly used by `//replace` with only one argument. It's basically a "not air". It is also used in `#air` but negated, so supporting `InverseMask`s were the original mask can be vectorized is simple and useful.

```[tasklist]
### Submitter Checklist
```

### DIFF
--- a/worldedit-core/src/main/java/com/fastasyncworldedit/core/internal/simd/SimdSupport.java
+++ b/worldedit-core/src/main/java/com/fastasyncworldedit/core/internal/simd/SimdSupport.java
@@ -2,14 +2,17 @@ package com.fastasyncworldedit.core.internal.simd;
 
 import com.fastasyncworldedit.core.configuration.Settings;
 import com.fastasyncworldedit.core.extent.filter.block.DelegateFilter;
+import com.fastasyncworldedit.core.function.mask.InverseMask;
 import com.fastasyncworldedit.core.function.mask.SingleBlockStateMask;
 import com.fastasyncworldedit.core.queue.Filter;
+import com.sk89q.worldedit.function.mask.ExistingBlockMask;
 import com.sk89q.worldedit.function.mask.InverseSingleBlockStateMask;
 import com.sk89q.worldedit.function.mask.Mask;
 import com.sk89q.worldedit.function.pattern.Pattern;
 import com.sk89q.worldedit.internal.util.LogManagerCompat;
 import com.sk89q.worldedit.world.block.BaseBlock;
 import com.sk89q.worldedit.world.block.BlockStateHolder;
+import com.sk89q.worldedit.world.block.BlockTypesCache;
 import jdk.incubator.vector.ShortVector;
 import jdk.incubator.vector.VectorOperators;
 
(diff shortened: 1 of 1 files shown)


worldedit-core/src/main/java/com/fastasyncworldedit/core/internal/simd/SimdSupport.java
```diff
@@ -44,10 +47,23 @@ public static boolean useVectorApi() {
...
+                }
+                yield (set, get) -> base.compareVector(set, get).not();
+            }
             default -> null;
         };
     }
 
+    private static VectorizedMask vectorizedTargetMaskNonAir() {
+        // everything > VOID_AIR is not air
+        return (set, get) -> get.compare(VectorOperators.UNSIGNED_GT, BlockTypesCache.ReservedIDs.VOID_AIR);
```
dordsor21, line 64: This should just be AIR

worldedit-core/src/main/java/com/fastasyncworldedit/core/internal/simd/SimdSupport.java
```diff
@@ -44,10 +47,23 @@ public static boolean useVectorApi() {
         return switch (mask) {
             case SingleBlockStateMask single -> vectorizedTargetMask(single.getBlockState().getOrdinalChar());
             case InverseSingleBlockStateMask inverse -> vectorizedTargetMaskInverse(inverse.getBlockState().getOrdinalChar());
+            case ExistingBlockMask ignored -> vectorizedTargetMaskNonAir();
+            case InverseMask inverse -> {
+                final VectorizedMask base = vectorizedTargetMask(inverse.inverse());
```
dordsor21, line 52: Any there any/many inverse operations that would be too intensive for this to be efficient?

worldedit-core/src/main/java/com/fastasyncworldedit/core/internal/simd/SimdSupport.java
SirYwell: `ExistingBlockMask` covers all blocks that are considered air, including cave_air and void_air. I'm basically copying that behavior here

dordsor21: AIr _can_ be void or cave sometimes, but afaik it never is and is always just AIR

worldedit-core/src/main/java/com/fastasyncworldedit/core/internal/simd/SimdSupport.java
dordsor21: Ah yeah I see

worldedit-core/src/main/java/com/fastasyncworldedit/core/internal/simd/SimdSupport.java
SirYwell: cave air is used in structures as it seems. void could probably be ignored, but I don't really want to touch that in any way.

dordsor21: Mm yeah fair

dordsor21 (approved):

NotMyFault (approved):

PR #2936, opened 2024-09-30
fix: don't require WEOrigin / WEOffset in MCEdit schematics

PierreSchwang:
## Overview

```[tasklist]
### Submitter Checklist
```


SirYwell (approved):
I guess not catching the exception but rather using the `find` methods and checking everything for null won't make the code any cleaner, so that's fine for me

PierreSchwang:
> I guess not catching the exception but rather using the `find` methods and checking everything for null won't make the code any cleaner, so that's fine for me

Yup, using #findTag was my first idea - but I just did it like WorldEdit does it to not differ that much. That format is EOL either way

dordsor21 (approved):

PR #2945, opened 2024-10-12
Use default nbt when parsing blocks

### LINKED ISSUE #2944: `//set oak_sign` places invisible signs
> Edit: After creating the issue I figured out this **also applies for player heads**

### Server Implementation

Paper

### Server Version

1.21

### Describe the bug

Using `//set oak_sign` results in invisible sings being placed. However, `//set hand` whilst holding an oak sign does not result in this behaviour.

### To Reproduce

1. Select a region with `//pos1` and `//pos2`
2. Use `//set oak_sign`
3. Observe that the signs are invisible

### Expected behaviour

Signs are visible

### Screenshots / Videos

1) Using `//set oak_sign`

2) Using `//set hand` whilst holding an oak sign

### Error log (if applicable)

None

### Fawe Debugpaste

https://athion.net/ISPaster/paste/view/6758d277497f47b4ba4763c112768842

### Fawe Version

2.11.3-SNAPSHOT-934;b343be4

### Checklist

### Anything else?

It's also worth mentioning that after having placed signs via `//set hand`(i.e. visible signs), then replacing the region to air (`//set 0`) and then undoing the action (`//undo`), then the signs are invisible. No error in console

SirYwell:
## Overview

Fixes #2944

## Description

Not setting the block with its default NBT data results in the (vanilla) client not rendering the blocks properly. We should also probably pull upstream changes in that area at some point, but that didn't look as straightforward as this fix.

```[tasklist]
### Submitter Checklist
```

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/extension/factory/parser/DefaultBlockParser.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/extension/factory/parser/DefaultBlockParser.java
@@ -465,6 +465,7 @@ public class DefaultBlockParser extends InputParser<BaseBlock> {
                 if (state == null) {
                     throw new NoMatchException(Caption.of("fawe.error.invalid-block-type", TextComponent.of(input)));
                 }
+                nbt = state.getNbtData();
             }
             //FAWE end
 


PierreSchwang (approved):
Looks good I guess.
I think we should apply upstream changes rather sooner than later, so that NBT data can be specified as well - so that would probably address that as well, kind of (https://worldedit.enginehub.org/en/latest/usage/general/patterns/#single-block-pattern)

NotMyFault (approved):

PR #2962, opened 2024-10-27
Make property changes more robust

### LINKED ISSUE #2961: Replacing walls with fences keeping equal block states throws error
### Server Implementation

Paper

### Server Version

1.21

### Describe the bug

An IllegalArgumentException is thrown when replacing walls with fences keeping equal block states.

### To Reproduce

1. Place any type of wall
2. Attempt to replace the wall with any type of fence of equal block state with `^`

### Expected behaviour

Connected fences replace the walls with equal block states as they do in reverse when replacing fences with equal block state walls.

### Screenshots / Videos

_No response_

### Error log (if applicable)

https://paste.gg/p/anonymous/880e3ccb23ac4171a071549e7a1eedb6

### Fawe Debugpaste

https://athion.net/ISPaster/paste/view/598f6942863948ec897815d43c9b5880

### Fawe Version

2.12.1-SNAPSHOT-949;c252850

### Checklist

### Anything else?

The latest version was installed after reading the checklist points.

SirYwell:
## Overview

Fixes #2961

## Description

Different block types can share a property with the same name but with different types. In such case, we shouldn't try to apply the property with invalid values.

We could probably also remove the `IllegalArgumentException` and just return the current block state instead in `BlockState#with(PropertyKey,V)`, but that might hide other problems. Thoughts?

```[tasklist]
### Submitter Checklist
```

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/world/block/BlockType.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/world/block/BlockType.java
@@ -247,6 +247,21 @@ public class BlockType implements Keyed, Pattern {
         return this.settings.propertiesMapArr.length > ordinal && this.settings.propertiesMapArr[ordinal] != null;
     }
 
+    /**
+     * {@return whether this block type has a property with given key of the given type}
+     *
+     * @param key          the key identifying the property
+     * @param propertyType the expected type of the property
+     * @since TODO
+     */
+    public boolean hasPropertyOfType(PropertyKey key, Class<?> propertyType) {
+        int ordinal = key.getId();
+        Property<?> property;
+        return this.settings.propertiesMapArr.length > ordinal
+                && (property = this.settings.propertiesMapArr[ordinal]) != null
+                && property.getClass() == propertyType;
+    }
+
     public <V> Property<V> getProperty(PropertyKey key) {
         try {
             return (Property<V>) this.settings.propertiesMapArr[key.getId()];
--- a/worldedit-core/src/main/java/com/fastasyncworldedit/core/registry/state/PropertyKey.java
+++ b/worldedit-core/src/main/java/com/fastasyncworldedit/core/registry/state/PropertyKey.java
@@ -137,4 +137,9 @@ public class PropertyKey implements Comparable<PropertyKey> {
         return Integer.compare(this.id, o.id);
     }
 
+    @Override
+    public String toString() {
+        return "PropertyKey[" + getName() + "]";
+    }
+
 }
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/world/block/BlockState.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/world/block/BlockState.java
@@ -351,7 +351,7 @@ public class BlockState implements BlockStateHolder<BlockState>, Pattern {
         BlockState newState = this;
         for (Property<?> prop : ot.getProperties()) {
             PropertyKey key = prop.getKey();
-            if (blockType.hasProperty(key)) {
+            if (blockType.hasPropertyOfType(key, prop.getClass())) {
                 newState = newState.with(key, other.getState(key));
             }
         }


dordsor21:
We should keep the exception, else we're just silently failing. No failure implies it should have "worked"

worldedit-core/src/main/java/com/sk89q/worldedit/world/block/BlockType.java
```diff
@@ -247,6 +247,21 @@ public boolean hasProperty(PropertyKey key) {
...
+     * {@return whether this block type has a property with given key of the given type}
+     *
+     * @param key          the key identifying the property
+     * @param propertyType the expected type of the property
+     * @since TODO
+     */
+    public boolean hasPropertyOfType(PropertyKey key, Class<?> propertyType) {
+        int ordinal = key.getId();
+        Property<?> property;
+        return this.settings.propertiesMapArr.length > ordinal
```
dordsor21, line 260: Does this definitely make sense to have? If the key ID is dynamic to the original block state, that may not work, and if it's global that wouldn't work either?

SirYwell: Our `PropertyKey` and its ID are unique, so they are at the same position in each `propertiesMapArr`. But the actual property class stored there can have different types in different arrays.

dordsor21: Okay cool, so just a quick and dirty check so we don't need to check further

SirYwell:
> We should keep the exception, else we're just silently failing. No failure implies it should have "worked"

Probably worth to note that the default impl of `Property#getIndex(T)` doesn't fail when passing incompatible types, but overridden methods will throw a CCE. That's also why fence -> wall doesn't throw in the related issue, but wall -> fence does.

dordsor21:
> > We should keep the exception, else we're just silently failing. No failure implies it should have "worked"
> 
> Probably worth to note that the default impl of `Property#getIndex(T)` doesn't fail when passing incompatible types, but overridden methods will throw a CCE. That's also why fence -> wall doesn't throw in the related issue, but wall -> fence does.

I think I'm looking at the wrong getIndex method - it seems like it returns -1 on failure?

SirYwell:
> I think I'm looking at the wrong getIndex method - it seems like it returns -1 on failure? in all overrides?

E.g. https://github.com/IntellectualSites/FastAsyncWorldEdit/blob/8c0195970b2267f11095ffb54df6174356e0d08b/worldedit-core/src/main/java/com/sk89q/worldedit/registry/state/BooleanProperty.java#L39-L42 only accepts `Boolean` arguments. The compiler adds a synthetic bridge method `getIndex(Object)` that basically just delegates *but with a cast*. That's where we fail then.

dordsor21:
> E.g.
> 
> https://github.com/IntellectualSites/FastAsyncWorldEdit/blob/8c0195970b2267f11095ffb54df6174356e0d08b/worldedit-core/src/main/java/com/sk89q/worldedit/registry/state/BooleanProperty.java#L39-L42
> 
> only accepts `Boolean` arguments. The compiler adds a synthetic bridge method `getIndex(Object)` that basically just delegates _but with a cast_. That's where we fail then.

Ah yeah of course, missed that. I wonder if we should keep that though - having the wrong value of the right type return -1 is okay, whereas the wrong type altogether feels like a larger error (as would be the case for nearly/if not all other generic-type situations like this)

dordsor21 (approved):

dordsor21:
Can the check also be added to other places (e.g. https://github.com/IntellectualSites/FastAsyncWorldEdit/pull/2962/files#diff-cea7d1d5239fdee3588b7b5f77b2daab52c17ae9417fac4b9e2975e65b9ce0f4L321 )? Related to #2963 - what should we do in that case?

dordsor21:
So this isn't actually fully fixing the issue. It's possible to have properties of the same type, with the same length of values, and the same name (i.e. slab and chest `type`).

PierreSchwang (approved):
LGTM

PR #2970, opened 2024-10-30
API: New Clipboard helper methods

JackerAcid:
## Overview

Make Clipboards easier to use and less boilerplate.

## Description

Adds `getCenter()`, `getDiagonal()`, `getDiagonalSq()`, `getRadius()`, and `getRadiusSq()` methods to `com.sk89q.worldedit.extent.clipboard.Clipboard`.


dordsor21:
Why is this needed? Isn't this something that would be better suited for being in CunoidRegion anyway? The diagonal method is clearly for a very specific use case and does not adequately describe which diagonal it is for (there are three) (and I cannot think of any good way of implementing anything proper anyway)

PR #3022, opened 2024-12-08
Optimize layer iteration to prevent performance overhead.

NonSwag:
## Overview

Clamp the min and max y values to the min and max values of the chunk

## Description

Adjusted the layer iteration logic in the Region class to use the minimum and maximum Y-values from the chunk. This prevents unnecessary iterations when getMinimumY or getMaximumY returns extreme values, improving performance efficiency.

```[tasklist]
### Submitter Checklist
```

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/regions/Region.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/regions/Region.java
@@ -403,7 +403,7 @@ public interface Region extends Iterable<BlockVector3>, Cloneable, IBatchProcess
         if (tx >= min.x() && bx <= max.x() && tz >= min.z() && bz <= max.z()) {
             // contains some
             boolean processExtra = false;
-            for (int layer = getMinimumY() >> 4; layer <= getMaximumY() >> 4; layer++) {
+            for (int layer = Math.max(getMinimumY(), chunk.getMinY()) >> 4; layer <= Math.min(getMaximumY(), chunk.getMaxY()) >> 4; layer++) {
                 if (!set.hasSection(layer)) {
                     continue;
                 }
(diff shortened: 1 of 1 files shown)


NonSwag:
I was not sure whether it would be more appropriate to use `getMinSectionPosition` or `getMaxSectionPosition` over `getMinY` or `getMaxY`

worldedit-core/src/main/java/com/sk89q/worldedit/regions/Region.java
```diff
@@ -403,7 +403,9 @@ default IChunkSet processSet(IChunk chunk, IChunkGet get, IChunkSet set) {
         if (tx >= min.x() && bx <= max.x() && tz >= min.z() && bz <= max.z()) {
             // contains some
             boolean processExtra = false;
-            for (int layer = getMinimumY() >> 4; layer <= getMaximumY() >> 4; layer++) {
+            final int max1 = Math.max(getMinimumY(), chunk.getMinY());
+            final int min1 = Math.min(getMaximumY(), chunk.getMaxY());
+            for (int layer = max1 >> 4; layer <= min1 >> 4; layer++) {
```
PierreSchwang, line 408: I think, something like

```java
final int minLayer = Math.max(getMinimumY(), chunk.getMinY()) >> 4;
// ...
for (int layer = minLayer; //...
```
would be more readable. It seems like the variable names are mixed up either way

NonSwag: Gonna rename them later
Thanks for the input
Btw the names are not mixed up

PierreSchwang:
Would be great if the variables would either be called `minY` instead of `minLayer` or keep them `minLayer` but put the bit shift directly after the Math.min/max instead of inside the loop - just so that the naming fits

SirYwell (approved):

PierreSchwang (approved):

PR #3051, opened 2024-12-30
fix: use a holder to allow "upgrading" LocalBV3Set to a full fat one

dordsor21:
There was a/some issue(s) that were due to a local set being used assumptively. The only instances of the local sets being explicitly used should now only be when using chunk coords, or for the floating tree remover

### DIFF
--- a/worldedit-core/src/main/java/com/fastasyncworldedit/core/math/BlockVectorSet.java
+++ b/worldedit-core/src/main/java/com/fastasyncworldedit/core/math/BlockVectorSet.java
@@ -12,8 +12,10 @@ import java.util.AbstractCollection;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Iterator;
+import java.util.Map;
 import java.util.NoSuchElementException;
 import java.util.Objects;
+import java.util.stream.Collectors;
 
 /**
  * The BlockVectorSet is a memory optimized Set for storing {@link BlockVector3}'s.
@@ -26,6 +28,26 @@ import java.util.Objects;
 public class BlockVectorSet extends AbstractCollection<BlockVector3> implements BlockVector3Set {
 
     private final Long2ObjectLinkedOpenHashMap<LocalBlockVectorSet> localSets = new Long2ObjectLinkedOpenHashMap<>(4);
+
+    public BlockVectorSet() {
+    }
+
+    public BlockVectorSet(LocalBlockVectorSet localSet) {
+        if (localSet != null && localSet.isInitialised()) {
+            int offsetX = localSet.offsetX();
+            int offsetY = localSet.offsetY();
+            int offsetZ = localSet.offsetZ();
+            if ((offsetX & 2048) != 0 && (offsetY & 512) != 0 && (offsetZ & 2048) != 0) { // Can plug and play
+                localSets.put(MathMan.tripleWorldCoord(offsetX >> 11, offsetY >> 9, offsetZ >> 11), localSet);
+            } else {
+                localSet.forEach((x, y, z, index) -> this.add(x, y, z));
+            }
+        }
+    }
+
+    private BlockVectorSet(Map<Long, LocalBlockVectorSet> sets) {
+        localSets.putAll(sets);
+    }
 
     @Override
     public int size() {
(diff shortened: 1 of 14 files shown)


SirYwell:
Do we need to expose `BlockVector3SetHolder`? I think having a private/package-private class would be better, with static factory methods like `BlockVector3Set.expandable()` and `BlockVector3Set.fromLocal(LocalBlockVectorSet)` or something similar.

Also the `Cloneable` interface/the `clone()` method are kinda flawed, maybe it's better to just have our own `copy()` methods.

dordsor21:
> Do we need to expose `BlockVector3SetHolder`? I think having a private/package-private class would be better, with static factory methods like `BlockVector3Set.expandable()` and `BlockVector3Set.fromLocal(LocalBlockVectorSet)` or something similar.
> 
> Also the `Cloneable` interface/the `clone()` method are kinda flawed, maybe it's better to just have our own `copy()` methods.

Yeah I can probably move it into the local directly as there's no reason to have a holder for a full fat bv3set.

I also just copied the fact the local already had a clone method tbh, but yeah copy would also be consistent with everything else in FAWE

worldedit-core/src/main/java/com/fastasyncworldedit/core/math/LocalBlockVectorSet.java
```diff
@@ -58,6 +58,48 @@ public LocalBlockVectorSet(int x, int y, int z) {
...
+    }
+
+    /**
+     * New wrapped LocalBlockVectorSet with initial offset allowing "upgrading" to a {@link BlockVectorSet} as necessary
+     *
+     * @param x x offset
+     * @param z z offset
+     * @since TODO
+     */
+    public static BlockVector3Set wrapped(int x, int z) {
```
SirYwell, line 87: The offset isn't used in any way, what's the intention here?

worldedit-core/src/main/java/com/fastasyncworldedit/core/math/BlockVectorSet.java
```diff
@@ -257,10 +279,17 @@ public boolean removeAll(Collection<?> c) {
...
     public void clear() {
         localSets.clear();
     }
 
+    @Override
+    public BlockVectorSet copy() {
+        return new BlockVectorSet(localSets
+                .long2ObjectEntrySet()
+                .stream()
+                .collect(Collectors.toMap(Long2ObjectMap.Entry::getLongKey, e -> (LocalBlockVectorSet) e.getValue().clone())));
```
SirYwell, line 292: This should use the copy method I guess?

worldedit-core/src/main/java/com/fastasyncworldedit/core/util/collection/BlockVector3Set.java
```diff
@@ -7,7 +7,7 @@
 
 import java.util.Set;
 
-public interface BlockVector3Set extends Set<BlockVector3> {
+public interface BlockVector3Set extends Set<BlockVector3>, Cloneable {
```
SirYwell, line 10: `Cloneable` should be unneeded here now?

SirYwell (changes requested):

worldedit-core/src/main/java/com/fastasyncworldedit/core/math/LocalBlockVectorSet.java
dordsor21: Had meant to pass it onto the localbv3set

worldedit-core/src/main/java/com/fastasyncworldedit/core/math/BlockVectorSet.java
```diff
@@ -27,6 +29,26 @@ public class BlockVectorSet extends AbstractCollection<BlockVector3> implements
...
 
+    public BlockVectorSet() {
+    }
+
+    public BlockVectorSet(LocalBlockVectorSet localSet) {
+        if (localSet != null && localSet.isInitialised()) {
+            int offsetX = localSet.offsetX();
+            int offsetY = localSet.offsetY();
+            int offsetZ = localSet.offsetZ();
+            if ((offsetX & 2048) != 0 && (offsetY & 512) != 0 && (offsetZ & 2048) != 0) { // Can plug and play
```
SirYwell, line 40: I think I'm looking at this the third time and still questioning my sanity. But I assume this is supposed to check whether the localSet is already (2048, 512, 2048)-aligned? In that case, it would need to be `& 2047 == 0` etc. Otherwise I need an explanation because I can't wrap my head around it.

dordsor21: No you're right, I wasn't really thinking when I wrote this. I had meant to come back to look again but forgot and just committed

SirYwell (commented):
I also noticed `CachedMask` catches `UnsupportedOperationException`, should that be adjusted too? I guess those try-catch blocks can just be removed this way?

SirYwell (approved):
Looks good

NotMyFault (approved):

PR #3057, opened 2025-01-02
feat: Using Block Registry to Replace Iterating Over the Material Enum

Xiao-MoMi:
The Material in Bukkit is an enum, which is not very flexible. This PR replaces it with Minecraft's built-in block registry to reduce unnecessary checks. Additionally, it changes values() to getAllDefaultBlockStates() to improve readability.

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/world/registry/BlockRegistry.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/world/registry/BlockRegistry.java
@@ -94,7 +94,7 @@ public interface BlockRegistry {
     /**
      * Register all blocks.
      */
-    default Collection<String> values() {
+    default Collection<String> getAllDefaultBlockStates() {
         return Collections.emptyList();
     }
 
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/world/registry/BlockRegistry.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/world/registry/BlockRegistry.java
@@ -94,7 +94,7 @@ public interface BlockRegistry {
     /**
      * Register all blocks.
      */
-    default Collection<String> getAllDefaultBlockStates() {
+    default Collection<String> values() {
         return Collections.emptyList();
     }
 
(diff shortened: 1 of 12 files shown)


dordsor21:
I would point out that (iirc) Spigot (and paper) plan to replace enums for materials with a registry of similar sorts (this has already been done for biomes).

SirYwell:
I assume this is supposed to fix #3060? In that case, I'd be fine with the change, at least until spigot marks it BlockType class/Block registry as non-internal.

dordsor21:
> I assume this is supposed to fix #3060? In that case, I'd be fine with the change, at least until spigot marks it BlockType class/Block registry as non-internal.

I'm not sure, I feel we'd have more reports if there was a consistent issue? It should only be an issue if Spigot/Paper have stopped maintaining the Material enum for some reason...

Regardless, the API-breaking change needs to go

worldedit-core/src/main/java/com/sk89q/worldedit/world/registry/BlockRegistry.java
```diff
@@ -94,7 +94,7 @@ default BlockMaterial getMaterial(BlockState state) {
     /**
      * Register all blocks.
      */
-    default Collection<String> values() {
```
dordsor21, line 97: This breaks API compatibility

dordsor21 (changes requested):

worldedit-core/src/main/java/com/sk89q/worldedit/world/registry/BlockRegistry.java
Xiao-MoMi: Sorry for not noticing this, I rolled back the method name and changed the comment in latest changes

dordsor21:
I'd like to test with P2 first as I'm not fully convinced this will allow data to be available at the point it's needed there

dordsor21 (approved):
Took a second to give it a test, but looks to be working. lgtm

SirYwell (approved):

PR #3059, opened 2025-01-04
fix: don't lock for chunk packets/section write on main thread

dordsor21:


worldedit-bukkit/src/main/java/com/fastasyncworldedit/bukkit/adapter/NMSAdapter.java
```diff
@@ -192,6 +199,9 @@ protected static void beginChunkPacketSend(String worldName, IntPair pair, Stamp
      * @since 2.12.0
      */
     protected static void endChunkPacketSend(String worldName, IntPair pair, StampLockHolder lockHolder) {
+        if (Fawe.isMainThread()) {
```
SirYwell, line 202: In the current code, this always happens on the main thread or am I missing something?

worldedit-bukkit/src/main/java/com/fastasyncworldedit/bukkit/adapter/NMSAdapter.java
```diff
@@ -171,6 +175,9 @@ protected static <LevelChunkSection> boolean setSectionAtomic(
      * @since 2.12.0
      */
     protected static void beginChunkPacketSend(String worldName, IntPair pair, StampLockHolder stampedLock) {
+        if (Fawe.isMainThread()) {
```
SirYwell, line 178: Does that mean we're not sending any packets when on main thread? The callers run
```java
if (lockHolder.chunkLock == null)
    return;
}
```
directly afterwards.

SirYwell (changes requested):

worldedit-bukkit/src/main/java/com/fastasyncworldedit/bukkit/adapter/NMSAdapter.java
dordsor21: I might have been overly eager for the chunk packet methods

SirYwell (approved):

NotMyFault (approved):

PR #3066, opened 2025-01-07
Check for last chunk before acquiring lock

SirYwell:
## Description

Acquiring/releasing the lock in `getOrCreateChunk` takes a fair amount of time, even with no contention involved. By simplifying from two volatile variables to one, we can already have a fast path before the lock, This cuts the time spent in the method in half in my measurements.

One downside is the scenario where many calls  result in `NullChunk.getInstance()` being returned, as we can't check for the coordinates there. If that *is* a problem, it might make sense to just create multiple NullChunk objects with their actual coordinates.

```[tasklist]
### Submitter Checklist
```


dordsor21 (approved):
I don't think there's any reason for NullChunk to be an issue

NotMyFault (approved):

PR #3068, opened 2025-01-08
Only compute solid blocks once

SirYwell:
## Description

`SolidBlockMask` previously went through all block states on construction to find all movement-blocking block states. While that doesn't take much time, it adds up as the mask is used in other places, e.g. on every click with a brush. Precalculating the values only once additionally allows us to use the information in other places where we previously called into NMS code through multiple layers of abstraction to check whether a block blocks movement.
For example operations using an angle mask can benefit from that.

```[tasklist]
### Submitter Checklist
```

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/function/mask/SolidBlockMask.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/function/mask/SolidBlockMask.java
@@ -20,23 +20,38 @@
 package com.sk89q.worldedit.function.mask;
 
 import com.sk89q.worldedit.extent.Extent;
+import com.sk89q.worldedit.math.BlockVector3;
+import com.sk89q.worldedit.world.block.BlockTypesCache;
 
-import javax.annotation.Nullable;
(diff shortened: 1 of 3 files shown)


PierreSchwang (approved):
untested, but looks good

worldedit-core/src/main/java/com/sk89q/worldedit/function/mask/SolidBlockMask.java
```diff
@@ -20,23 +20,48 @@
...
+    public boolean test(final Extent extent, final BlockVector3 position) {
+        final int ordinal = position.getOrdinal(extent);
+        return SOLID[ordinal];
+    }
+
     @Override
-    public Mask2D toMask2D() {
-        return null;
+    public boolean test(final BlockVector3 vector) {
+        return test(getExtent(), vector);
```
dordsor21, line 53: `vector.test(getExtent())`?

NotMyFault (approved):

dordsor21 (approved):

PR #3071, opened 2025-01-09
Return air instead of __reserved__ when reading non-cuboid clipboard

### LINKED ISSUE #3065: The construction diagram highlighted using //sel cyl is inserted without air blocks
### Server Implementation

Paper

### Server Version

1.20.4

### Describe the bug

Hello everyone, I have clearly shown the problem in the video. I selected the building and copied it using //sel cyl, using the usual //paste - everything is inserted as it should be, but when creating schematics, for some reason the building is inserted without air blocks inside. How can I fix this?

Watch the video at the link

### To Reproduce

1. Select //sel cyl
2. Select all buildings
3. //copy
4. Try //paste
Further
5. //schem save name
6. //schem load name
7. //paste
See the difference

### Expected behaviour

I need the schematics to be inserted with air blocks (just like the construction is inserted using //copy and //paste)

### Screenshots / Videos

Video - https://drive.google.com/file/d/1ySi5vTUnZ-pc6UGN05lDP7HmIu7GKKCR/view?usp=sharing

### Error log (if applicable)

_No response_

### Fawe Debugpaste

https://athion.net/ISPaster/paste/view/21e6d34dc55c4696942ddca40c1e760f

### Fawe Version

FastAsyncWorldEdit version 2.12.4-SNAPSHOT-1012;84e7219

### Checklist

### Anything else?

_No response_

SirYwell:
## Overview

Fixes #3065

## Description

Returning the `__reserved__` block state results in clipboards containing that block state, means they can't be loaded by other tools correctly and FAWE won't place air correctly.

```[tasklist]
### Submitter Checklist
```


dordsor21:
Can we potentially add a flag somewhere for allowing the reserved? If we set a clipboard within FAWE it might be wanted not to set outside the original copy region, but still set air within. Presumably this is also an issue for the other FAWE implementations as well?

SirYwell:
> Can we potentially add a flag somewhere for allowing the reserved? If we set a clipboard within FAWE it might be wanted not to set outside the original copy region, but still set air within.

This just mirrors the behavior of WE. Manually checking if the point is in the region when needed sounds more appropriate.

> Presumably this is also an issue for the other FAWE implementations as well?

It seems like `WorldCopyClipboard` will just return the actual blocks outside of its region, so that's wrong in a different way. The memory clipboard seems to work correctly from a quick test, not sure why though.

dordsor21:
> > Can we potentially add a flag somewhere for allowing the reserved? If we set a clipboard within FAWE it might be wanted not to set outside the original copy region, but still set air within.
> 
> This just mirrors the behavior of WE. Manually checking if the point is in the region when needed sounds more appropriate.

We won't necessarily have the selection when pasting - it could have changed between the copy and paste, or have been made before a restart

> > Presumably this is also an issue for the other FAWE implementations as well?
> 
> It seems like `WorldCopyClipboard` will just return the actual blocks outside of its region, so that's wrong in a different way. The memory clipboard seems to work correctly from a quick test, not sure why though.

memory optimized doesn't "store" air so I believe would assume reserved is just normal air. CPU optimized is potentially problematic as it'd read a char of 0 from the array

SirYwell:
> We won't necessarily have the selection when pasting - it could have changed between the copy and paste, or have been made before a restart

Hm, that's an interesting scenario. I guess `SimpleRegion` always returning a `CuboidRegion` is wrong then in first place, and we need to somehow improve on that?

dordsor21:
> Hm, that's an interesting scenario. I guess `SimpleRegion` always returning a `CuboidRegion` is wrong then in first place, and we need to somehow improve on that?

Yeah potentially some more thought to that. I do think it's a feature to allow "non-rectangular" clipboards to have air without overwriting a selection

PR #3072, opened 2025-01-10
Optimize AngleMask

SirYwell:
## Description

`AngleMask` caused unnecessarily slow block lookups due to its access patterns. That can be improved. I'll add some inline comments about specific changes.

```[tasklist]
### Submitter Checklist
```

### DIFF
--- a/worldedit-core/src/main/java/com/fastasyncworldedit/core/function/mask/AngleMask.java
+++ b/worldedit-core/src/main/java/com/fastasyncworldedit/core/function/mask/AngleMask.java
@@ -133,10 +133,10 @@ public class AngleMask extends AbstractExtentMask implements ResettableMask {
         int y = base.y();
         // we expect the data for blocks above and below to be loaded already
         // in which case caching/reading from cache has more overhead
-        if (y != maxY && !fastMask.test(base.getStateRelativeY(extent, 1))) {
+        if (y != maxY && !SolidBlockMask.isSolid(base.getStateRelativeY(extent, 1))) {
             return true;
         }
-        if (y != minY && !fastMask.test(base.getStateRelativeY(extent, -1))) {
+        if (y != minY && !SolidBlockMask.isSolid(base.getStateRelativeY(extent, -1))) {
             return true;
         }
         // other positions might be in different chunks, go a slower, cached path there
--- a/worldedit-core/src/main/java/com/fastasyncworldedit/core/function/mask/AngleMask.java
+++ b/worldedit-core/src/main/java/com/fastasyncworldedit/core/function/mask/AngleMask.java
@@ -14,6 +14,7 @@ public class AngleMask extends AbstractExtentMask implements ResettableMask {
     protected static double ADJACENT_MOD = 0.5;
     protected static double DIAGONAL_MOD = 1 / Math.sqrt(8);
 
+    private final SolidBlockMask fastMask;
     protected final CachedMask mask;
     protected final double max;
     protected final double min;
@@ -32,7 +33,8 @@ public class AngleMask extends AbstractExtentMask implements ResettableMask {
 
     public AngleMask(Extent extent, double min, double max, boolean overlay, int distance) {
         super(extent);
-        this.mask = new CachedMask(new SolidBlockMask(extent));
+        this.fastMask = new SolidBlockMask(extent);
+        this.mask = new CachedMask(this.fastMask);
         this.min = min;
         this.max = max;
         this.checkFirst = max >= (Math.tan(90 * (Math.PI / 180)));
@@ -126,10 +128,21 @@ public class AngleMask extends AbstractExtentMask implements ResettableMask {
         }
     }
 
-    private boolean adjacentAir(Extent extent, MutableBlockVector3 mutable) {
-        int x = mutable.x();
-        int y = mutable.y();
-        int z = mutable.z();
+    // for optimal performance, this method should be called with base being a CharFilterBlock
+    private boolean adjacentAir(Extent extent, BlockVector3 base) {
+        int y = base.y();
+        // we expect the data for blocks above and below to be loaded already
+        // in which case caching/reading from cache has more overhead
+        if (y != maxY && !fastMask.test(base.getStateRelativeY(extent, 1))) {
+            return true;
+        }
+        if (y != minY && !fastMask.test(base.getStateRelativeY(extent, -1))) {
+            return true;
+        }
+        // other positions might be in different chunks, go a slower, cached path there
+        MutableBlockVector3 mutable = new MutableBlockVector3();
+        int x = base.x();
+        int z = base.z();
         if (!mask.test(extent, mutable.setComponents(x + 1, y, z))) {
             return true;
         }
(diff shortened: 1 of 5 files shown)


worldedit-core/src/main/java/com/fastasyncworldedit/core/queue/implementation/blocks/CharBlocks.java
```diff
@@ -20,6 +20,7 @@ public char[] get(CharBlocks blocks, int layer) {
             char[] arr = blocks.blocks[layer];
             if (arr == null) {
                 // Chunk probably trimmed mid-operations, but do nothing about it to avoid other issues
+                blocks.sections[layer] = EMPTY;
```
SirYwell, line 23: I got a `StackOverflowError` without this change due to the section being full but the array being null

worldedit-core/src/main/java/com/fastasyncworldedit/core/function/mask/AngleMask.java
```diff
@@ -14,6 +14,7 @@ public class AngleMask extends AbstractExtentMask implements ResettableMask {
     protected static double ADJACENT_MOD = 0.5;
     protected static double DIAGONAL_MOD = 1 / Math.sqrt(8);
 
+    private final SolidBlockMask fastMask;
```
SirYwell, line 17: `CachedMask` is somewhat expensive, I'm not exactly sure why but I assume due to the need of exception handling. Skipping it when we pass a CharFilterBlock as position seems to be always faster.

dordsor21: Which exception handling exactly? It shouldn't be slow unless we're filling the stacktrace

SirYwell: Just the presence of the exception region making optimizations harder. But that's just a wild guess, the profiler shows that 2/3 of the time in `CachedMask#test(Extent, BV3)` is actually spent in the method itself rather than in methods it is calling. But that might also be misattribution due to lacking debug info or something like that.

Also something I forgot to mention: The `CachedMask` is only really useful if `overlay` is true, otherwise positions won't be checked multiple times normally.

SirYwell: > I don't think it should have any effect on performance though, unless you're testing the mask on a large range where the exception would actually be thrown? Afaik the try-catch basically means nothing until the exception walks back up the stack

I'm not sure what's going on exactly there, but I didn't want to do a deep-dive before merging your changes. I'll take a closer look once both changes are merged, that should give a clearer image.

worldedit-core/src/main/java/com/fastasyncworldedit/core/queue/implementation/blocks/CharBlocks.java
dordsor21: It's probably best to use the getSkipFull method - we probably don't want to touch the sections array here? I make the change in #2912 as it is anyway

SirYwell: Remioved the change

worldedit-core/src/main/java/com/fastasyncworldedit/core/function/mask/AngleMask.java
```diff
@@ -126,38 +128,42 @@ protected boolean testSlope(Extent extent, int x, int y, int z) {
...
+        // in which case caching/reading from cache has more overhead
+        if (y != maxY && !fastMask.test(base.getStateRelativeY(extent, 1))) {
             return true;
         }
-        if (!mask.test(extent, mutable.setComponents(x - 1, y, z))) {
+        if (y != minY && !fastMask.test(base.getStateRelativeY(extent, -1))) {
             return true;
         }
-        if (!mask.test(extent, mutable.setComponents(x, y, z + 1))) {
+        // other positions might be in different chunks, go a slower, cached path there
```
dordsor21, line 142: Potentially worth coming back to this after the merge of this and #3051 - if the cache performance is still poor then probably worth testing if x/z are actually on the edges of the chunk. Maybe this could be implemented into the cache mask constructor - we can potentially look at ways of compressing the data further.

worldedit-core/src/main/java/com/fastasyncworldedit/core/function/mask/AngleMask.java
```diff
@@ -126,38 +128,42 @@ protected boolean testSlope(Extent extent, int x, int y, int z) {
...
         }
-        return y != minY && !mask.test(extent, mutable.setComponents(x, y - 1, z));
+        return !mask.test(extent, mutable.setComponents(x, y, z - 1));
     }
 
     @Override
     public boolean test(BlockVector3 vector) {
 
-        if (!mask.test(vector)) {
+        if (!fastMask.test(vector)) {
```
dordsor21, line 161: I think we should cache this at least if on the edge of the chunk, as there's little point using the cache mask on x/z +/- 1 if there's unlikely to be a value already in the cache. We know we're going to be testing every block here

SirYwell: Not sure if it's worth the complexity to have different code paths for those cases. I tried again to revert this change to use the cached mask and it makes the whole mask slower even in the overlay scenario.

I also replaced the usages of the `CachedMask` in `adjacentAir`, bringing time spent in `adjacentAir` down from ~25s (when using the cached mask everywhere) to ~10s in my test... so I guess we really need to look into CachedMask after your PR.

dordsor21: Okay that's definitely concerning then. I suppose the caching makes it slower if the chunks are already available, and we will already have those GET chunks loaded in FAWE

SirYwell: Yes that code path doesn't seem to run into chunk loading, so any access is more or less direct (going through STQE)

SirYwell:
After #3051 got merged, it still seems to be a performance win avoiding the CachedMask usage completely: 
comparison (*old* is the current state of this PR, *new* is replacing the remaining usages of `mask` with `fastMask`.

dordsor21:
I suppose it's probably worth removing the CachedMask in this case entirely?

SirYwell:
I think so, yeah. Maybe it also comes down to how chunks are spread across threads, that might make caching values of surrounding chunks less useful too.

The CachedMask field is protected, do you think it's still fine to remove it?

dordsor21:
> The CachedMask field is protected, do you think it's still fine to remove it?

Yes, we've no contract to use a CachedMask here. I assume there are child classes that use the "mask" that don't necessarily know it's Cached?

SirYwell:
> Yes, we've no contract to use a CachedMask here. I assume there are child classes that use the "mask" that don't necessarily know it's Cached?

The field is/was a CachedMask, but no subclass in FAWE was using the field in any way.

dordsor21 (approved):

NotMyFault (approved):

PR #3082, opened 2025-01-19
Improve BiomeMask performance

SirYwell:
## Description

BiomeMask and biome access generally are rather slow, and a lot of the slowness is caused by inefficient translation and slow contains checks. The adaption from NMS biomes to WE BiomeType can be simplified as we already set the legacy id (= NMS-internal id). The HashSet can be replaced by a boolean array, similar to how BlockMask works.

```[tasklist]
### Submitter Checklist
```

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/world/biome/BiomeTypes.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/world/biome/BiomeTypes.java
@@ -20,7 +20,10 @@
 package com.sk89q.worldedit.world.biome;
 
 import javax.annotation.Nullable;
+import java.util.Arrays;
 import java.util.Collection;
+import java.util.Comparator;
+import java.util.List;
 
 /**
  * Stores a list of common {@link BiomeType BiomeTypes}.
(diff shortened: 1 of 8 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/world/biome/BiomeTypes.java
```diff
@@ -297,10 +300,18 @@ public static BiomeType register(final BiomeType biome) {
...
+
+            private static List<BiomeType> create() {
+                BiomeType[] array = values().toArray(new BiomeType[0]);
+                Arrays.sort(array, Comparator.comparing(BiomeType::getLegacyId));
+                return List.of(array);
             }
+
+        }
+        if (legacyId >= 0 && legacyId < BiomeTypeIdCache.byId.size()) {
+            return BiomeTypeIdCache.byId.get(legacyId);
```
PierreSchwang, line 314: Any specific reason on using a List instead of an array?

SirYwell: Not really, the list is immutable and its backing array is marked stable which might help performance in some cases, but it likely doesn't matter.

dordsor21: Yeah I think any performance drops from using a list should be compiled away given it's a static final list, so it should just act as an array in runtime

dordsor21 (approved):

PierreSchwang (approved):

PR #3092, opened 2025-01-26
fix: don't attempt to write structure void blocks in minecraft structure

### LINKED ISSUE #3090: Structure Voids are replaced with a placeholder (in structure schem)
### Server Implementation

Paper

### Server Version

1.21.3

### Describe the bug

When saving my clipboard as an nbt structure, FAWE seems to replace structure void blocks with a random block relating to the structure, but only if the structure blocks cover an entire layer's air blocks.

### To Reproduce

1. Build a small hut / structure with open interior
2. Select the hut and fill all air blocks with structure void
3. Remove all structure void that remain inside the structure
4. Copy the selection
5. Save the selection as an nbt structure schematic
6. Load the schematic
7. Paste

### Expected behaviour

The expected behavior would be for the structure void to be pasted with OR removed (based on how Minecraft's jigsaw structure saving works - I'm not 100% aware of the block placement result, but I am aware that it should either place structure voids, or nothing - definitely not a random block)

### Screenshots / Videos

Example Build
Image

Structure NBT save (after replacing blocks at the door holes and below with structure void)
Image

Loading, and pasting the saved structure schematic.

Image

### Error log (if applicable)

_No response_

### Fawe Debugpaste

https://athion.net/ISPaster/paste/view/aefc6b24ad0b49a3b3767963f8e635cc

### Fawe Version

2.12.4-SNAPSHOT-1025;63f7871

### Checklist

### Anything else?

Thanks for taking a look! I hope this gets fixed in a somewhat timely manner if this is indeed a bug and not a misuse on my end.

PierreSchwang:
## Overview
Fixes #3090

```[tasklist]
### Submitter Checklist
```

### DIFF
--- a/worldedit-core/src/main/java/com/fastasyncworldedit/core/extent/clipboard/io/schematic/MinecraftStructure.java
+++ b/worldedit-core/src/main/java/com/fastasyncworldedit/core/extent/clipboard/io/schematic/MinecraftStructure.java
@@ -225,7 +225,10 @@ public class MinecraftStructure implements ClipboardReader, ClipboardWriter {
         // Blocks
         LinListTag.Builder<@org.jetbrains.annotations.NotNull LinCompoundTag> blocks = LinListTag.builder(LinTagType.compoundTag());
         for (final BlockVector3 pos : clipboard) {
-            final BlockState block = clipboard.getBlock(pos);
+            final BaseBlock block = clipboard.getFullBlock(pos);
+            if (block.getBlockType() == BlockTypes.STRUCTURE_VOID) {
+                continue;
+            }
             LinCompoundTag.Builder entry = LinCompoundTag.builder()
                     .putInt("state", ordinals.get(block.getOrdinalChar()))
                     .put("pos", LinListTag.of(LinTagType.intTag(), List.of(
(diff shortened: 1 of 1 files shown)


IronApollo (approved):

worldedit-core/src/main/java/com/fastasyncworldedit/core/extent/clipboard/io/schematic/MinecraftStructure.java
```diff
@@ -225,20 +225,20 @@ public void write(Clipboard clipboard) throws IOException {
         // Blocks
         LinListTag.Builder<@org.jetbrains.annotations.NotNull LinCompoundTag> blocks = LinListTag.builder(LinTagType.compoundTag());
```
dordsor21, line 226: Change this top javax while we're here I guess?

PierreSchwang: That would be great, but i hate that 
image
image

guess i could suppress that simply

RedstoneFuture:
By default, _Structure Void_ blocks should be pasted because they are simply part of the structure / schematic, I think.

In vanilla, `structure_void` is like the `//paste` command flag “-a” from FAWE, but “-a” affects `air` and `structure_void` would remain. So the vanilla behavior is exactly the opposite of the FAWE behavior when using "-a".

Maybe a paste command flag for FAWE would be an idea, which would paste AIR like in Vanilla and leave out `structure_void` blocks. Anything else would perhaps be unusual.

Leomelonseeds:
Just ran into this issue saving Minecraft structure schematics for my plugin, would appreciate this getting merged!

PierreSchwang:
Should be fine now ig.
Saving a structure with structure void blocks doesn't save the structure void blocks in the palette or block compound (they are ignored). This seems to be consistent with the actual Structure File format, as I haven't found any builtin structure file which actually contains structure void blocks. I would guess they are simply ignored as a whole, which fits structure void blocks.

When loading Structure File clipboards / schematics the resulting clipboard is filled with structure void blocks, so that every clipboard position actually has a block. Known blocks are overriding the structure void blocks.

Pasting without any arguments pastes the clipboard including the structure void blocks (suitable for knowing where they are of if the structure needs to be changed). Passing the -v switch (e.g. `//paste -v`) doesn't paste structure void blocks (inverse block type source mask) and keeps the target block unmodified (the one which would've been overridden with the structure void, kinda like `-a`). That should follow the spec as described in https://minecraft.wiki/w/Structure_Void

PR #3096, opened 2025-01-31
Avoid ThreadLocal lookup in ParallelQueueExtent

SirYwell:
## Description

`ThreadLocal` has some overhead that can be avoided as we're controlling our thread pools ourselves. By just having specific thread types with direct access, we can improve performance of commands that go through the `ParallelQueueExtent#getExtent()` code a lot (i.e. `//copy`). As most time is spent loading chunks there, it's somewhat hard to measure, but the time spent in the `getExtent()` method is reduced by ~90%.

```[tasklist]
### Submitter Checklist
```


dordsor21 (approved):

dordsor21:
This feels surprising; I had always understood that threadlocals were efficient and performant

NotMyFault (approved):

PR #3102, opened 2025-02-07
Locality-aware chunk distribution

SirYwell:
## Description

`ParallelQueueExtent` deals with distributing work onto multiple threads to speed up edits. In its current implementation, this is achieved by just using the `Region#getChunks().iterator()` as a queue and let all threads pull the next chunk to work on from there. This works well typically, but it has one downside: Different threads work on neighboring chunks. There are multiple reasons this isn't optimal:

- Filters accessing data of surrounding chunks will cause more cache misses in their `STQE`. As currently this also means that one `STQE` will trim the chunk cached/processed by another `STQE`, we do a lot of extra work in such cases.
- Due to how Minecraft generates chunks, processing neighboring chunks on different threads can be slower if both threads wait for the same resources.

By distributing work more location-aware, these downsides can be prevented, at the cost of a somewhat more complicated work-splitting logic. This means threads will process more tasks, but those tasks are smaller and faster to process. As an additional benefit, this allows better work distribution if multiple edits are done concurrently, as the threads don't have to finish one edit before they are free again for the next edit. Consequently, thread-local data needs to be cleared and re-set more often.

Exception handling is a bit of a question mark with this approach, I tried to get it near to the original handling, but I'm not sure if it makes sense the way it is right now.

```[tasklist]
### Submitter Checklist
```

### DIFF
--- a/worldedit-core/src/main/java/com/fastasyncworldedit/core/queue/implementation/ApplyTask.java
+++ b/worldedit-core/src/main/java/com/fastasyncworldedit/core/queue/implementation/ApplyTask.java
@@ -8,6 +8,9 @@ import com.fastasyncworldedit.core.queue.Filter;
 import com.sk89q.worldedit.internal.util.LogManagerCompat;
 import com.sk89q.worldedit.math.BlockVector3;
 import com.sk89q.worldedit.regions.Region;
+import jdk.jfr.Category;
+import jdk.jfr.Event;
+import jdk.jfr.Name;
 import org.apache.logging.log4j.Logger;
 
 import java.util.Collection;
@@ -21,7 +24,7 @@ class ApplyTask<F extends Filter> extends RecursiveAction implements Runnable {
     private static final Logger LOGGER = LogManagerCompat.getLogger();
 
     private static final int INITIAL_REGION_SHIFT = 5;
-    private static final int SHIFT_REDUCTION = 2;
+    private static final int SHIFT_REDUCTION = 1;
 
     private final CommonState<F> commonState;
     private final ApplyTask<F> before;
@@ -116,7 +119,7 @@ class ApplyTask<F extends Filter> extends RecursiveAction implements Runnable {
             // This task covers multiple regions. Create one subtask per region
             for (int regionX = minRegionX; regionX <= maxRegionX; regionX++) {
                 for (int regionZ = minRegionZ; regionZ <= maxRegionZ; regionZ++) {
-                    if (ForkJoinTask.getSurplusQueuedTaskCount() > Settings.settings().QUEUE.PARALLEL_THREADS) {
+                    if (shouldProcessDirectly()) {
                         // assume we should do a bigger batch of work here - the other threads are busy for a while
                         processRegion(regionX, regionZ, this.shift);
                         continue;
@@ -155,6 +158,10 @@ class ApplyTask<F extends Filter> extends RecursiveAction implements Runnable {
         if (this.shift == INITIAL_REGION_SHIFT) {
             onCompletion();
         }
+    }
+
+    private boolean shouldProcessDirectly() {
+        return ForkJoinTask.getSurplusQueuedTaskCount() > Math.max(3, 1 << this.shift);
     }
 
     private void processRegion(int regionX, int regionZ, int shift) {
--- a/worldedit-core/src/main/java/com/fastasyncworldedit/core/queue/implementation/ApplyTask.java
+++ b/worldedit-core/src/main/java/com/fastasyncworldedit/core/queue/implementation/ApplyTask.java
@@ -1,16 +1,12 @@
 package com.fastasyncworldedit.core.queue.implementation;
 
 import com.fastasyncworldedit.core.Fawe;
-import com.fastasyncworldedit.core.configuration.Settings;
 import com.fastasyncworldedit.core.extent.filter.block.ChunkFilterBlock;
 import com.fastasyncworldedit.core.internal.exception.FaweException;
 import com.fastasyncworldedit.core.queue.Filter;
 import com.sk89q.worldedit.internal.util.LogManagerCompat;
 import com.sk89q.worldedit.math.BlockVector3;
 import com.sk89q.worldedit.regions.Region;
-import jdk.jfr.Category;
-import jdk.jfr.Event;
-import jdk.jfr.Name;
 import org.apache.logging.log4j.Logger;
 
 import java.util.Collection;
--- a/worldedit-core/src/main/java/com/fastasyncworldedit/core/queue/implementation/ApplyTask.java
+++ b/worldedit-core/src/main/java/com/fastasyncworldedit/core/queue/implementation/ApplyTask.java
@@ -99,8 +99,8 @@ class ApplyTask<F extends Filter> extends RecursiveAction implements Runnable {
 
     @Override
     protected void compute() {
-        ApplyTask<F> subtask = null;
         if (this.minChunkX != this.maxChunkX || this.minChunkZ != this.maxChunkZ) {
+            ApplyTask<F> subtask = null;
             int minRegionX = this.minChunkX >> this.shift;
             int minRegionZ = this.minChunkZ >> this.shift;
             int maxRegionX = this.maxChunkX >> this.shift;
@@ -131,18 +131,18 @@ class ApplyTask<F extends Filter> extends RecursiveAction implements Runnable {
                     subtask.fork();
                 }
             }
+            // try processing tasks in reverse order if not processed already, otherwise "wait" for completion
+            while (subtask != null) {
+                if (subtask.tryUnfork()) {
+                    subtask.invoke();
+                } else {
+                    subtask.quietlyJoin();
+                }
+                subtask = subtask.before;
+            }
         } else {
             // we reached a task for a single chunk, let's process it
             processChunk(this.minChunkX, this.minChunkZ);
-        }
-        // try processing tasks in reverse order if not processed already, otherwise "wait" for completion
-        while (subtask != null) {
-            if (subtask.tryUnfork()) {
-                subtask.invoke();
-            } else {
-                subtask.quietlyJoin();
-            }
-            subtask = subtask.before;
         }
         if (this.shift == INITIAL_REGION_SHIFT) {
             onCompletion();
--- a/worldedit-core/src/main/java/com/fastasyncworldedit/core/queue/implementation/ApplyTask.java
+++ b/worldedit-core/src/main/java/com/fastasyncworldedit/core/queue/implementation/ApplyTask.java
@@ -23,6 +23,8 @@ class ApplyTask<F extends Filter> extends RecursiveAction implements Runnable {
     private final int minChunkZ;
     private final int maxChunkX;
     private final int maxChunkZ;
+    // Note: shift == INITIAL_REGION_SHIFT means we are in the root node.
+    // compute() relies on that when triggering postProcess
     private final int shift;
 
     @Override
@@ -112,7 +114,7 @@ class ApplyTask<F extends Filter> extends RecursiveAction implements Runnable {
                         continue;
                     }
                     if (this.shift == 0 && !this.commonState.region.containsChunk(regionX, regionZ)) {
-                        // if shift == 0, region ccords are chunk coords
+                        // if shift == 0, region coords are chunk coords
                         continue; // chunks not intersecting with the region don't need a task
                     }
 
@@ -136,22 +138,19 @@ class ApplyTask<F extends Filter> extends RecursiveAction implements Runnable {
         // try processing tasks in reverse order if not processed already, otherwise "wait" for completion
         while (subtask != null) {
             if (subtask.tryUnfork()) {
-                subtask.compute();
+                subtask.invoke();
             } else {
                 subtask.quietlyJoin();
             }
             subtask = subtask.before;
         }
+        if (this.shift == INITIAL_REGION_SHIFT) {
+            onCompletion();
+        }
     }
 
     private void processRegion(int regionX, int regionZ, int shift) {
-        final ThreadState<F> state = this.commonState.stateCache.computeIfAbsent(
-                Thread.currentThread(),
-                __ -> new ThreadState<>(
-                        (SingleThreadQueueExtent) this.commonState.parallelQueueExtent.getNewQueue(),
-                        (F) this.commonState.originalFilter.fork()
-                )
-        );
+        final ThreadState<F> state = getState();
         this.commonState.parallelQueueExtent.enter(state.queue);
         try {
             for (int chunkX = regionX << shift; chunkX <= ((regionX  + 1) << shift) - 1; chunkX++) {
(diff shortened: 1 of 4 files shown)


worldedit-core/src/main/java/com/fastasyncworldedit/core/queue/implementation/ApplyTask.java
```diff
@@ -0,0 +1,241 @@
...
+        for (ForkJoinTask<?> task : postProcess()) {
+            if (task.tryUnfork()) {
+                task.invoke();
+            } else {
+                task.join();
+            }
+        }
+    }
+
+    private ForkJoinTask<?>[] postProcess() {
```
dordsor21, line 230: Is postProcess necessarily the best name here? Given the use of postprocess already to define actions per-chunk. Perhaps just flushQueues is fine as we're not doing any processing (postly or otherwise) here

worldedit-core/src/main/java/com/fastasyncworldedit/core/queue/implementation/ApplyTask.java
```diff
@@ -0,0 +1,241 @@
...
+        }
+    }
+
+    private void processRegion(int regionX, int regionZ, int shift) {
+        final ThreadState<F> state = getState();
+        this.commonState.parallelQueueExtent.enter(state.queue);
+        try {
+            for (int chunkX = regionX << shift; chunkX <= ((regionX  + 1) << shift) - 1; chunkX++) {
+                for (int chunkZ = regionZ << shift; chunkZ <= ((regionZ  + 1) << shift) - 1; chunkZ++) {
+                    if (!this.commonState.region.containsChunk(chunkX, chunkZ)) {
```
dordsor21, line 166: Is it worth adding a micro-optimisation here to bound the initial and final chunkX|Z to a precomputed min/max? I suppose given it's not in the edit itself isn't necessarily an issue, but this could be a relatively complex operation depending on the region implementation

worldedit-core/src/main/java/com/fastasyncworldedit/core/queue/implementation/ApplyTask.java
```diff
@@ -0,0 +1,241 @@
...
+                    }
+
+                    // creating more tasks will likely help parallelism as other threads aren't *that* busy
+                    subtask = new ApplyTask<>(
+                            this.commonState,
+                            subtask,
+                            regionX << this.shift,
+                            ((regionX + 1) << this.shift) - 1,
+                            regionZ << this.shift,
+                            ((regionZ + 1) << this.shift) - 1,
```
dordsor21, line 136: Like below, do we want to bound this to the true region min/max x/z?

worldedit-core/src/main/java/com/fastasyncworldedit/core/queue/implementation/ApplyTask.java
```diff
@@ -0,0 +1,241 @@
...
+            ApplyTask<F> subtask = null;
+            int minRegionX = this.minChunkX >> this.shift;
+            int minRegionZ = this.minChunkZ >> this.shift;
+            int maxRegionX = this.maxChunkX >> this.shift;
+            int maxRegionZ = this.maxChunkZ >> this.shift;
+            // This task covers multiple regions. Create one subtask per region
+            for (int regionX = minRegionX; regionX <= maxRegionX; regionX++) {
+                for (int regionZ = minRegionZ; regionZ <= maxRegionZ; regionZ++) {
+                    if (ForkJoinTask.getSurplusQueuedTaskCount() > Settings.settings().QUEUE.PARALLEL_THREADS) {
+                        // assume we can do a bigger batch of work here - the other threads are busy for a while
```
dordsor21, line 120: I would replace can with should. Whilst it's a trade-off between cross-thread mask access, I would still call higher parallelism the "wanted" outcome

worldedit-core/src/main/java/com/fastasyncworldedit/core/queue/implementation/ApplyTask.java
```diff
@@ -0,0 +1,241 @@
...
+        this.maxChunkX = maxChunkX;
+        this.minChunkZ = minChunkZ;
+        this.maxChunkZ = maxChunkZ;
+        this.before = before;
+        this.shift = Math.max(0, higherShift - SHIFT_REDUCTION);
+    }
+
+    @Override
+    protected void compute() {
+        if (this.minChunkX != this.maxChunkX || this.minChunkZ != this.maxChunkZ) {
```
dordsor21, line 110: `shift == 0` ? Or I haven't worked it out right

worldedit-core/src/main/java/com/fastasyncworldedit/core/queue/implementation/ApplyTask.java
```diff
@@ -0,0 +1,241 @@
...
+            int maxRegionZ = this.maxChunkZ >> this.shift;
+            // This task covers multiple regions. Create one subtask per region
+            for (int regionX = minRegionX; regionX <= maxRegionX; regionX++) {
+                for (int regionZ = minRegionZ; regionZ <= maxRegionZ; regionZ++) {
+                    if (ForkJoinTask.getSurplusQueuedTaskCount() > Settings.settings().QUEUE.PARALLEL_THREADS) {
+                        // assume we can do a bigger batch of work here - the other threads are busy for a while
+                        processRegion(regionX, regionZ, this.shift);
+                        continue;
+                    }
+                    if (this.shift == 0 && !this.commonState.region.containsChunk(regionX, regionZ)) {
```
dordsor21, line 124: Is it actually possible to reach here with shift = 0?

worldedit-core/src/main/java/com/fastasyncworldedit/core/queue/implementation/ApplyTask.java
```diff
@@ -0,0 +1,241 @@
...
+                    this.commonState.full
+            );
+        } catch (Throwable t) {
+            if (t instanceof FaweException faweException) {
+                Fawe.handleFaweException(this.commonState.faweExceptionReasonsUsed, faweException, LOGGER);
+            } else if (t.getCause() instanceof FaweException faweException) {
+                Fawe.handleFaweException(this.commonState.faweExceptionReasonsUsed, faweException, LOGGER);
+            } else {
+                throw t;
+            }
```
dordsor21, line 216: This looks like it should be correct to me

worldedit-core/src/main/java/com/fastasyncworldedit/core/queue/implementation/QueueHandler.java
```diff
@@ -50,7 +51,13 @@ public abstract class QueueHandler implements Trimable, Runnable {
             Settings.settings().QUEUE.PARALLEL_THREADS,
             new FaweForkJoinWorkerThreadFactory("FAWE Fork Join Pool Primary - %s"),
             null,
-            false
+            false,
+            Settings.settings().QUEUE.PARALLEL_THREADS,
```
dordsor21, line 55: The previous maximum pool size would have been the default of (what seems to be) `0x7fff` (32767) - is this change made on purpose? The secondary pool would have this default value

worldedit-core/src/main/java/com/fastasyncworldedit/core/queue/implementation/QueueHandler.java
```diff
@@ -50,7 +51,13 @@ public abstract class QueueHandler implements Trimable, Runnable {
...
+            Settings.settings().QUEUE.PARALLEL_THREADS,
+            0,
+            pool -> true,
```
dordsor21, line 58: Likewise is there a reason for changing the default here? It does "undo" the above change, but is there a reason for going about it this way?

dordsor21 (commented):
I can see a scenario where we end up with one ApplyTask instance effectively only utilising one thread for a 512x512 region of the edit, causing the whole edit to wait for it when everything else is finished. I don't know if there's a particular way around that, but in a  complex operation this could be quite a perceived performance hit

worldedit-core/src/main/java/com/fastasyncworldedit/core/queue/implementation/QueueHandler.java
SirYwell: Yes, I didn't find any better solution to this (besides e.g. using a semaphore per edit), but otherwise, waiting for chunk loads will increase the allowed numbers of threads, and if there are more tasks enqueued, this will cause more threads to be started. Those threads will process tasks and might end up waiting for a chunk to load, resulting in even more threads. I often had around as much threads as chunks in my tests, so this hard limit looked like a good solution. As we now allow processing multiple edits on one thread, this also won't cause other edits to be stalled.

I'm open for alternatives though.

worldedit-core/src/main/java/com/fastasyncworldedit/core/queue/implementation/QueueHandler.java
SirYwell: This is to prevent RejectedExecutionExceptions when the pool wants to create more threads due to threads waiting for chunks to load

dordsor21: Are the threads created for the async chunk load counted for the pool's thread limit then?

dordsor21: Yeah it makes sense with the change above

worldedit-core/src/main/java/com/fastasyncworldedit/core/queue/implementation/QueueHandler.java
dordsor21: I wonder if it's worth replacing the Future extension in the chunk flow with ForkJoinTask - we can then look at ways to better implement the chunk async await (alongside the earlier PR to not await chunk load when calling)

worldedit-core/src/main/java/com/fastasyncworldedit/core/queue/implementation/ApplyTask.java
SirYwell: ` shift == 0` basically means "we're splitting this task into per-chunk-tasks now, so the current task could still cover multiple chunks

worldedit-core/src/main/java/com/fastasyncworldedit/core/queue/implementation/ApplyTask.java
SirYwell: It might help for certain region implementations, but I wanted to keep this change simple for now in case people encounter problems with it. I guess this could be a future improvement.

SirYwell:
> I can see a scenario where we end up with one ApplyTask instance effectively only utilising one thread for a 512x512 region of the edit, causing the whole edit to wait for it when everything else is finished. I don't know if there's a particular way around that, but in a complex operation this could be quite a perceived performance hit

Yes, that's possible in theory, and the heuristic when to split is rather simple currently. In practice, it would mean the thread pool is already busy *and* there are even more tasks waiting. We could e.g. force splitting on 512x512 regions, or check `ForkJoinTask.getSurplusQueuedTaskCount() > Settings.settings().QUEUE.PARALLEL_THREADS * f(this.shift)` (where `f` is some "damping" function) to make that case even less likely.

I also though about re-checking the heuristics during processing, but that complicates things too much I think.

worldedit-core/src/main/java/com/fastasyncworldedit/core/queue/implementation/QueueHandler.java
dordsor21: I suppose for this PR it's a bit out of scope, but I don't think such a change would be a "breaking"/v3 change anyway as none of this is particularly API

dordsor21:
> > I can see a scenario where we end up with one ApplyTask instance effectively only utilising one thread for a 512x512 region of the edit, causing the whole edit to wait for it when everything else is finished. I don't know if there's a particular way around that, but in a complex operation this could be quite a perceived performance hit
> 
> Yes, that's possible in theory, and the heuristic when to split is rather simple currently. In practice, it would mean the thread pool is already busy _and_ there are even more tasks waiting. We could e.g. force splitting on 512x512 regions, or check `ForkJoinTask.getSurplusQueuedTaskCount() > Settings.settings().QUEUE.PARALLEL_THREADS * f(this.shift)` (where `f` is some "damping" function) to make that case even less likely.
> 
> I also though about re-checking the heuristics during processing, but that complicates things too much I think.

If we are processing a larger region by smaller region at a time in the first place, rather than the for x; for z I think it would be much more reasonable to check if we want to parallelise further at the end of the completion of each subregion, and can therefore submit the rest. E.g. create a queue at the start of a region's processing and just submit the remaining as/when needed

SirYwell:
> If we are processing a larger region by smaller region at a time in the first place, rather than the for x; for z I think it would be much more reasonable to check if we want to parallelise further at the end of the completion of each subregion, and can therefore submit the rest. E.g. create a queue at the start of a region's processing and just submit the remaining as/when needed

Hm I guess processing chunks using a [hilbert curve](https://en.wikipedia.org/wiki/Hilbert_curve) might be useful there. Otherwise, just always splitting and then unforking is basically that, and the overhead of creating that many tasks eagerly is most likely not a problem.

dordsor21:
> I reworked the heuristic to directly process a region a bit:
> 
> * With larger shifts, we fork more aggressive (e.g. on the highest level where we split into 512x512 (blocks) regions, we need to split off 32 tasks (32 region files) that aren't taken by other threads yet before processing a full region directly)
> * Even 32x32 (blocks) regions are split further unless there are more than 3 tasks in the queue of the current thread
> 
> With this design, I think it is very unlikely that we process overly large regions that take more time than what the pool can process currently. We can explore more dynamic approaches in future, but I'd like to start with a mainly simple solution.

I suppose it's hard to make much judgement without some kinda of usage statistics from servers with a lot of players on. I don't know if we would want to effectively enforce single-threaded operation per edit. For now it's probably fine I suppose, but worth seeing if we can keep an eye on this in bstats somehow?

SirYwell:
> I suppose it's hard to make much judgement without some kinda of usage statistics from servers with a lot of players on. I don't know if we would want to effectively enforce single-threaded operation per edit. For now it's probably fine I suppose, but worth seeing if we can keep an eye on this in bstats somehow?

It's generally difficult to tell for me how people are using FAWE (which patterns, masks, brushes, typical edit sizes, etc) so I'd like to explore some kind of observability that can be shared (probably similar to `/fawe debugpaste`). But running into the scenario where an edit becomes single-threaded when multiple cores are available has an extremely low likelihood (previous long-running tasks would need to take up all but one thread and end all basically at the same time). I'll try to work on something more dynamic in the future, but that will most likely increase complexity and it might take me some weeks.

dordsor21:
Yeah I think the groundwork here is probably worth a merge for now. As it is I don't know that large edits would be done very often, and I do also wonder if perhaps for small edits (<ncpu chunks) we should just throw it in single-thread anyway

dordsor21 (approved):

NotMyFault (approved):

PR #3105, opened 2025-02-09
feat: remove char block sections array

dordsor21:
- replace with "forced" memory access order to eliminate race conditions between sections and blocks arrays

### DIFF
--- a/worldedit-core/src/main/java/com/fastasyncworldedit/core/queue/implementation/blocks/CharBlocks.java
+++ b/worldedit-core/src/main/java/com/fastasyncworldedit/core/queue/implementation/blocks/CharBlocks.java
@@ -18,32 +18,13 @@ public abstract class CharBlocks implements IBlocks {
     protected static final Section FULL = new Section() {
         @Override
         public char[] get(CharBlocks blocks, int layer, char[] arr) {
-            if (arr == null) {
-                // Chunk probably trimmed mid-operations, but do nothing about it to avoid other issues
-                synchronized (blocks.sectionLocks[layer]) {
-                    LOGGER.warn("Unexpected null section, please report this occurrence alongside a debugpaste.");
-                    return getSkipFull(blocks, layer, false);
-                }
-            }
             return arr;
         }
 
         // Ignore aggressive switch here.
         @Override
         public char[] get(CharBlocks blocks, int layer, char[] arr, boolean aggressive) {
-            if (arr == null) {
-                // Chunk probably trimmed mid-operations, but do nothing about it to avoid other issues
-                synchronized (blocks.sectionLocks[layer]) {
-                    LOGGER.warn("Unexpected null section, please report this occurrence alongside a debugpaste.");
-                    return getSkipFull(blocks, layer, aggressive);
-                }
-            }
             return arr;
-        }
-
-        @Override
-        public boolean isFull() {
-            return true;
         }
     };
     protected static final Section EMPTY = new Section() {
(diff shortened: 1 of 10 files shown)


worldedit-core/src/main/java/com/fastasyncworldedit/core/queue/implementation/blocks/CharBlocks.java
```diff
@@ -18,12 +17,11 @@ public abstract class CharBlocks implements IBlocks {
 
     protected static final Section FULL = new Section() {
         @Override
-        public char[] get(CharBlocks blocks, int layer) {
-            char[] arr = blocks.blocks[layer];
+        public char[] get(CharBlocks blocks, int layer, char[] arr) {
             if (arr == null) {
```
SirYwell, line 21: in FULL, the array is never null if we always call it with a null check for that array directly before. I guess the warnings could be removed?

worldedit-core/src/main/java/com/fastasyncworldedit/core/queue/implementation/blocks/CharBlocks.java
```diff
@@ -296,12 +257,32 @@ static char[] getSkipFull(CharBlocks blocks, int layer, boolean aggressive) {
...
+        public abstract boolean isFull();
+
+        public final char get(CharBlocks blocks, int layer, int index, char[] data) {
+            int normalized = layer - blocks.minSectionPosition;
+            char[] section = get(blocks, normalized, data);
+            if (section == null) {
+                synchronized (blocks.sectionLocks[normalized]) {
+                    blocks.reset(layer);
+                    section = EMPTY.get(blocks, normalized, data, false);
+                }
```
SirYwell, line 276: This shouldn't be possible anymore from my understanding

worldedit-core/src/main/java/com/fastasyncworldedit/core/queue/implementation/blocks/CharBlocks.java
```diff
@@ -296,12 +257,32 @@ static char[] getSkipFull(CharBlocks blocks, int layer, boolean aggressive) {
...
+            }
+            return section[index];
+        }
+
+        public final synchronized void set(CharBlocks blocks, int layer, int index, char value, char[] data) {
```
SirYwell, line 281: this `synchronized` looks a bit odd too, is it needed?

worldedit-core/src/main/java/com/fastasyncworldedit/core/queue/implementation/blocks/CharBlocks.java
```diff
@@ -296,12 +257,32 @@ static char[] getSkipFull(CharBlocks blocks, int layer, boolean aggressive) {
...
-                blocks.sections[layer] = FULL;
-            }
             return arr;
         }
 
+        abstract char[] get(CharBlocks blocks, int layer, char[] data);
+
+        abstract char[] get(CharBlocks blocks, int layer, char[] data, boolean aggressive);
+
+        public abstract boolean isFull();
```
SirYwell, line 267: `isFull` seems to be unused now

worldedit-core/src/main/java/com/fastasyncworldedit/core/queue/implementation/blocks/CharBlocks.java
dordsor21: Yeah tbh I just did only the changes required to test that stuff didn't implode

SirYwell (approved):

NotMyFault (approved):

PR #3111, opened 2025-02-13
chore: don't prepend FAWE prefix to WorldGuard messages

### LINKED ISSUE #3110: Get rid of "(FAWE)" prefix in WorldGuard messages
Kinda bug, kinda feature. ~~Somehow~~ a `(FAWE)` prefix sneaks into WorldGuard messages, which leads to confusion of users. This should be removed and WorldGuard vanilla chat formatting should be used when running alongside FAWE.

Most certainly due to FAWE injecting it's prefix in `BukkitPlayer#print(Component)`

PierreSchwang:
## Overview

Fixes #3110 

## Description
Checks if the caller is WorldGuard, and don't prepend the `(FAWE)` prefix to the message. Requires hidden frames for stuff like https://github.com/EngineHub/WorldGuard/blob/version/7.0.x/worldguard-core/src/main/java/com/sk89q/worldguard/commands/region/RegionCommands.java#L496

```[tasklist]
### Submitter Checklist
```

### DIFF
--- a/worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitPlayer.java
+++ b/worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitPlayer.java
@@ -77,6 +77,9 @@ public class BukkitPlayer extends AbstractPlayerActor {
     private final WorldEditPlugin plugin;
     //FAWE start
     private final PermissionAttachment permAttachment;
+    // We require the StackWalker to show hidden frames, due to the usage of the c.sk89q.we...AsyncCommandBuilder in WorldGuard
+    // (in that case, all WG references are hidden in lambdas and not shown in the default walker)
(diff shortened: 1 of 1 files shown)


dordsor21:
Is it potentially better to search for it only being WE/FAWE ? This could be a problem on some addon plugins, but I can't see it being an issue if we ensure users know what plugin their brush/etc is from

PierreSchwang:
Yeah, sounds more reasonable. Can just dropUntil com.sk89q.we / com.fawe and check if the next / first frame is Bukkit I guess

NotMyFault (approved):

dordsor21:
Quick thought, is it worth adding an overload to skip having to stack walk when we know messages are from FAWE (somehow?)

PierreSchwang:
Could be possible - would have to look further into it. 

iirc, I also tried to implement the more generic approach (like, don't just handle WG but also other third party plugins) - but there has been some reason why that did not work (don't remember what it was exactly tbf, could've been the builtin stuff like message boxes). But WG alone would be fine as well ig

Incluuu:
bump, when fix?

PR #3128, opened 2025-02-28
chore: remove 1.21.3

dordsor21:
As discussed, the chunk/packet issues are specific to 1.21.3, there's no point in expending effort when it's nonexistent in 1.21.4, and so let's not support it


SirYwell (approved):

NotMyFault (approved):

v404:
xd

Issue #1061, opened 2021-05-09
//undo does not work on tool tree


SecretlyJealous:
### Server Implementation

Paper

### Server Version

1.16.5

### Describe the bug

A `/tool tree` cannot be undone:
![2021-05-09_15 20 53](https://user-images.githubusercontent.com/10485646/117575556-4613e480-b0da-11eb-8d16-d0a3b0abec35.png)

`//brush inspect` does not log it either:
![2021-05-09_15 21 06](https://user-images.githubusercontent.com/10485646/117575590-693e9400-b0da-11eb-89fd-1f86545dffc1.png)

### To Reproduce

1. Do /tool tree
2. Do //undo
3. Do //brush inspect

### Expected behaviour

//undo should undo the edit

### Error log (if applicable)

https://pastebin.com/UPuQuLee

### Fawe Debugpaste

https://pastebin.com/a0KgYVX0

### Fawe Version

FastAsyncWorldEdit version 1.16-673;00d689a

### Checklist

- [X] I have included a Fawe debugpaste.
- [X] I am using the newest build from https://ci.athion.net/job/FastAsyncWorldEdit-1.16/ and the issue still persists.

NotMyFault:
Resolved in https://github.com/EngineHub/WorldEdit/issues/1745, incase someone is interesting in submitting a pr to paper.

Issue #151, opened 2019-05-18
Brush not working


PixelNotFound:
# Bug report for FastAsyncWorldEdit 1.13.2

**[REQUIRED] FastAsyncWorldEdit Version Number:** 

fawe version=> [10:33:30 INFO]: FastAsyncWorldEdit-1.13-110 by Empire92
[10:33:30 INFO]: ------------------------------------
[10:33:30 INFO]: - DATE: May 12, 2019 12:00:00 AM
[10:33:30 INFO]: - COMMIT: e72a5cf
[10:33:30 INFO]: - BUILD: 110
[10:33:30 INFO]: - PLATFORM: bukkit
[10:33:30 INFO]: ------------------------------------
[10:33:30 INFO]: Platforms:
[10:33:30 INFO]: - Bukkit
[10:33:30 INFO]: Capabilities:
[10:33:30 INFO]: - GAME_HOOKS: Bukkit
[10:33:30 INFO]: - CONFIGURATION: Bukkit
[10:33:30 INFO]: - USER_COMMANDS: Bukkit
[10:33:30 INFO]: - PERMISSIONS: Bukkit
[10:33:30 INFO]: - WORLDEDIT_CUI: Bukkit
[10:33:30 INFO]: - WORLD_EDITING: Bukkit
[10:33:30 INFO]: ------------------------------------
[10:33:30 INFO]: Wiki: https://github.com/boy0001/FastAsyncWorldedit/wiki

Issue is on both breaking and normal/
**[REQUIRED] Spigot/Paper Version Number:** 

[10:33:53 INFO]: Previous version: git-Paper-625 (MC: 1.13.2)
[10:33:53 INFO]: Checking version, please wait...
[10:33:53 INFO]: You are running the latest version
**Links to config.yml and config-legacy.yml file:** 

**[REQUIRED] Description of the problem:**

When someone uses brush it outputs alot of error. https://hastebin.com/hajuhoyibi.bash
**Plugins being used on the server:**

**How to replicate**:

/br sphere #copy
**Checklist**:

- [x] I included all information required in the sections above
- [x] I made sure there are no duplicates of this report [(Use Search)](https://github.com/IntellectualSites/FastAsyncWorldEdit-1.13/issues?utf8=%E2%9C%93&q=is%3Aissue)
[...]

Issue #103, opened 2019-04-06
catenary is a little off


AeSix:
# Bug report for FastAsyncWorldEdit 1.13.x

**Debug paste link**:

https://incendo.org/paste/view/34176f64484b4410ac5fb46e530c94da

**Description of the problem:**

https://puu.sh/DaZH0/89ba5becd1.png
![javaw_2019-04-06_16-35-06](https://user-images.githubusercontent.com/5191757/55675026-75681f80-588a-11e9-992e-73e0da17d933.png)

**How to replicate**:

create catenary line

**Checklist**:

- [x] I included a `/fawe debugpaste` link
- [x] I made sure there aren't duplicates of this report [(Use Search)](https://github.com/IntellectualSites/FastAsyncWorldEdit-1.13/issues?utf8=%E2%9C%93&q=is%3Aissue)
- [x] I made sure I am using an up-to-date version of [FAWE for 1.13.x](https://ci.athion.net/job/FAWE-1.13/)
- [x] I made sure the bug/error is not caused by any other plugin

PR #3130, opened 2025-03-02
Set Extent on the Copy -m Mask flag

Zeranny:
## Overview
While developing ezEdits @eztaK-red and I noticed that not using a MaskTraverser to set the extent on a mask results in significantly slower performance.

We assumed that this could also be why we noticed slow performance in the `-m` flag for `//copy`, and that does appear to be the case.

There may be other instances of masks in FAWE not having an extent set where performance could be improved, but I didn't want to get ahead of myself in case there were issues with the way that I have fixed it here.
Especially as during testing it appears older versions of FAWE (2.11.1) were even slower, so changes since then may have already had a positive impact on `-m` performance.

## Description
Sets the extent for the `finalMask` in the `createCopy()` method used by `//copy` which (going by previous FAWE commits) avoids using a WorldWrapper.

Rough runtime comparison using `-m #solid,>[!air],[/[0][90]],[|air]` (Before & After fix):

Copying a large selection with a "complex" mask & pasting (After fix):

### Submitter Checklist

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/command/ClipboardCommands.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/command/ClipboardCommands.java
@@ -306,6 +306,7 @@ public class ClipboardCommands {
         }
         final Mask firstSourceMask = mask != null ? mask : sourceMask;
         final Mask finalMask = MaskIntersection.of(firstSourceMask, new RegionMask(allowedRegion)).optimize();
+        new MaskTraverser(finalMask).setNewExtent(editSession);
         if (finalMask != Masks.alwaysTrue()) {
             copy.setSourceMask(finalMask);
         }


eztaK-red:
Investigating a bit further:

Paper-1.20.6 (tested #151) or older:
  newest FAWE (#1061): //copy -m: 52s
  this-PR's FAWE version: //copy -m: <1s

Paper-1.21 (tested #103) or newer:
  newest FAWE (#1061): //copy -m: <1s
  this-PR's FAWE version: //copy -m: <1s

This bug, of `//copy -m` and `//cut -m` taking significantly longer than it should, appears to affect even the newest FAWE versions if they are running on Paper 1.20.6 or older.

Furthermore, somehow, it's much better when the server is running on 1.21, independent of the FAWE version.

Zeranny:
Re-activated this PR. I had set it back to draft as it was unclear exactly what scenarios led to slow performance.
From my own testing it does appear that on MC 1.21.4 the performance for very large copies is still sometimes faster with this change, but that my be a bit random.

I would appreciate any further insight into the underlying performance difference :)

SirYwell:
I don't know if you tested that on the same FAWE build on the different versions, and also with the same configuration. There were quite a few changes in the last months that could affect performance. Can you provide profiles (e.g. through spark, or the IntelliJ profiler) especially of the slow case?

Other than that: Would it make sense to move your change into `setSourceMask` instead?

SirYwell:
Hm, from a quick look, I assume that chunks are loaded in the fast case but not in the slow one. Could that be the case?

eztaK-red:
Is it
that in 1.20.6
`net.minecraft.world.level.Level.getChunk()` calls
`net.minecraft.world.level.ServerChunkCache.getChunk()`
Screenshot_20250304_193942

while in 1.21
`net.minecraft.world.level.Level.getChunk()` calls
`net.minecraft.server.level.ServerChunkCache.getChunkAtIfLoadedImmediately()`
image
?

SirYwell:
Interesting - though it seems we currently directly access the world here, which is what we should avoid and what this PR addresses I guess.

Zeranny:
I will test out your suggestion to move this to setSourceMask tomorrow and update :)

SirYwell:
After taking a closer look, setSourceMask is probably not the right place. But we have similar code already here: https://github.com/IntellectualSites/FastAsyncWorldEdit/blob/f3860618ad157e406a8b6999348f8477317defbe/worldedit-core/src/main/java/com/sk89q/worldedit/function/operation/ForwardExtentCopy.java#L352-L355
This only affects situations with transformations involved, where https://github.com/IntellectualSites/FastAsyncWorldEdit/blob/f3860618ad157e406a8b6999348f8477317defbe/worldedit-core/src/main/java/com/sk89q/worldedit/function/operation/ForwardExtentCopy.java#L414-L420
is missing that part, basically. Adding it in that if would be good I think.

dordsor21:
@Zeranny bumping this

PR #3142, opened 2025-03-09
Fix //hollow not behaving correctly

### LINKED ISSUE #3083: //hollow incorrect behavior
### Server Implementation

Paper

### Server Version

1.21.3

### Describe the bug

Actual version is 1.21.4. 
//hollow is behaving very incorrectly. It fails to properly hollow out the inside, and it removes parts of the "shell"

### To Reproduce

Get some 

### Expected behaviour

//hollow reduces the selected object to a 1-block thick shell with a hollow interior (assuming the given object has a clearly defined inside and outside)

### Screenshots / Videos

Image

Image

Image

### Error log (if applicable)

_No response_

### Fawe Debugpaste

https://athion.net/ISPaster/paste/view/1d5690aa00a84e72992fb85b229a5be3

### Fawe Version

2.12.4-SNAPSHOT-1024;69d8543

### Checklist

### Anything else?

_No response_

eztaK-red:
## Overview

Fixes #3083

## Description
Turning the mutable position vectors of the region iterator to immutable ones, to allow the original implementation of the hollowOutRegion to function as expected.

### Submitter Checklist


eztaK-red:
Result of //hollow in latest FAWE versions   vs   result with fix from this PR

worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
```diff
@@ -3298,7 +3298,7 @@ public int hollowOutRegion(Region region, int thickness, Pattern pattern, Mask m
                 outer:
                 for (BlockVector3 position : region) {
                     for (BlockVector3 recurseDirection : recurseDirections) {
-                        BlockVector3 neighbor = position.add(recurseDirection);
+                        BlockVector3 neighbor = position.toImmutable().add(recurseDirection);
```
dordsor21, line 3301: Nice find. I would say to use the mutable bv3 as is done in the code above to avoid creating many BV3s here though

dordsor21 (approved):
Assuming this still works (don't see why not), lgtm, thanks!

eztaK-red:
Oh hold up, there's a specific case for which it doesn't work

eztaK-red:
The fix worked. The case I was just talking about was something separate. `recurseHollow` was incorrectly marking north-east-upwards concave corners as inside. Adding toImmutable() to a bv3 fixed it. In this case, it is not obvious to me whether there is a better way.

dordsor21:
> The fix worked. The case I was just talking about was something separate. `recurseHollow` was incorrectly marking north-east-upwards concave corners as inside. Adding toImmutable() to a bv3 fixed it. In this case, it is not obvious to me whether there is a better way.

Make it not immutable after adding, this current way you're creating extra BV3s

worldedit-core/src/main/java/com/sk89q/worldedit/EditSession.java
```diff
@@ -3298,7 +3298,7 @@ public int hollowOutRegion(Region region, int thickness, Pattern pattern, Mask m
                 outer:
                 for (BlockVector3 position : region) {
                     for (BlockVector3 recurseDirection : recurseDirections) {
-                        BlockVector3 neighbor = position.add(recurseDirection);
+                        BlockVector3 neighbor = mutable.setComponents(position).add(recurseDirection);
```
dordsor21, line 3301: I guess these should have `//FAWE start - mutable` `//FAWE end` around them

dordsor21:
> Like this? As far as I can see, //hollow now/still works as it should.

Yeah I don't see why it would need to be immutable here - the queue does not actually store the BV3 and stores as an index in a long array

dordsor21:
You added the comments in the wrong place - where you added is already covered. Should be around 3301 and 3316

SirYwell (approved):
Looks fine. I wonder if we should change MutableBV3 to create an immutable instance on `add` and instead have specific mutating variants instead. But that's out of scope for now.

dordsor21:
> Looks fine. I wonder if we should change MutableBV3 to create an immutable instance on `add` and instead have specific mutating variants instead. But that's out of scope for now.

Mm probably, but I think that should be a forcibly breaking change in v3; relocating/renaming the class e.g.

PR #3152, opened 2025-03-20
Fix crash due to data race when removing block entity

### LINKED ISSUE #3151: Server-Crash
### Server Implementation

Paper

### Server Version

1.21.1

### Describe the bug

If you were to remove or even change blocks which seem to try to update their neighbours, which are also getting removed by an operation, e.g. `//air`, there are chances the server will crash, with an `Exception while updating neighbours`-Error.

This also seems to be in direct connection with a PlotSquared Bug, as the error is basically the same, and from my understanding (this may be wrong), if FAWE is installed on the server with PS, it will be used, e.g. to delete plots.
https://github.com/IntellectualSites/PlotSquared/issues/4337

### To Reproduce

- Create an area of e.g. Sculk Sensors, with a layer of Trapdoors on top, stack it horizontally, e.g. to create a rectangle
- Select the entire region with the sculk sensors and trapdoors
- Trigger the sculk sensors by walking on top of the trapdoors
- Remove the area while the sculk sensors / trapdoors are still triggering the neighbour blocks, e.g. via `//air`
- If nothing happens, repeat

### Expected behaviour

The server should continue working, and the blocks should have all been removed, without any crash.

### Screenshots / Videos

Image
Image
Image

### Error log (if applicable)

https://gist.github.com/gamecrash534/f136283f2a55884c850e5f8acdb2c083

### Fawe Debugpaste

https://athion.net/ISPaster/paste/view/fec39af532b64718bb62e49b5cc5971c

### Fawe Version

FastAsyncWorldEdit version 2.13.1-SNAPSHOT-1065;cca3b71

### Checklist

### Anything else?

Sorry, if the description is a bit inaccurate. I couldn't really think of anything else ;)

SirYwell:
## Overview

Fixes #3151

## Description

This is rather a quick fix for the server crash. Performance impact seems to be okay, but we might want to take a closer look in future. We could probably also use Minecraft's new tick freezing mechanism instead. The beacon special-casing can be removed too probably.

### Submitter Checklist

### DIFF
--- a/worldedit-bukkit/adapters/adapter-1_20_2/src/main/java/com/sk89q/worldedit/bukkit/adapter/impl/fawe/v1_20_R2/PaperweightGetBlocks.java
+++ b/worldedit-bukkit/adapters/adapter-1_20_2/src/main/java/com/sk89q/worldedit/bukkit/adapter/impl/fawe/v1_20_R2/PaperweightGetBlocks.java
@@ -13,6 +13,7 @@ import com.fastasyncworldedit.core.nbt.FaweCompoundTag;
 import com.fastasyncworldedit.core.queue.IChunkSet;
 import com.fastasyncworldedit.core.util.MathMan;
 import com.fastasyncworldedit.core.util.NbtUtils;
+import com.fastasyncworldedit.core.util.TaskManager;
 import com.fastasyncworldedit.core.util.collection.AdaptedMap;
 import com.sk89q.worldedit.bukkit.BukkitAdapter;
 import com.sk89q.worldedit.bukkit.BukkitEntity;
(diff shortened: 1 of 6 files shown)


dordsor21:
Is it better worth to locate the tiles async and only perform the removes synchronously, if required, rather than needing a (potentially pointless) main-thread task on every chunk call?

SirYwell:
Using the map returned by `getBlockEntities()` in any way isn't thread-safe to begin with, so I'm not sure if we can skip it at all. If we have GET data already we could check if any of those blocks has a block entity, but that could also be out-of-date information...

dordsor21:
> Using the map returned by `getBlockEntities()` in any way isn't thread-safe to begin with, so I'm not sure if we can skip it at all. If we have GET data already we could check if any of those blocks has a block entity, but that could also be out-of-date information...

If the map is empty when we retrieve it I think it's acceptable to skip the main thread call entirely tbh. Otherwise yes I suppose it probably is best to iterate the map in the sync job. Also, with the previous changes made to the futures, I wonder if we want do something similar to how we handle the chunk load waiting by returning a future for the main thread operation, which continues with the operation when complete

SirYwell:
> If the map is empty when we retrieve it I think it's acceptable to skip the main thread call entirely tbh. Otherwise yes I suppose it probably is best to iterate the map in the sync job.

Checking whether the map is empty asynchronously is already flawed, so I don't think we can skip that.

> Also, with the previous changes made to the futures, I wonder if we want do something similar to how we handle the chunk load waiting by returning a future for the main thread operation, which continues with the operation when complete

I was thinking about that too, but that makes it more difficult to get things back to the threads we want to run it on, so I decided to go the easiest route for now. Maybe we should think about switching to e.g. an actor model in future, that would allow far more flexibility in such scenarios.

dordsor21:
> Checking whether the map is empty asynchronously is already flawed, so I don't think we can skip that.

Arguably the same can be said for the whole operation at that point. The rest of FAWE assumes the chunk isn't modified after reading (cached GET for masks could be cached a while) and halting everything awaiting a potentially empty main thread task feels suboptimal.

> I was thinking about that too, but that makes it more difficult to get things back to the threads we want to run it on, so I decided to go the easiest route for now. Maybe we should think about switching to e.g. an actor model in future, that would allow far more flexibility in such scenarios.

Currently after the chunk async load we just "resubmit" the task and return that future. Here we would submit a main thread task instead, which after completion would perform the resubmission to the FAWE queue(s)

jquery-package:
Any news on this?

PR #3173, opened 2025-04-13
feat: implement "mutable" transform holder for random rotations

### LINKED ISSUE #3129: Function not add, while it is announced in the wiki.

Hello, I have a question because I’m trying to use the// schematic loadall. But I can’t find or set the -r to apply random rotation. When I place it, it puts unknown clipboard format. while on the wiki he says that -r can be used.

The function is not used while on the wiki it is marked as usable. 
Version 1.21.4.

Thanks you

Add la fonction rotate random.

### Anything else?

_No response_

dordsor21:
- also implement random rotation to schematic loadall
 - "mutable" transform allows random rotation of schematics outside of clipboard brush
 - fixes #3129

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/math/transform/Transform.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/math/transform/Transform.java
@@ -58,4 +58,18 @@ public interface Transform {
      */
     Transform combine(Transform other);
 
+    //FAWE start- mutating transforms
+
+    /**
+     * Perform a mutation on this transform, if supported. Else, does nothing. This mutation will depend on the specific
+     * transform implementation.
+     * <p>
+     * Implementation detail: it may be possible for this method to be called multiple times before an operation actually occurs.
+     *
+     * @since TODO
+     */
+    default void mutate() {
+    }
+    //FAWE end
+
 }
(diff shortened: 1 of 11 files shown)


PierreSchwang (approved):
Looks reasonable

worldedit-core/src/main/java/com/sk89q/worldedit/math/transform/Transform.java
```diff
@@ -58,4 +58,18 @@ public interface Transform {
...
+
+    /**
+     * Perform a mutation on this transform, if supported. Else, does nothing. This mutation will depend on the specific
+     * transform implementation.
+     * <p>
+     * Implementation detail: it may be possible for this method to be called multiple times before an operation actually occurs.
+     *
+     * @since TODO
+     */
+    default void mutate() {
```
SirYwell, line 71: I'm not sure if I like this design, it's somewhat invasive. Would it make sense to have a static method in `MutatingOperationTransform` instead that handles that part instead?

SirYwell:
Hmm how does that deal with multithreading?

dordsor21:
> Hmm how does that deal with multithreading?

It doesn't, but currently there should be no reason for there to be any kind of contention, and regardless multi-threading should probably be handled implementation-specifically

SirYwell:
What about `RandomFullClipboardPattern`? Doesn't that cause multiple threads to update the same transform instance?

SirYwell (approved):

PR #3182, opened 2025-04-26
fix: attempt to retrieve online from offline player for perm attachments

dordsor21:


dordsor21:
Thoughts?

image

MsMaciek123:
@dordsor21 by offline player i meant player instance that returns false for isOnline(), not necessarily OfflinePlayer instance.
See https://github.com/IntellectualSites/FastAsyncWorldEdit/issues/3183

dordsor21:
> @dordsor21 by offline player i meant player instance that returns false for isOnline(), not necessarily OfflinePlayer instance. See #3183

Altered

MsMaciek123:
I don't think returning null here is a good idea, especially if function isn't marked with nullable itself. I think it's better to return noop attachment or proper attachment but not add it to the hashmap.

dordsor21:
> I don't think returning null here is a good idea, especially if function isn't marked with nullable itself. I think it's better to return noop attachment or proper attachment but not add it to the hashmap.

The attachment is only used for setting permissions (e.g. when toggling bypass) so null is acceptable here. I've marked the method as nullable as well

SirYwell (approved):

PierreSchwang (approved):
I hope the WeakMap really works ^^'

PR #3249, opened 2025-07-25
fix: we should just always clone the region in ApplyTask

### LINKED ISSUE #3231: //sel polyhedral broken
### Server Implementation

Paper

### Server Version

1.21.4

### Describe the bug

When I use //sel polyhedral to be exact to select 3 vertex spots and I type //set (blockid) it creates weird effect with air spots.
Its happening after we update the plugin to the newest version.

### To Reproduce

1. type //sel polyhedral
2. Select 3 spots in the triangle shape
3. type //set 2 e.x
4. Summary effect

### Expected behaviour

It should generate smooth terrain as always. As visible below for comparison.

### Screenshots / Videos

Image
Image

### Error log (if applicable)

_No response_

### Fawe Debugpaste

https://athion.net/ISPaster/paste/view/a755fdca354247b38897d67dc529748c

### Fawe Version

2.13.1-SNAPSHOT-1128;67ecacf

### Checklist

### Anything else?

_No response_

dordsor21:
- fixes #3231

### DIFF
--- a/worldedit-core/src/main/java/com/fastasyncworldedit/core/queue/implementation/ApplyTask.java
+++ b/worldedit-core/src/main/java/com/fastasyncworldedit/core/queue/implementation/ApplyTask.java
@@ -23,6 +23,7 @@ class ApplyTask<F extends Filter> extends RecursiveAction implements Runnable {
     private static final int SHIFT_REDUCTION = 1;
 
     private final CommonState<F> commonState;
+    private final Region region;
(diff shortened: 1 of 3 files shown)


PierreSchwang (approved):

dordsor21:
> A bit sad that we actually have Region implementations mutating during contains checks, but I guess we can live with that.

It probably does make a somewhat meaningful difference for the Poly*Region implementations... When they are very large

PR #3257, opened 2025-07-31
feat: start work on paper anti-x-ray support

PierreSchwang:
## Description
Initial work on supporting papers anti x-ray system via SideEffects.
Doesn't work yet, but the actual bitstorage is overriden for affected chunks. No idea why yet. And surrounding chunks are updated (new chunk packets are sent) - no idea why, but that shouldn't happen (though it's just a packet, so not too bad lol)

### Submitter Checklist


worldedit-bukkit/adapters/adapter-1_21_6/src/main/java/com/sk89q/worldedit/bukkit/adapter/impl/fawe/v1_21_6/PaperweightPlatformAdapter.java
```diff
@@ -376,8 +391,19 @@ public static void sendChunk(IntPair pair, ServerLevel nmsWorld, int chunkX, int
                             nmsWorld.getLightEngine(),
                             null,
                             null,
-                            false // last false is to not bother with x-ray
+                            false // false so we can handle anti-X-Ray ourselves
```
SirYwell, line 394: what happens if we just set this to `obfuscateAntiXRay`?

dordsor21: We already don't guarantee chunks being sent before close returns

PR #3260, opened 2025-08-03
Avoid copying already copied arrays in history

SirYwell:
## Description

When writing changes, we already have copies of the get and set data. That means we don't need (and shouldn't) copy it again. This reduces memory allocations in the history by around 90%.

### Submitter Checklist


dordsor21:
Which command(s) was this tested with?

SirYwell:
> Which command(s) was this tested with?

Only `//air`, but bot copying at this point *must* be safe, otherwise the history would be broken already

SirYwell:
I did some more testing (height brush, //replace with several masks) and didn't find any issues beyond what's already broken.
For example the angle mask breaks history already, but that is unrelated to copying the array.

NotMyFault (approved):

PR #3269, opened 2025-08-08
feat: add BlockVector3 radius square to Ellipsoid to reduce rounding

dordsor21:


dordsor21:
> Looks fine, do you have any numbers for comparison or a scenario where the change is noticeable?

Not really, just noticed that the Ellipsoid contain was taking somewhat longer than I would have expected, and this is the only reasonable optimisation

PR #3295, opened 2025-08-26
fix: sponge v3 palette writer with __reserved__

PierreSchwang:
## Overview
Fixes an issue where __reserved__ was written into the block palette for sponge v3 schematics (example would be masked copies).

### Submitter Checklist


dordsor21:
What exactly is the fix? The change is a little abstract aha

PierreSchwang:
Uhm, it was something related to an issue reported on discord. When saving a clipboard with some (maybe specific) mask, the writer would add a minecraft:reserved block state into the palette iirc. 

I think this fix kinda streamlines it with the FastV2Writer? I'll have to look into it again lol. But this should write reserved as air for some reason (I guess because the state cache is setup that way or something)

PierreSchwang:
Ah, yeah: https://github.com/IntellectualSites/FastAsyncWorldEdit/blob/main/worldedit-core/src/main/java/com/fastasyncworldedit/core/extent/clipboard/io/FastSchematicWriterV3.java#L144

I replace the ordinal in the resolver function if it's reserved, but don't use the ordinal to retrieve the string representation. I guess updating / replacing the actually BlockStateHolder could be cleaner?

dordsor21:
Yeah we shouldn't have `__RESERVED__` going into schematics at all ideally

PR #3307, opened 2025-09-18
Added -f flag to prevent heightmap recalculation

B1ue-Bandit:
## Overview
This adds a -f flag to certain operations to prevent heightmap recalculation.

## Description
Pretty straight forward, just prevents heightmap recalculations when used with //replace, //paste, //walls & //cyl
These are just the tools I need, nobody else has asked for this to my knowledge, I just figured I'd create a pull request in case someone decides this is a feature that should be added. You shouldn't be modifying the height map unless you know what you're doing. I just needed rain to be able to pass through certain parts of the world.


SirYwell:
As mentioned on Discord already, heightmaps are just like other side effects. This feature can already be achieved by using `//perf heightmaps off`.

PR #3349, opened 2025-10-24
Fix WallMakeMask for regions not overriding contains(x, z)

SirYwell:
## Description

For some region implementations, precise `contains(x, z)` implementations would require heavy computation, far more heavy than just also passing the Y coordinate. `FuzzyRegion` is a good example for that.
For such regions, `//walls` was broken before. This simple change addresses this problem.

### Submitter Checklist


dordsor21:
Should the fix rather be to overload the contains(x,y,z) or to add them to the relevant regions?

SirYwell:
Overriding `contains(x, z)` should probably be done too, but especially for FuzzyRegion there is no fast way to check whether any y coordinate is included, you'd have to go from minY to maxY potentially. For some other regions it's similarly complex. Generally I think the `contains(x, z)` method does not provide much benefit, maybe we should just deprecate it.

PR #3353, opened 2025-10-26
Remove custom Residence repository

NotMyFault:


worldedit-bukkit/build.gradle.kts
```diff
@@ -33,6 +33,11 @@ repositories {
     maven {
         name = "JitPack"
         url = uri("https://jitpack.io")
+        content {
+            includeGroup("com.github.Zrips")
+            includeGroup("com.github.MilkBowl")
+            includeGroup("com.github.TechFortress")
```
NotMyFault: Content restriction, because I would rather not pull anything else in.

Issue #3359, opened 2025-10-29
Loading 1 wide/long schematics with biomes causes / by zero in FastSchematicReaderV3


ZECHEESELORD:
### Server Implementation

Paper

### Server Version

1.21.8

### Describe the bug

Sponge v3 schematics whose width OR length equals 1 error FAWE’s fast clipboard loader. When the default clipboard factory runs it mutates the mutable dimensions vector that `FastSchematicReaderV3` passes in, decrementing the X/Z size to zero. Later, `indexToPosition` divides by `dimensions.x() * dimensions.z()` while replaying biome data, which throws an ArithmeticException (divide by zero) and aborts the load. The stored NBT is valid; the crash is entirely within the reader pipeline.

### To Reproduce

1. In-game, select a 1x3x1 column and run `//setbiome` desert to ensure biome data exists.
2. Copy the region with `//copy -b`.
3. Save it as a Sponge v3 schematic: `//schem save test sponge.3`.
4. Load the file: `//schem load test`

### Expected behaviour

loading a Sponge v3 schematic- even one that’s only 1xNx1 and carries biome data- completes without any errors

### Error log (if applicable)

https://mclo.gs/crma3pq

### Fawe Debugpaste

https://athion.net/ISPaster/paste/view/0d6c321946f942138fdba69de0ed204e

### Fawe Version

2.14.1-SNAPSHOT-1208+79582f5

### Checklist

- [x] I have included a Fawe debugpaste.
- [x] I am using the newest build from https://ci.athion.net/job/FastAsyncWorldEdit/ and the issue still persists.

PR #3360, opened 2025-10-29
fix FastSchematicReaderV3 throwing / by 0 for 1-wide/long schematics

ZECHEESELORD:
## Overview
Fixes [#3359](https://github.com/IntellectualSites/FastAsyncWorldEdit/issues/3359)

## Description
Ensure FastSchematicReaderV3 passes an immutable dimensions vector to the clipboard factory so the default pipeline can’t mutate the reader’s state and zero out `1xNx1` schematics, preventing a / by 0 when applying biome data.

### Submitter Checklist


PierreSchwang (approved):
interesting, that there was no report of missing clipboard data or arithmetic exceptions yet

ZECHEESELORD:
> interesting, that there was no report of missing clipboard data or arithmetic exceptions yet

Yeah, It's an odd edge case that only really blows up when the schematic actually carries biome data

PierreSchwang:
> > interesting, that there was no report of missing clipboard data or arithmetic exceptions yet
> 
> 
> 
> Yeah, It's an odd edge case that only really blows up when the schematic actually carries biome data

Ah, I see - doesn't really happen with linear palettes for block data.

PR #3365, opened 2025-11-04
setBlockLight should define setBlockLight and resolve a few other minor inconsistencies

R00tB33rMan:
I addressed these in my larger Folia pull request, and because of rejection, I would like to forward these upstream separately. I am confident a lot of these issues aren't intentional. If any of these changes are deliberate, I'd be more than happy to revise!


SirYwell (approved):
Thanks

PierreSchwang (approved):

PR #3405, opened 2025-12-17
Use Java 25 on Jenkins

NotMyFault:
Let's see if I've installed this correctly...

### DIFF
--- a/.github/workflows/build-pr.yml
+++ b/.github/workflows/build-pr.yml
@@ -3,10 +3,10 @@ on: [pull_request]
 jobs:
   build_pr:
     if: github.repository_owner == 'IntellectualSites'
-    runs-on: ${{ matrix.os }}
+    runs-on: ubuntu-latest
     strategy:
       matrix:
-        os: [ubuntu-latest, windows-latest, macos-latest]
+        java: [21, 25]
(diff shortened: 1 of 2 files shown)


NotMyFault:
Screenshot 2025-12-17 at 17 46 49
Edit: worked 🚀

SirYwell:
Given we use the Gradle toolchains feature which uses Java 21, this change will only mean Gradle itself runs on Java 25, right? Would it make sense to switch to Java 25 with the toolchain directly? We set --release 21 still, so I would expect to still have Java 21 compatibility this way.

NotMyFault:
> Given we use the Gradle toolchains feature which uses Java 21, this change will only mean Gradle itself runs on Java 25, right? Would it make sense to switch to Java 25 with the toolchain directly? We set --release 21 still, so I would expect to still have Java 21 compatibility this way.

We can surely do that. The change in the Jenkinsfile solely specifies the JDK to be used during the build, and has no impact on the outcome. Although:

> > Configure project :
Kotlin does not yet support 25 JDK, falling back to Kotlin JVM_24 JVM

I would put that on hold for now, given it's not yet supported by Kotlin DSL.

PR #3421, opened 2026-01-02
Improve BlendBall performance using custom Filter

SirYwell:
## Description

BlendBall is implemented acting on a cube, applying a block-by-block action with no dependencies between these actions. Significant performance overhead is caused by the `EditSession#getBlock(...)` indirection, going through the extent stack and the chunk caching mechanism. On top of that, for larger radii, we still process each block sequentially instead of making use of multiple threads.

By implementing a custom Filter, we lower the overhead of accessing blocks as well as easily introducing parallelism.

### Submitter Checklist


dordsor21:
Presumably the result is no longer deterministic for the same input given the multi-threading?

SirYwell:
> Presumably the result is no longer deterministic for the same input given the multi-threading?

If we read from chunks that were flushed before in the same edit, yes. But if that's a problem for typical brush sizes, then it's rather critical in general.

dordsor21:
As long as it's not "corrupting" I think it's fine

PR #3445, opened 2026-01-26
Quote schematic file names in //schem list 

Madelyyn:
## Description

Adds quotes to schematic file names that would be invalid w/o in the auto complete of //schem list
### Submitter Checklist

### DIFF
--- a/worldedit-core/src/main/java/com/fastasyncworldedit/core/util/StringMan.java
+++ b/worldedit-core/src/main/java/com/fastasyncworldedit/core/util/StringMan.java
@@ -542,4 +542,21 @@ public class StringMan {
         return UUID_PATTERN.matcher(str).find();
     }
 
+    public static String escape(String value) {
+        if (value.isEmpty()) {
+            return "\"\"";
+        }
+
+        boolean quote = value.indexOf(' ') >= 0 || value.indexOf('\t') >= 0;
+        if (!quote) {
+            return value;
+        }
+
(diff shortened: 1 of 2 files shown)


worldedit-core/src/main/java/com/fastasyncworldedit/core/util/StringMan.java
```diff
@@ -542,4 +542,21 @@ public static boolean containsUuid(String str) {
...
+        }
+
+        boolean quote = value.indexOf(' ') >= 0 || value.indexOf('\t') >= 0;
+        if (!quote) {
+            return value;
+        }
+
+        String escaped = value
+                .replace("\\", "\\\\")
+                .replace("\"", "\\\"");
```
dordsor21: Take a look at this formatting again please - ensure you have the .editorconfig, etc. applied from the repo root

Madelyyn: My editor (intellij) is only reformatting the humanReadableByteCountBin function, not the escape one.

SirYwell (changes requested):
(FA)WE currently doesn't allow saving schematics with spaces in their name, even when quoting it. Should that be changed too?

Madelyyn:
> (FA)WE currently doesn't allow saving schematics with spaces in their name, even when quoting it. Should that be changed too?

If you want

PR #3507, opened 2026-04-08
Set Java toolchain to version 21 in build.gradle.kts for bukkit adapter modules for versions: 1.21.4, 1.21.5, 1.21.6, 1.21.9

gaboss44:
Set Java toolchain to version 21 in build.gradle.kts for bukkit adapter modules: 1.21.4, 1.21.5, 1.21.6, 1.21.9

These modules just won't compile with java 25

### Submitter Checklist


SirYwell:
Can you expand on why this is necessary? Why set the toolchain rather than the target/source version? Why aren't the GitHub Actions failing already?

NotMyFault:
@gaboss44 I'm experiencing the same issue locally building the default branch, while this solves the issue for me.
Could you explain why this is needed and what issue we're running into? It seems to be an environmental issue, given it doesn't occur on GH actions.

PR #3548, opened 2026-05-24
fix: release drafter v7

Timongcraft:
## Overview
fix release-drafter v7 changing gh token usage

> If you were using a specific access token using env.GITHUB_TOKEN: $MY_CUSTOM_TOKEN, migrate to the new token: input [...]
see https://github.com/release-drafter/release-drafter/pull/1475

### Submitter Checklist
~~- [ ] New public fields and methods are annotated with `@since TODO`.~~

### DIFF
--- a/.github/workflows/release-drafter.yml
+++ b/.github/workflows/release-drafter.yml
@@ -10,8 +10,8 @@ on:
 jobs:
   update_release_draft:
     if: ${{ github.event_name != 'pull_request' || github.repository != github.event.pull_request.head.repo.full_name }}
+    permissions:
+      contents: write # required for creating draft releases
     runs-on: ubuntu-latest
     steps:
       - uses: release-drafter/release-drafter@v7
-        with:
-          token: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/release-drafter.yml
+++ b/.github/workflows/release-drafter.yml
@@ -1,4 +1,5 @@
 name: draft release
+
 on:
   push:
     branches:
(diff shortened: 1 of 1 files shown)


dordsor21 (approved):

.github/workflows/release-drafter.yml
```diff
@@ -13,5 +13,5 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: release-drafter/release-drafter@v7
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
```
PierreSchwang, line 23: Should work without, no?

> Introduce the token input, replaces the (need for) GITHUB_TOKEN env variable. Defaults to ${{ github.token }}

Timongcraft: Yea, right...

So it seems like you need to set the permissions in the workflow maybe?
Because they need to be set on the token and other workflows seem to set a permission
e.g. https://github.com/open-telemetry/opentelemetry-erlang-contrib/blob/main/.github/workflows/release-drafter.yml#L14-L15
And the old token via environment might have circumvented that maybe?

> Error: Resource not accessible by integration - https://docs.github.com/rest/releases/releases#create-a-release

https://docs.github.com/en/rest/releases/releases?apiVersion=2026-03-10#create-a-release--fine-grained-access-tokens

NotMyFault:
```
permissions:
  contents: read
```
and
```
with:
  token: ${{ secrets.GITHUB_TOKEN }}

```
should lead to the desired result. That's what we're using at https://github.com/jenkinsci/jenkins/blob/master/.github/workflows/changelog.yml, where I got this original workflow from.

Timongcraft: The job needs write permission, no?
https://github.com/jenkinsci/jenkins/blob/master/.github/workflows/changelog.yml#L19

NotMyFault: That's needed as well, yea.

Timongcraft: But as Pierre said, the token configuration shouldn't be needed anymore, right?

NotMyFault: I mean, the PR is failing still :/

Timongcraft: But aren't external PRs restricted or am I missing something?
Maybe open one from this repo?

PierreSchwang: Yes, iirc release drafter has always failed for PRs from foreign repositories - as it should. Could potentially update the action to not run in that case.

Timongcraft: Alright, but then the condition is broken/inverted because currently it skips for prs from the main repo: https://github.com/IntellectualSites/FastAsyncWorldEdit/actions/runs/26391439091/job/77681855616?pr=3540
https://github.com/IntellectualSites/FastAsyncWorldEdit/actions/runs/26363033492/job/77601805506?pr=3545

PierreSchwang: Not sure why it has to run on PRs either way, on push to main should suffice (just like Jenkins does it). Basically https://github.com/jenkinsci/jenkins/blob/ce830a65166fab4c7358bf6da44f8fdb29c1ce93/.github/workflows/changelog.yml#L5-L21

PR #3543, opened 2026-05-24
fix: improve lock and sync when calling and trimming

dordsor21:
- also fix error when editing newly generated chunks due to postprocessgen not having been completed
 - fixes #3490

PR #3564, opened 2026-06-18
Remove block entities on server thread

HaHaWTH:
## Overview
Fix unsafe async block entity removal in Paperweight chunk write adapters.

## Description

This PR fixes an issue where PaperweightGetBlocks#internalCall could remove block entities directly while running on FAWE queue worker threads.

Previously, when an edit replaced blocks in a chunk, existing block entities were removed immediately during chunk processing via LevelChunk#removeBlockEntity or the Paper beacon removal path. As the backing `blockEntities` map is not designed for concurrent writes, this can cause backing map corruption.

### Existing Issues

After removing BEs in the final sync tasks, this may spam logs like `Block entity A @ BlockPos state Block{B}[facing=north,type=single,waterlogged=false] invalid for ticking:` as the removal happens after the chunk section swap.

I am not sure whether the preferred long-term fix is to:

- keep this approach as the minimal thread-safety fix,
- move block entity removal into an earlier synchronous phase before the section swap,

Feedback would be appreciated here, especially around the expected ordering between block entity removal, beacon deactivation events, and the chunk section replacement performed by internalCall.

### Submitter Checklist

### DIFF
--- a/worldedit-bukkit/adapters/adapter-1_21/src/main/java/com/sk89q/worldedit/bukkit/adapter/impl/fawe/v1_21_R1/PaperweightGetBlocks.java
+++ b/worldedit-bukkit/adapters/adapter-1_21/src/main/java/com/sk89q/worldedit/bukkit/adapter/impl/fawe/v1_21_R1/PaperweightGetBlocks.java
@@ -347,6 +347,7 @@ public class PaperweightGetBlocks extends AbstractBukkitGetBlocks<ServerLevel, L
         // Remove existing tiles. Create a copy so that we can remove blocks
         Map<BlockPos, BlockEntity> chunkTiles = new HashMap<>(nmsChunk.getBlockEntities());
         List<BlockEntity> beacons = null;
+        List<BlockEntity> tilesToRemove = new ArrayList<>();
         if (!chunkTiles.isEmpty()) {
             for (Map.Entry<BlockPos, BlockEntity> entry : chunkTiles.entrySet()) {
                 final BlockPos pos = entry.getKey();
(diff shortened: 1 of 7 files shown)


SirYwell:
Relevant: #3152

SirYwell:
Also #3543 I think

HaHaWTH:
Workaround for log spam: set the `remove` field through reflection inside the loop, this should be a minimal fix

HaHaWTH:
Reference for event contract: BlockEntity removal should happen after the event
image
image

dordsor21:
Is it not the case that we need to remove the tiles before we perform the edit? Do ticks occurring mid-edit not have an issue if we've replaced the tile in a chunk section it then attempts to do something with

HaHaWTH:
> Is it not the case that we need to remove the tiles before we perform the edit? Do ticks occurring mid-edit not have an issue if we've replaced the tile in a chunk section it then attempts to do something with

Yes, that's what I'm talking about in the Existing Issues section, because Paper has validation checks before block entity ticking, this won't cause unexpected behavior but spam logs like `Block entity A @ BlockPos state Block{B}[facing=north,type=single,waterlogged=false] invalid for ticking:`.

dordsor21:
> Yes, that's what I'm talking about in the Existing Issues section, because Paper has validation checks before block entity ticking, this won't cause unexpected behavior but spam logs like `Block entity A @ BlockPos state Block{B}[facing=north,type=single,waterlogged=false] invalid for ticking:`.

Yep I see. w.r.t #3152 this is fixed by putting the sync before the rest of the function - I would still request the same as I did in that PR where we should be chaining the future(s) together (if not already on main thread) similar to how it is handled for if we do/do not yet have the chunk loaded

PR #3577, opened 2026-07-08
Disable global megabuffer

SirYwell:
## Description

The global megabuffer is subject to race conditions in FAWE. If at all, it should be implemented in a way that is thread-safe and player(or other context)-scoped.

The megabuffer also only grows, which is also a terrible design.

### Submitter Checklist

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/internal/expression/Functions.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/internal/expression/Functions.java
@@ -133,17 +133,18 @@ public final class Functions {
         map.put("swap", lookup.findStatic(Functions.class, "swap",
                 methodType(double.class, Variable.class, Variable.class)
         ));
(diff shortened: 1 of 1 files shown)


PierreSchwang (approved):
The documentation should be updated when this is merged, so that the removal is explicitly stated (potentially with the reason)

NotMyFault (approved):

NotMyFault:
> The documentation should be updated when this is merged, so that the removal is explicitly stated (potentially with the reason)

I didn't find anything in our docs about it?

PierreSchwang:
> > The documentation should be updated when this is merged, so that the removal is explicitly stated (potentially with the reason)
> 
> 
> 
> I didn't find anything in our docs about it?

It should be noted somewhere imo. It's not like most attempt to read documentation, but at least there'll be information somewhere if anyone intends to use the functions

PR #3592, opened 2026-07-14
Address gradle deprecations

NotMyFault:
The change proposed addresses deprecations gradle complained about.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -12,28 +12,33 @@ plugins {
     id("xyz.jpenilla.run-paper") version "3.0.2"
 }
 
-var rootVersion by extra("2.15.4")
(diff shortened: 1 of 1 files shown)


build.gradle.kts
```diff
@@ -12,28 +12,33 @@ plugins {
...
+    snapshot + "-" + project.property("buildnumber") as String
+} else {
+    project.property("snapshot") as String
+}
+
+extra.set("rootVersion", rootVersion)
+extra.set("snapshot", snapshot)
+extra.set("revision", revision)
+extra.set("buildNumber", buildNumber)
+extra.set("date", date)
```
NotMyFault, line 36: Unsure if this needs to be exposed, but there's a file in the jar these information are stored in, iirc.

Issue #3420, opened 2026-01-01
FAWE commands not running at all


DoogiePC:
### Server Implementation

Paper

### Server Version

1.21.11

### Describe the bug

FAWE build 1257
Every worldedit command completely freezes up, no command will run, not even a message saying anything. Even /fawe debugpaste freezes up.

It seems like if certain worldedits are done which we can't figure out yet, everything locks and freezes.

Additionally, if I run /save-all flush
it lags a TON https://pastebin.com/SuUjCtq5
It only lags when this issue is occurring, if I run that command normally, no issue occurs. It seems like chunks are being held up somewhere.

(Addons: gobrush, gopaint, arceon, voxelsniper all updated to the latest version)

### To Reproduce

We're not sure.

### Expected behaviour

The worldedit to go through as normal and not freeze up the servers worldedit.

### Error log (if applicable)

[crash-2026-01-01_04.02.21-server.txt](https://github.com/user-attachments/files/24399199/crash-2026-01-01_04.02.21-server.txt)
The server also crashed abruptly once which may be related?

[latest (7).log](https://github.com/user-attachments/files/24399200/latest.7.log)
Latest log, the log ends where the issue began.

### Fawe Debugpaste

https://athion.net/ISPaster/paste/view/a2c3aa2f35d040d4b505991f75d5151c

PierreSchwang:
Does it only happen with the named build and or did that happen with earlier builds as well?

DoogiePC:
> Does it only happen with the named build and or did that happen with earlier builds as well?

I don't recall it happening before this, so I will downgrade to #1253 and see if it does it on that version.
(On version #1253, it doesn't do that weird thing with //brush sphere where it ignores where you look to place the sphere)

DoogiePC:
> Does it only happen with the named build and or did that happen with earlier builds as well?

Ok, on version #1253 it still does it. I managed to run fawe threads and heres the log, idk if it helps:
https://pastebin.com/N0VhvDqc

<img width="1270" height="84" alt="Image" src="https://github.com/user-attachments/assets/368c95e7-8c12-4438-b1af-cdab1c6e4ef2" />
I tested by setting 1 block with worldedit, it took 118 seconds.

DoogiePC:
Ok so update:

I've tested with papermc builds including the earliest paper build for 1.21.11
and the earliest FAWE build for 1.21.11
it does it with all of them.

[...]

PR #3600, opened 2026-07-16
chore: raise minimum Java version to 25

MattBDev:
## Summary

Raises the minimum Java version required to run FAWE from 21 to 25.

## Motivation

Java 25 aligns with the current stable LTS path and enables future optimizations:
- Virtual threads are now safe for consideration (JEP 491 fixed `synchronized` pinning in Java 24)
- Better long-term compatibility as the Java ecosystem moves forward

## Changes

- Bump `sourceCompatibility` and `targetCompatibility` from Java 21 to Java 25
- Set `options.release` to 25 for consistent classfile targets
- Update `TargetJvmVersion` attribute from 21 to 25 in library configurations
- Conditionally disable `javadoc`/`javadocJar` tasks unless publishing or signing is requested, to reduce build overhead on regular `build`/`assemble` runs

## Testing

Verified compilation across:
- `worldedit-core`
- `worldedit-bukkit`
- `worldedit-bukkit:adapters:adapter-1_21` (oldest)
- `worldedit-bukkit:adapters:adapter-26.2` (newest)

All pass with no errors on Java 25.

### DIFF
--- a/build-logic/src/main/kotlin/buildlogic.common-java.gradle.kts
+++ b/build-logic/src/main/kotlin/buildlogic.common-java.gradle.kts
@@ -15,7 +15,9 @@ tasks
         val disabledLint = listOf(
             "processing", "path", "fallthrough", "serial", "overloads", "this-escape",
         )
-        sourceCompatibility = "21"
+        sourceCompatibility = "25"
+        targetCompatibility = "25"
+        options.release.set(25)
         options.compilerArgs.addAll(listOf("-Xlint:all") + disabledLint.map { "-Xlint:-$it" })
         options.isDeprecation = true
         options.encoding = "UTF-8"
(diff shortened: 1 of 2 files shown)


SirYwell:
Not a fan of this as long as people are actively using Minecraft versions that don't support bytecode rewriting of Java 25 bytecode. Running the server with Java 25 is already supported and there is no immediate need for Java 25 features from our side.

MattBDev:
@SirYwell I proposed this because @dordsor21 was interested in possibly using virtual threads to address #3420. The only way this is feasible is if we update to Java 25.
Also, it we have typically raised required Java versions with Minecraft. This just continues our existing policy that wasn't enforced earlier in the year.

SirYwell:
> @SirYwell I proposed this because @dordsor21 was interested in possibly using virtual threads to address #3420. The only way this is feasible is if we update to Java 25. Also, it we have typically raised required Java versions with Minecraft. This just continues our existing policy that wasn't enforced earlier in the year.

We're already actively using virtual threads.

The change in update frequency as well as our usage statistics basically mean we would cut off a large portion of servers from FAWE updates with little benefit.

MattBDev:
> > @SirYwell I proposed this because @dordsor21 was interested in possibly using virtual threads to address #3420. The only way this is feasible is if we update to Java 25. Also, it we have typically raised required Java versions with Minecraft. This just continues our existing policy that wasn't enforced earlier in the year.
> 
> We're already actively using virtual threads.
> 
> The change in update frequency as well as our usage statistics basically mean we would cut off a large portion of servers from FAWE updates with little benefit.

In order to use virtual threads to fix this issue specifically we need the fix from JEP 491 because FAWE's chunk-editing code is wrapped in synchronized blocks. Without JEP 491 we are just trading one issue for another (well, they both still deadlock).
We can close this if we want to go ahead with a different route.

Zoriot:
I wanted to verify: Should this issue be resolved or happen less often on Java 25? If so, maybe it would be useful to notify users when applicable. Most servers in the supported range should have the ability to upgrade Java.

MattBDev:
> I wanted to verify: Should this issue be resolved or happen less often on Java 25? If so, maybe it would be useful to notify users when applicable. Most servers in the supported range should have the ability to upgrade Java. 

I abandoned the PR after it seemed to be a non starter with the team.
It will probably be revisited by the end of the year. A solution for issue #3420 that doesn't use virtual threads is in the works.

PR #3622, opened 2026-08-14
fix(tmp): try make clipboard patterns safe for Filter

Timongcraft:
## Overview
Fix corruption when using ClipboardPattern and disabling on disk clipboards as some implementations have an unsafe cache.

This does not fix RandomFullClipboardPattern, which also seems affected but I'm not sure how to wrap the clipboard properly as it uses the old single get clipboard (seemingly on purpose) but idk if or how that should be wrapped as we'd need a map or list, no?

## Description
Add a temporary wrapper to sync block access in clipboards for ClipboardPattern.

### Submitter Checklist
~~- [ ] New public fields and methods are annotated with `@since TODO`.~~

### DIFF
--- a/worldedit-core/src/main/java/com/fastasyncworldedit/core/extent/clipboard/ConcurrentReadClipboard.java
+++ b/worldedit-core/src/main/java/com/fastasyncworldedit/core/extent/clipboard/ConcurrentReadClipboard.java
@@ -22,12 +22,7 @@ public class ConcurrentReadClipboard implements Clipboard {
     private final Clipboard parent;
 
     public static Clipboard tryWrap(Clipboard clipboard) {
-        if (clipboard instanceof ConcurrentReadClipboard
-                || clipboard instanceof WorldCopyClipboard
-                || clipboard instanceof EmptyClipboard
-                || clipboard instanceof DiskOptimizedClipboard // FastSchematicReaderV2/V3 no min offset
-                || (clipboard instanceof BlockArrayClipboard blockArrayClipboard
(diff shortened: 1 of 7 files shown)


Timongcraft:
> Caused by: org.gradle.internal.resolve.ArtifactNotFoundException: Could not find yarn-1.21.1+build.3-mergedv2.jar (net.fabricmc:yarn:1.21.1+build.3).

hm...

worldedit-core/src/main/java/com/fastasyncworldedit/core/extent/clipboard/ConcurrentReadClipboard.java
```diff
@@ -0,0 +1,151 @@
...
+
+    private final Clipboard parent;
+
+    public static Clipboard tryWrap(Clipboard clipboard) {
+        if (clipboard instanceof ConcurrentReadClipboard
+                || clipboard instanceof WorldCopyClipboard
+                || clipboard instanceof EmptyClipboard
+                || clipboard instanceof DiskOptimizedClipboard // FastSchematicReaderV2/V3 no min offset
+                || (clipboard instanceof BlockArrayClipboard blockArrayClipboard
+                && blockArrayClipboard.getParent() instanceof DiskOptimizedClipboard)) {
```
dordsor21: I know it's still in draft, but perhaps we can add a #concurrentAccessSupported function to clipboard

Timongcraft: I appreciate the feedback!
that would be better I just wasn't sure if I should touch that but will do!

PR #3626, opened 2026-08-19
fix: write correct entity y/z coordinates when saving clipboard to disk

NyaNexNah:
## Overview

Fixes a bug in `DiskOptimizedClipboard` where entity positions saved to a disk clipboard are corrupted. When writing entity NBT data, all three elements of the `Pos` list were set to the entity's x coordinate:

```java
pos.add(new DoubleTag(entity.getLocation().x()));
pos.add(new DoubleTag(entity.getLocation().x())); // should be y
pos.add(new DoubleTag(entity.getLocation().x())); // should be z
```

The read side (`loadNBTFromFileFooter()`) correctly reads `Pos[0..2]` as x, y, z, so every entity loaded from a disk clipboard ends up with its y and z position replaced by x. This fixes the write side to use the y and z coordinates respectively.

### Submitter Checklist


PierreSchwang (approved):
seems reasonable (weird that this wasn't an issue until now?)

PR #3635, opened 2026-08-25
Modernize instanceof checks, switch expressions, and error handling

MattBDev:
## Summary
- Consolidate `instanceof` checks to pattern-variable form and convert legacy `switch` statements to arrow-form switch expressions across the codebase (NBT streams, block/direction/entity registries, expression compiler, YAML, math utils, etc.)
- Improve error message formatting and use try-with-resources for `Closer` in `BukkitImplLoader`

- Errorprone library was added to `libs.versions.toml` for future work when catching up to upstream. Included it just in case we choose to adopt it too.


worldedit-bukkit/src/main/java/com/sk89q/worldedit/bukkit/BukkitAdapter.java
```diff
@@ -153,12 +153,11 @@ public static BukkitPlayer adapt(Player player) {
...
-        if (actor instanceof com.sk89q.worldedit.entity.Player) {
-            return adapt((com.sk89q.worldedit.entity.Player) actor);
-        } else if (actor instanceof BukkitBlockCommandSender) {
-            return ((BukkitBlockCommandSender) actor).getSender();
-        }
-        return ((BukkitCommandSender) actor).getSender();
+        return switch (actor) {
+            case com.sk89q.worldedit.entity.Player player -> adapt(player);
+            case BukkitBlockCommandSender bukkitBlockCommandSender -> bukkitBlockCommandSender.getSender();
+            case null, default -> ((BukkitCommandSender) actor).getSender();
```
MattBDev, line 159: I think this may result in an NPE. I just want a second opinion on this before I go and fix it before the PR is approved and merged.

SirYwell:
1. I feel like we shouldn't touch WE classes/code. If they applied similar changes, cherry-picking them might make more sense.
2. So errorprone is included because WE uses it? I'm not exactly a fan of it, due to its tight integration with internal javac apis.
3. The code style seems rather inconsistent (e.g., usage of `final` in instanceof pattern matching, that said I'm not a fan of noisy `final` usage at all).
4. It might make more sense to replace `Closer` usage with try-with-resources completely. AFAIK it's a pre TWR artifact to make dealing with closables easier.

MattBDev:
> 1. I feel like we shouldn't touch WE classes/code. If they applied similar changes, cherry-picking them might make more sense.
> 2. So errorprone is included because WE uses it? I'm not exactly a fan of it, due to its tight integration with internal javac apis.
> 3. The code style seems rather inconsistent (e.g., usage of `final` in instanceof pattern matching, that said I'm not a fan of noisy `final` usage at all).
> 4. It might make more sense to replace `Closer` usage with try-with-resources completely. AFAIK it's a pre TWR artifact to make dealing with closables easier.
@SirYwell I'll address each comment in the order you made them.
1. Most, if not all of the WE code that was touched was because they have already made the same change. Cherry-picking didn't seem appropriate since some of the commits had other changes included. Not to mention the fact that many of the commits where these changes were made happened 2-3 years ago.
2. Yes, errorprone is used by WE. I can take it out if that's a dealbreaker on this PR. Considering I didn'tgo about actually adding it as a dependency and listed it only in the `libs.versions.toml` file it is an easy change.
3. I can address the usage of final in the pattern matching. Exactly the feedback I was looking fof. 😅 
4. No comment.

PR #3641, opened 2026-09-06
Bind BlockTypes constants to explicit ids instead of field order

MattBDev:
### What this changes

Our changes to `BlockTypes` have led to falling behind updating constants for new Minecraft versions and to out-of-order constants. The reflective scheme used to bind
them to their ids is also fragile. This change removes the reflective scheme and matches the upstream `BlockTypes` class, which binds each constant to its id explicitly.
Before, each block constant called a no-argument `init()` that walked the class's own field array and derived
the id from whichever field happened to sit at the next index:

```java
private static Field[] fieldsTmp;
private static int initIndex;

public static BlockType init() {
    if (fieldsTmp == null) {
        fieldsTmp = BlockTypes.class.getDeclaredFields();
        BlockTypesCache.$NAMESPACES.isEmpty(); // initialize cache
    }
    String name = fieldsTmp[initIndex++].getName().toLowerCase(Locale.ROOT);
    return BlockType.REGISTRY.get(name);
}

// Clears memory after initialization
static {
    // we should be at the first non-BlockType field now
    if (!fieldsTmp[initIndex].getName().equals("fieldsTmp")) {
        throw new IllegalStateException("improper initialization of block type fields");
    }
    fieldsTmp = null;
}
```

Each constant now states its id and uses the existing `get(String)` method to resolve it:

```java
@Nullable
public static final BlockType ACACIA_BUTTON = get("minecraft:acacia_button");
```

```java
@Nullable
public static BlockType get(String id) {
    return BlockType.REGISTRY.get(id);
}
```

No new method was introduced.

### Why

The biggest problem is that constant resolution relies on `Class.getDeclaredFields()`. This method **does
not specify an order**. We just happened to luck into the fact that the JVM returned fields in source order, but that is not
guaranteed and we shouldn't be left relying on that.

The consequence of this is that every constant from the first disturbance
onward is bound to the *wrong block*. This can lead to silent corruption of world data, which is a severe issue.

Writing the id straight into each constant just makes the whole problem impossible instead of trying to catch it after the fact.

**Upstream already does it this way.** WorldEdit's own `BlockTypes` writes every constant as `get("minecraft:acacia_button")`. We were doing our own thing here for no real reason; now we're not, so pulling in upstream changes to this file gets a lot easier.
If we fall behind updating Minecraft blocks in the future, we should be able to just cherry-pick the upstream change.

### Also in this change

**Cache priming moved to a static block.** Reading `BlockType.REGISTRY` triggers `BlockType`'s class initializer, not `BlockTypesCache`'s, and `REGISTRY` starts empty. `BlockTypesCache`'s static block is what queries the platform and registers every block into it, so something has to force that class to initialize before the first constant resolves.
That was the `BlockTypesCache.$NAMESPACES.isEmpty()` call, guarded by a `cachePrimed` flag so it ran once.

Since `init` no longer has lazy state of its own, the priming moves to a static block placed above the constants:

```java
static {
    // Initializing the cache is what populates BlockType.REGISTRY from the running platform. Static initializers run in
    // declaration order, so this must stay above the constants for them to resolve to anything.
    //noinspection ResultOfMethodCallIgnored
    BlockTypesCache.$NAMESPACES.isEmpty();
}
```

Static blocks and field initializers compile into the same `<clinit>` in source order, so this runs first and exactly once. The `cachePrimed` field and the branch it guarded on every call are gone. `BlockTypesCache` holds no reference back to `BlockTypes`, so there is no initialization cycle.

**Javadoc added to the public parsing and lookup methods** (`parse`, `get`, `size`, and friends), which previously had none.
`__RESERVED__` gains a Javadoc explaining what it is and that it must never be placed in a world.

### Constant list reconciled with upstream

FAWE's list had drifted from WorldEdit's. This change brings it back into line:

| | |
|---|---|
| **Added** | `DRIED_GHAST`, declared upstream, absent from FAWE |
| **Removed** | `POTTED_AZALEA`, `POTTED_FLOWERING_AZALEA`, declared by FAWE, not by upstream |
| **Un-deprecated** | `POTTED_AZALEA_BUSH`, `POTTED_FLOWERING_AZALEA_BUSH`, which carried a fork-local `@Deprecated //No longer has "bush"` that upstream does not have, and whose comment pointed at the two constants being removed |
| **Kept as the one deliberate exception** | `__RESERVED__`, FAWE-only and load-bearing: it occupies the null index so block types can be represented as primitives |

The block declarations now match upstream.

### Performance

This touches `<clinit>` only. It runs once per JVM, at plugin load.

The one-time startup cost does improve slightly:

| | Removed per JVM |
|---|---|
| `getDeclaredFields()` walk | one native metadata walk over a ~1,210-field class, allocating a defensive `Field` copy for each |
| `toLowerCase(Locale.ROOT)` | one call per constant, each allocating (the names are `SCREAMING_CASE`, so none hit the identity fast path) |
| namespace concatenation | 1,202 `"minecraft:" + key` allocations inside `orDefaultNamespace` |

Roughly 6,000 transient allocations, all of them young-gen garbage, at a moment when the server is already doing far more expensive work. It is not measurable in practice.

The tradeoff: We are now adding about 42 KB to `BlockTypes.class` and roughly 85 KB of memory that sticks around permanently. But the old version wasn't free either. It kept a full copy of the class's fields cached in memory the whole time, which was a similar amount of memory. So the two roughly cancel out, and either way it's nothing on a server with gigabytes of heap.

### DIFF
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/world/block/BlockTypes.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/world/block/BlockTypes.java
@@ -28,6 +28,7 @@ import com.sk89q.worldedit.util.formatting.text.TextComponent;
 import com.sk89q.worldedit.world.registry.LegacyMapper;
 
 import javax.annotation.Nullable;
+import java.lang.invoke.MethodHandles;
 import java.util.Locale;
 import java.util.Set;
 import java.util.stream.Collectors;
@@ -45,17 +46,17 @@ public final class BlockTypes {
     static {
         // Initializing the cache is what populates BlockType.REGISTRY from the running platform. Static initializers run in
         // declaration order, so this must stay above the constants for them to resolve to anything.
-        //noinspection ResultOfMethodCallIgnored
-        BlockTypesCache.$NAMESPACES.isEmpty();
+        try {
+            MethodHandles.lookup().ensureInitialized(BlockTypesCache.class);
+        } catch (IllegalAccessException e) {
+            throw new ExceptionInInitializerError(e);
+        }
     }
 
-    /**
-     * Placeholder occupying the null index, for use where block types are represented as primitives.
-     *
-     * <p>Bound to the synthetic id {@code minecraft:__reserved__}, which {@link BlockTypesCache} registers at internal id
-     * {@link BlockTypesCache.ReservedIDs#__RESERVED__} rather than obtaining from the platform. It is not a real block and
-     * must never be placed in the world.</p>
-     */
+    /* Placeholder occupying the null index, for use where block types are represented as primitives.
+       Bound to the synthetic id minecraft:__reserved__, which BlockTypesCache registers at internal id
+       BlockTypesCache.ReservedIDs.__RESERVED__ rather than obtaining from the platform. Not a real block,
+       must never be placed in the world. */
     @Nullable
     public static final BlockType __RESERVED__ = get("minecraft:__reserved__");
     @Nullable
@@ -2474,13 +2475,6 @@ public final class BlockTypes {
      -----------------------------------------------------
      */
 
-    /**
-     * Parses user input into a block type, using a default parser context.
-     *
-     * @param type the input to parse, with or without a namespace and with or without a property specification
-     * @return the matching block type
-     * @throws InputParseException if the input matches no block type
-     */
     public static BlockType parse(final String type) throws InputParseException {
         return parse(type, new ParserContext());
     }
--- a/worldedit-core/src/main/java/com/sk89q/worldedit/world/block/BlockTypes.java
+++ b/worldedit-core/src/main/java/com/sk89q/worldedit/world/block/BlockTypes.java
@@ -28,7 +28,6 @@ import com.sk89q.worldedit.util.formatting.text.TextComponent;
 import com.sk89q.worldedit.world.registry.LegacyMapper;
 
 import javax.annotation.Nullable;
-import java.lang.reflect.Field;
 import java.util.Locale;
 import java.util.Set;
 import java.util.stream.Collectors;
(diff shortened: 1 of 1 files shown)


worldedit-core/src/main/java/com/sk89q/worldedit/world/block/BlockTypes.java
```diff
@@ -43,2441 +42,2430 @@
...
+        BlockTypesCache.$NAMESPACES.isEmpty();
+    }
+
+    /**
+     * Placeholder occupying the null index, for use where block types are represented as primitives.
+     *
+     * <p>Bound to the synthetic id {@code minecraft:__reserved__}, which {@link BlockTypesCache} registers at internal id
+     * {@link BlockTypesCache.ReservedIDs#__RESERVED__} rather than obtaining from the platform. It is not a real block and
+     * must never be placed in the world.</p>
+     */
```
SirYwell: These are implementation details that shouldn't be part of the documentation

worldedit-core/src/main/java/com/sk89q/worldedit/world/block/BlockTypes.java
```diff
@@ -43,2441 +42,2430 @@
 public final class BlockTypes {
 
     //FAWE start - init
-    // Doesn't really matter what the hardcoded values are, as FAWE will update it on load
+    static {
+        // Initializing the cache is what populates BlockType.REGISTRY from the running platform. Static initializers run in
+        // declaration order, so this must stay above the constants for them to resolve to anything.
+        //noinspection ResultOfMethodCallIgnored
+        BlockTypesCache.$NAMESPACES.isEmpty();
```
SirYwell: Maybe use `MethodHandles.lookup().ensureInitialized(...)` instead?

worldedit-core/src/main/java/com/sk89q/worldedit/world/block/BlockTypes.java
```diff
@@ -2486,9 +2474,28 @@ public static BlockType init() {
      -----------------------------------------------------
      */
 
+    /**
+     * Parses user input into a block type, using a default parser context.
+     *
+     * @param type the input to parse, with or without a namespace and with or without a property specification
+     * @return the matching block type
+     * @throws InputParseException if the input matches no block type
+     */
```
SirYwell: The added documentation is unrelated, can you remove it?

SirYwell (changes requested):

worldedit-core/src/main/java/com/sk89q/worldedit/world/block/BlockTypes.java
MattBDev: I could just make this a comment instead of a javadoc if that's acceptable if we'd rather leave it without a doc?

worldedit-core/src/main/java/com/sk89q/worldedit/world/block/BlockTypes.java
MattBDev: 🤔 I didn't know that existed. I'll look it up and change it if it works.

worldedit-core/src/main/java/com/sk89q/worldedit/world/block/BlockTypes.java
MattBDev: Sure.

SirYwell:
I think it makes sense to target main with this one. But could you remove the unrelated (javadoc) changes?

For v3 we could take a closer look at the methods FAWE adds here, potentially removing them or moving them to a different place. We could also consider putting `__reserved__` into our own namespace rather than abusing the `minecraft` one.

MattBDev:
@SirYwell I thought I removed and pushed the removal of the Javadocs but I guess it didn't go through. It did now though so those comments have been addressed. I will cherry-pick this into main soon after it is approved and merged into v3.

MattBDev:
regarding the comment on `__reserved__` we should put that under a FAWE namespace for v3 at some point.

SirYwell:
MC 26.3 will target main, so doing this work there directly will simplify our lives. Then we should be able to cherry-pick most of the WE 26.3 changes directly.

I'd keep v3 for more significant changes.

MattBDev:
@SirYwell All set. Now based on main.

SirYwell (approved):

PR #3649, opened 2026-09-10
Migrate JSON2NBT to LinBus tree API

MattBDev:
## Summary
- Migrates `JSON2NBT` off the deprecated `com.sk89q.jnbt` tag wrappers to parse directly into `org.enginehub.linbus.tree` types (`LinCompoundTag`, `LinListTag`, `LinByteTag`, etc.), and modernizes the hand-rolled parser internals along the way: `ArrayDeque` instead of `Stack`, char literals instead of magic numbers, an iterative escape-detector instead of a recursive one, and dropped Guava (`Splitter`/`Iterables`/`Lists`) in favor of `String.split`/`Arrays.stream`.
- `getTagFromJson` now returns `LinCompoundTag` instead of the old `CompoundTag`, matching what it actually produces at runtime and how its one caller (`DefaultBlockParser`) already consumes NBT elsewhere in that method.
- `BlanketBaseBlock` gains a `LinCompoundTag` constructor (the old `CompoundTag` one is kept, marked `@Deprecated`, for compatibility) since `DefaultBlockParser` now threads a `LinCompoundTag` through `parseLogic` after switching off the deprecated `getNbtData()`/`CompoundTag` accessors in favor of `getNbt()`.

## Test plan
- [x] `:worldedit-core:compileJava` — the three changed files compile cleanly with no new errors or warnings (pre-existing, unrelated compile errors in `YAMLNode.java` from earlier work on this branch remain and are out of scope here)
- [ ] Manual smoke test: setting a block with inline NBT JSON via the block parser (e.g. `//set chest[...]{...}` syntax) once the branch is otherwise buildable


MattBDev:
Looks like I forgot to fix an issue on modern Java PR. 😅

SirYwell:
What do we use this for? Can we completely replace it instead?

MattBDev:
@SirYwell Addressed and reworked the PR

worldedit-core/src/main/java/com/sk89q/worldedit/extension/factory/parser/DefaultBlockParser.java
```diff
@@ -508,8 +509,8 @@ private BaseBlock parseLogic(String input, ParserContext context) throws InputPa
         if (blockAndExtraData.length > 1 && blockAndExtraData[1].startsWith("{")) {
             String joined = StringMan.join(Arrays.copyOfRange(blockAndExtraData, 1, blockAndExtraData.length), "|");
```
SirYwell, line 510: This looks rather odd. Maybe we can ignore it for now, but I feel like this code generally should be cleaned up at some point.

SirYwell (approved):

SirYwell:
Can we target v3 directly?

MattBDev:
@SirYwell If you prefer, sure.

worldedit-core/src/main/java/com/sk89q/worldedit/extension/factory/parser/DefaultBlockParser.java
MattBDev: I said the same thing. The parser code is generally messy and is definitely due for some cleanup but I figured this would be sufficient for now.

